PR #27, opened 2021-02-19
connector created by bitnine

shinhanbyeol:
connector created by bitnine


emotionbug:
**wrote an initial Type parser for Node.js, actually this is not connector.**

firstly, AGE is part of extension of PG.
so, this library adds a method of parsing if `agtype` exists in the PG.

And, I decided that the type ID was flexible.
So, use the following SQL to get the type ID and use it.
```sql
select typelem
from pg_type
where typname = '_agtype';
```

PR #31, opened 2021-03-03
jdbc connector created by bitnine

Hyundong-Seo:
This connector is AGE-JDBC prototype project.
I developted only SELECT clause using pg-jdbc yet, so I commit only this source.

I executed CREATE clause but I heard that it is not developted yet, so it occured un error.
-> _org.postgresql.util.PSQLException: ERROR: unrecognized node type: -2063706632_
I'll commit other source after developing other clause.


aysko75:
Hello Hyundong, thanks for the pull request! What you have suggested is more like a guide on how to connect to / use AGE running on PostgreSQL. Could you please have this posted on either GitHub Issues or Discussions?

PR #68, opened 2021-05-13
feat: (driver/jdbc) Write Agtype parser for JDBC.

emotionbug:


aysko75:
@emotionbug Thank you!

PR #78, opened 2021-06-03
chore: Sample project using apache age through java.

seo-kw:
Sample project using apache age through java.


audiBookning:
Nice!

Personally to me, it is always easier to get into a project that has simple practical code (samples) for me to "wet my hands".

As a side-note, the [node project](https://github.com/apache/incubator-age/tree/master/drivers/nodejs) in the driver folder (and the jdbc one, i didn't look at it yet), should also be moved to the sample directory and the Agtype parser might be put on its own sub-package?

seo-kw:
Hi there.

Thanks for your opinion.

We're hopefully looking into your opinions.

Thanks again.

Sent with [ProtonMail](https://protonmail.com) Secure Email.

‐‐‐‐‐‐‐ Original Message ‐‐‐‐‐‐‐

Issue #85, opened 2021-06-09
project build problem( make -j option .., bison dependency )


emotionbug:
when use make with -j option like `make -j16` throws error with `cypher_gram_def.h`

**on first trying make,**
```
.//src/include/parser/cypher_gram.h:66:10: fatal error: parser/cypher_gram_def.h: no such file or directory
   66 | #include "parser/cypher_gram_def.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
```

**on twice trying make,**
```
src/backend/parser/cypher_gram.c:147:10: fatal error: cypher_gram_def.h: no such file or directory
  147 | #include "cypher_gram_def.h"
      |          ^~~~~~~~~~~~~~~~~~~
```

it seems trying to compile, before bison generate codes.
so, how about change `Makefile` like behind?

PS, looks like bison not generating with directory part prefix since Bison 3.7.
so, I added `CPPFLAGS` for include header without prefix.
`PG_CPPFLAGS = -I$(ag_include_dir) -I$(ag_include_dir)/parser`

I don't think this is a good way. therefore, how about giving options according to the version or falling old version of bison.
```
Directive: %define api.header.include {"header.h"}
History: Introduced in Bison 3.4. Defaults to ‘"basename.h"’ since Bison 3.7, unless the header file is y.tab.h.
https://www.gnu.org/software/bison/manual/html_node/_0025define-Summary.html
```

**build log**
https://travis-ci.com/github/emotionbug/incubator-age/jobs/512275520

```diff
74c74
< 	  cypher_delete \
---
>           cypher_delete \
82c82
< EXTRA_CLEAN = $(addprefix $(ag_regress_dir)/, $(ag_regress_out))
---
> EXTRA_CLEAN = $(addprefix $(ag_regress_dir)/, $(ag_regress_out)) src/backend/parser/cypher_gram.c src/include/parser/cypher_gram_def.h
85c85
< PG_CPPFLAGS = -I$(ag_include_dir)
---
> PG_CPPFLAGS = -I$(ag_include_dir) -I$(ag_include_dir)/parser
91c91,96
< src/backend/parser/cypher_gram.c: BISONFLAGS += --defines=$(ag_include_dir)/parser/$(basename $(notdir $@))_def.h
---
> src/include/parser/cypher_gram_def.h: src/backend/parser/cypher_gram.c
> 
[...]

PR #86, opened 2021-06-18
fix: Fix errors that occur during parallel builds

emotionbug:
When trying to build `cypher_keywords.c` and `cypher_keywords.c`, `cypher_gram.y` must have been prebuilt, but it does not.
thus, by adding dependencies on this, we correct the problem encountered in Parallel make.

also, including fixes for include header without directory.

related  #85


dehowef (approved):
Looks good to me.

PR #87, opened 2021-06-21
Adding AGType parsers and driver supports for Python & Go  

rhizome-ai:
Hi. I'd like to contribute python & golang drivers(precisely AGType parser and driver support)  for apache/incubator-age.
Main features are 
* Unmarshal AGE result data(AGType) to Vertex, Edge, Path
* Cypher query support for 3rd. Party sql driver (enables to use cypher queries directly)

with some instructions(README and samples) and some test cases in each  package.

Please check out README files.


JoshInnis:
There is a data type that Agtype supports called numeric. Agtype values that are numeric end in '::numeric'. It looks like both python and go parsers are not covering this case. I would also recommend taking a look at https://github.com/apache/incubator-age/blob/master/drivers/Agtype.g4. The type annotation system '::vertex', '::numeric' has the potential to expand in the future if we support more datatypes. Accounting for that in the parser now may save some headaches in the future.

bitninesw:
Thank you, Josh. I missed numeric type. I will add it soon (a couple of days), and I should check the documents and source code more. And according to your recommendation, it takes more time. I need more study about custom type extension mechanism in Go and Python drivers and client side standard usages (including object mappings - ORM and GraphQL)

rhizome-ai:
> There is a data type that Agtype supports called numeric. Agtype values that are numeric end in '::numeric'. It looks like both python and go parsers are not covering this case. I would also recommend taking a look at https://github.com/apache/incubator-age/blob/master/drivers/Agtype.g4. The type annotation system '::vertex', '::numeric' has the potential to expand in the future if we support more datatypes. Accounting for that in the parser now may save some headaches in the future.

Thank you, Josh. I missed numeric type. I will add it soon (a couple of days), and I should check the documents and source code more. And according to your recommendation, it takes more time. I need more study about custom type extension mechanism in Go and Python drivers and client side standard usages (including object mappings - ORM and GraphQL)

muhammadshoaib:
## General Comments:

What is difference between execCypher and queryCipher

Could you please check on the following queries if the driver in question works according to cipher's requirements 

```
MATCH (a:Person) RETURN a.name, a.age, a.country
MATCH (a:Person) RETURN a;

```

Also, path model is relationship/edge model as it does not consider the multiple hop paths. Considering following query the driver's behaviors is unknown 

```
MATCH p=(a)-[:know]-(b)-[:knows]-(c) RETURN a, b, c, p
```

## Subsequent errors: 

I also have observed that if one query fails because of Cypher error all consequences queries start giving following exceptions. A new session solves this problems but this doesn't seems a normally behaviors to me. 

Following is my test case for this. 

```
>>> ag.queryCypher("MATCH (n:Person)")

Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/incubator_age-0.0.2-py3.8.egg/age/age.py", line 90, in execSql
psycopg2.errors.SyntaxError: syntax error at end of input
LINE 1: ... from cypher('my_test_graph', $$ MATCH (n:Person) $$) as (v ...
                                                             ^


During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.8/dist-packages/incubator_age-0.0.2-py3.8.egg/age/age.py", line 107, in queryCypher
  File "/usr/local/lib/python3.8/dist-packages/incubator_age-0.0.2-py3.8.egg/age/age.py", line 100, in execCypher
  File "/usr/local/lib/python3.8/dist-packages/incubator_age-0.0.2-py3.8.egg/age/age.py", line 94, in execSql
age.exceptions.SqlExcutionError: ("SELECT * from cypher('my_test_graph', $$ MATCH (n:Person) $$) as (v agtype);", SyntaxError("syntax error at end of input\nLINE 1: ... from cypher('my_test_graph', $$ MATCH (n:Person) $$) as (v ...\n                                                             ^\n"))
```


From here all subsequent queries starts failing unless I quit the python session and restart a new session 

```
>>> ag.execCypher("MATCH (n:Person) RETURN n")
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/incubator_age-0.0.2-py3.8.egg/age/age.py", line 90, in execSql
psycopg2.errors.UndefinedFunction: function cypher(unknown, unknown) does not exist
LINE 1: SELECT * from cypher('my_test_graph', $$ MATCH (n:Person) RE...
                      ^
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.


During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.8/dist-packages/incubator_age-0.0.2-py3.8.egg/age/age.py", line 100, in execCypher
  File "/usr/local/lib/python3.8/dist-packages/incubator_age-0.0.2-py3.8.egg/age/age.py", line 94, in execSql
age.exceptions.SqlExcutionError: ("SELECT * from cypher('my_test_graph', $$ MATCH (n:Person) RETURN n $$) as (v agtype);", UndefinedFunction("function cypher(unknown, unknown) does not exist\nLINE 1: SELECT * from cypher('my_test_graph', $$ MATCH (n:Person) RE...\n                      ^\nHINT:  No function matches the given name and argument types. You might need to add explicit type casts.\n"))
```



## Record Traversing Errors 

The following code prints nothing 

```
import age

print("Loading age")

DSN = "host=127.0.0.1 port=5432 dbname=age user=age"
TEST_GRAPH_NAME = "my_test_graph"

ag = age.connect(dsn=DSN, graph=TEST_GRAPH_NAME)

ag.execCypher("CREATE (n:Person {name: 'Joe'})")
ag.execCypher("CREATE (n:Person {name: 'Smith'})")

ag.queryCypher("MATCH (n:Person) RETURN n")

for vertex in ag.rows():
    print(vertex)
```

rhizome-ai:
Thank you, Shoaib.  Thanks to good comments from you and Josh, I could easily spot important points.
So I reviewed some alternatives, and revised the structure and API of Python driver.

1. Type converter was fixed to type extension api of Psycopg2. (Psycopg2 is widely used among Python developers, so I don't think this tight coupling will be a big problem.) And I added a usage sample (Jupyter note) for developers familiar with Psycopg2.

2. The execution API is simplified to match the characteristics of the Psycopg2 transaction API. The 'queryCypher' was removed and merged into execCypher. 
   In using Psycopg driver,
   - You must explicitly commit after executing the data changing statements.
   - When an error occurred while a statement is executed (even though data retrieving), the session is blocked before rollback called. 

3. I added **numeric annotation**, **multiple column return**, **multiple hop paths**. 

Refactoring of Golang driver will be made next week. Due to the characteristics of PG drivers that follow Golang's sql standard, different approaches are required from Python.

muhammadshoaib:
When an error occurred while a statement is executed (even though data retrieving), the session is blocked before rollback called.


Is this the normal behavior of Psycopg? I am not sure I need to check, but if error occurs, rollback should occur and there shall not be any blocking.

rhizome-ai:
> > "When an error occurred while a statement is executed (even though data retrieving), the session is blocked before rollback called."
> 
> Is this the normal behavior of Psycopg? I am not sure I need to check, but if error occurs, rollback should occur and there shall not be any blocking.

[https://www.psycopg.org/docs/usage.html#transactions-control](https://www.psycopg.org/docs/usage.html#transactions-control)

The following database commands will be executed in the context of the same transaction – not only the commands issued by the first cursor, but the ones issued by all the cursors created by the same connection. Should any command fail, the transaction will be aborted and no further command will be executed until a call to the rollback() method.

uriaash:
@rhizome-ai you should probably not call you root package `main` because when importing your code one would get this error: `import "github.com/apache/incubator-age/drivers/golang" is a program, not an importable package`

drivers/golang/models.go
```diff
@@ -0,0 +1,269 @@
...
+
+func (e *SimpleEntity) AsArr() []interface{} {
+	return e.value.([]interface{})
+}
+
+type LabeledEntity struct {
+	Entity
+	id    int64
+	label string
+	props map[string]interface{}
```
uriaash, line 122: The `id`, `label`, and `props` should probably be public and changed to `Id`, `Label`, and `Props`.
Otherwise, you won't be able to use them as described in the README file (see usage example section).

drivers/golang/models.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+func (v *Vertex) String() string {
+	return fmt.Sprintf("V{id:%d, label:%s, props:%v}", v.id, v.label, v.props)
+}
+
+type Edge struct {
+	*LabeledEntity
+	start_id int64
+	end_id   int64
```
uriaash, line 160: Fields should be public

drivers/golang/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+    // Unmarsal result data to Vertex row by row
+	for cypherCursor.Next() {
+		entity, err := cypherCursor.GetRow()
+		if err != nil {
+			t.Fatal(err)
+		}
+		vertex := entity.(*age.Vertex)
+
+		fmt.Println(vertex.id, vertex.label, vertex.props["name"], vertex.props)
```
uriaash, line 109: If you copy this usage sample, you won't be able to use the private fields `id`, `label`, and `props`.

drivers/golang/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+	for cypherCursor.Next() {
+		entity, err := cypherCursor.GetRow()
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		path := entity.(*Path)
+        startVertex = path.start
+        edge = path.rel
```
uriaash, line 141: no such field `rel`

drivers/golang/README.md
```diff
@@ -0,0 +1,160 @@
...
+        endVertex = path.end
```
uriaash, line 142: no such field `end`

drivers/golang/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+		fmt.Println(startVertex, edge, endVertex)
+
+		fmt.Println(startVertex.label, startVertex.props["name"], startVertex.props)
+		fmt.Println(edge.label, edge.props["weight"], edge.props)
+		fmt.Println(endVertex.label, endVertex.props["name"], endVertex.props)
```
uriaash, line 148: cannot use private fields `label` and `props`

Uria1:
@rhizome-ai you should probably not call your root package `main` because when importing your code one would get this error: `import "github.com/apache/incubator-age/drivers/golang" is a program, not an importable package`

drivers/golang/README.md
Uria1: fmt.Println(vertex.id, vertex.label, vertex.props["name"], vertex.props)

drivers/golang/README.md
Uria1: no such field `rel`

drivers/golang/README.md
Uria1: no such field `end`

drivers/golang/README.md
Uria1, line 148: cannot use private fields `label` and `props`

rhizome-ai: Thank you for informing me of my ridiculous but crucial mistake, Uria. As an excuse, I was confused about field publicity for a moment because I developed it in parallel with Python  driver. And the go driver needs additional restructuring and refactoring to meet the current level of python driver. Unfortunately, I would like to have more time because I have other works to do now and cannot concentrate on development immediately. (I promise you to fix it sometime in August.)

Uria1: Sure, you don't need to apologize. Other than these minor issues I think you did a great job, it is much appreciated.

drivers/golang/models.go
Uria1, line 122: There is currently no way of getting all properties of a vertex or edge, we definitely need it.

rhizome-ai:
I am sorry to have kept you waiting so long, Uria. 
I just finished refactoring. I have modified the package structure, and there are some changes to the API.
1. To support data changing queries with return values, I merged ExecCypher, QueryCypher into single ExecCypher function.
2. To support multiple return columns, I added the columnCount(number of RETURN columns) parameter to the ExecCypher function.
`func (a *AgeTx) ExecCypher(columnCount int, cypher string, args ...interface{}) (*CypherCursor, error)`
And you can get an Entity array through CyberCursor's GetRow.
`(c *CypherCursor) GetRow() ([]Entity, error)`
3. To support multi-hop path, Vertex and Edge objects can be referenced using index in a Path object.

Please refer to samples for details.

And, if you want to use this package in 'go mod', you can use the github.com/rhizome-ai/apache-age-go v0.0.3 package.
`require( github.com/rhizome-ai/apache-age-go v0.0.3  )`

JoshInnis:
Every file needs to have the ASF Header in it.

PR #115, opened 2021-09-09
Labels creation functions

muhammadshoaib:
> Patch to create labels for vertices and edges
>
> Two more function create_vlabel and create_elabel has been added.
> User can call these functions to create blank labels.
>
> Regression tests have not been added yet.
>
> Usage as following:
>
> SELECT create_vlabel('cypher', 'vlabel_1');
> SELECT create_elabel('cypher', 'elabel_1');
>
> modified makefile to support UTF.
> This test was failing on my docker.


src/backend/commands/graph_commands.c
```diff
@@ -149,6 +259,8 @@ static Oid create_schema_for_graph(const Name graph_name)
 
 PG_FUNCTION_INFO_V1(drop_graph);
 
+// SELECT create_vlabel("Person'):
```
JoshInnis, line 262: I believe this line needs to be removed

muhammadshoaib: I have removed this line.

jrgemignani:
Everything looks great! It just needs a few more regression tests with a few comment for failure cases i.e. showing it failing.

PR #123, opened 2021-09-30
Age333: for primitive types, push the value directly instead of using add_agtype

quocvietvuong:
Hi guys,
I think this pull request is OK because the additions and deletions in agtype.c file are accurate. 
We should ignore the previous commits shown before the latest commit I made.
Thank you.


dehowef:
HI Viet it is best you remove the irrelevant commits, either through squashing them or stashing your changes. Having unnecessary commits will pollute the commit history for when we need to look back at the git log.

Issue #160, opened 2021-12-16
Set clause with invalid syntax causes abnormal termination


pdpotter:
**Describe the bug**
When using a set clause with an invalid syntax, the server is terminated abnormally and the connection is closed.

**How are you accessing AGE (Command line, driver, etc.)?**
- psql

**What data setup do we need to do?**
```pgsql
...
SELECT * FROM cypher('test_graph', $$CREATE (p:person {id: 1, name: 'Jane'}) return p$$) as (p agtype);
                                              p                                              
---------------------------------------------------------------------------------------------
 {"id": 844424930132088, "label": "person", "properties": {"id": 1, "name": "Jane"}}::vertex
(1 row)
...
```

**What is the necessary configuration info needed?**
/

**What is the command that caused the error?**
```pgsql
SELECT * FROM cypher('test_graph', $$Match (p:person {id: 1}) set name = "Janine"  return p$$) as (p agtype);
```
```
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
```

**Expected behavior**
- A syntax error message like
```
ERROR:  syntax error at or near "n"
LINE 1: ...from cypher('test_graph', $$Match (p:person {id: 1}) set name = "Janine"  return p...
- no server termination
```

**Environment (please complete the following information):**
- Version: bf9d8367bf4f9ba03d8b79aee5e2065c7a9edbc8

**Additional context**
- The correct syntax is
```pgsql
SELECT * FROM cypher('test_graph', $$Match (p:person {id: 1}) set p.name = "Janine"  return p$$) as (p agtype);
```
- This causes indexes to be corrupted as well.

PR #165, opened 2021-12-21
Corrects abnormal termination due to incorrect variable in Set Clause.

emotionbug:
Previously, `ColumnRef` may come according to the Parser rule. but, this case is not considered.
Therefore, add logic to detect these errors.

#160

### DIFF
--- a/src/backend/parser/cypher_clause.c
+++ b/src/backend/parser/cypher_clause.c
@@ -747,10 +747,20 @@ cypher_update_information *transform_cypher_set_item_list(
         TargetEntry *target_item;
         cypher_update_item *item;
         ColumnRef *ref;
-        A_Indirection *ind = (A_Indirection *)set_item->prop;
+        A_Indirection *ind;
         char *variable_name, *property_name;
         Value *property_node, *variable_node;
 
+        // ColumnRef may come according to the Parser rule.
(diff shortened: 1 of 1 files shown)


pdpotter:
This PR indeed fixes #160.

I wanted to add following test for this issue, but I wasn't able to run the tests (using `make installcheck`; I will try again with a clean Debian install when I find the time to see if I can reproduce the issue), so I didn't feel confident about creating a PR towards https://github.com/emotionbug/incubator-age/tree/fix-160 for adding this test.

sql (right after https://github.com/apache/incubator-age/blob/master/regress/sql/cypher_set.sql#L134)
```
SELECT * FROM cypher('cypher_set', $$MATCH (n) SET i = 3$$) AS (a agtype);
```

regress (right after https://github.com/apache/incubator-age/blob/master/regress/expected/cypher_set.out#L375)
```
SELECT * FROM cypher('cypher_set', $$MATCH (n) SET i = 3$$) AS (a agtype);
ERROR:  SET clause expects a variable name
LINE 1: ...ELECT * FROM cypher('cypher_set', $$MATCH (n) SET i = 3$$) A...
                                                             ^
```

Do you think it would be worth adding the test?

emotionbug:
Good sounds, i'll add tests.

jrgemignani (commented):
Looks good

PR #166, opened 2021-12-21
pg_dump support

deem0n:
Allow pg_dump to dump content of the `ag_catalog.*` tables
Fix typo in the `flush_label_oid_cache`


src/backend/utils/cache/ag_cache.c
```diff
@@ -633,7 +633,7 @@ static void flush_label_oid_cache(void)
 {
     HASH_SEQ_STATUS hash_seq;
 
-    hash_seq_init(&hash_seq, label_name_graph_cache_hash);
+    hash_seq_init(&hash_seq, label_oid_cache_hash);
```
JoshInnis, line 636: What was your reasoning for doing this? Our goal is to move away from oids?

deem0n: We doing tests with dump & restore of the database with AGE extension and some graph data. After restore we have broken oids in the ag_catalog.* tables. Which can be easily fixed and I will commit the fix procedure later. But even after the oid fix we still have errors in the `flush_label_oid_cache(void)`. I think that using `label_name_graph_cache_hash` is just a typo. At least after my fix we see no errors with broken cache anymore after dump & restore and oid fixes

deem0n: Eliminating oids would make dump&restore a breeze, but probably will impact query speed ? I did not spent much time in AGE internals to participate in decision making, I guess.

emotionbug:
Looks good to me.

JoshInnis:
@deem0n Please provide a pull request that only includes adding our ag_catalog tables as configuration tables, the saved_result_relinfo issue looks to be a separate issue, given what the next line is. We need to review that as a separate issue from the pg_dump support issue that is the primary change in this PR.

jrgemignani:
Closing this PR as the issues (typos) brought up here have been resolved in commit 0e96fbbf9d6cfb997b8d75da2c8dab1211915319

PR #181, opened 2022-01-27
tests: Move to Github Actions from Travis CI

emotionbug:
In Travis CI, the test of the build with `gnu89` format was impossible, so we modify it.


Makefile
```diff
@@ -92,7 +92,7 @@ EXTRA_CLEAN = $(addprefix $(ag_regress_dir)/, $(ag_regress_out)) src/backend/par
 ag_include_dir = $(srcdir)/src/include
 PG_CPPFLAGS = -I$(ag_include_dir) -I$(ag_include_dir)/parser
 
-PG_CONFIG = pg_config
+PG_CONFIG ?= pg_config
```
emotionbug, line 95: Previously, it always assigning a new variable. so, It can't assigning from Shell.

like ```PG_CONFIG=~/pg11/pg_config make -j16```

.github/workflows/installcheck.yml
```diff
@@ -0,0 +1,45 @@
...
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Get latest commit id of PostgreSQL 11
+        run: |
+          echo "PG_COMMIT_HASH=$(git ls-remote git://git.postgresql.org/git/postgresql.git refs/heads/REL_11_STABLE | awk '{print $1}')" >> $GITHUB_ENV
```
emotionbug, line 17: `REL_11_STABLE` has been specified, but it can be change with the specific version required.

.github/workflows/installcheck.yml
```diff
@@ -0,0 +1,45 @@
...
+        with:
+          path: ~/pg11
+          key: ${{ runner.os }}-v1-pg11-${{ env.PG_COMMIT_HASH }}
+
+      - name: Install PostgreSQL 11
+        if: steps.pg11cache.outputs.cache-hit != 'true'
+        run: |
+          git clone --depth 1 --branch REL_11_STABLE git://git.postgresql.org/git/postgresql.git ~/pg11source
+          cd ~/pg11source
+          ./configure --prefix=$HOME/pg11 CFLAGS="-std=gnu89"
```
emotionbug, line 31: Ansiflag can be given, but the comment in the age contains a single comment that is not compatible with ansi(`-ansi`), so specify it as std98(`-std=gnu98`).

JoshInnis:
these changes are now included in the PG12 updates.

PR #184, opened 2022-02-11
feat: Introduce `enable_auto_column_conversion`

emotionbug:
This is used for untyped cypher clauses. The current cypher query must be of a type to be used, which requires a lot of work from users who only use cypher.

In addition, it is thought that many functional implementations will be needed to determine what data is returned when implementing CLI functions in the future.

Thus, it is implemented in this way and is designated as GUC's 'age.enable_auto_column_conversion', and the default value is FALSE.


emotionbug:
Commit history is too dirty, so I create a new PR. #185

PR #185, opened 2022-02-11
feat: Introduce `enable_auto_column_conversion`

emotionbug:
This is used for untyped cypher clauses. The current cypher query must be of a type to be used, which requires a lot of work from users who only use cypher.

In addition, it is thought that many functional implementations will be needed to determine what data is returned when implementing CLI functions in the future.

Thus, it is implemented in this way and is designated as GUC's 'age.enable_auto_column_conversion', and the default value is FALSE.


src/backend/parser/cypher_analyze.c
```diff
@@ -612,10 +623,18 @@ static Query *analyze_cypher_and_coerce(List *stmt, RangeTblFunction *rtfunc,
     // check the number of attributes first
     if (attr_count != rtfunc->funccolcount)
     {
-        ereport(ERROR,
-                (errcode(ERRCODE_DATATYPE_MISMATCH),
-                 errmsg("return row and column definition list do not match"),
-                 parser_errposition(pstate, exprLocation(rtfunc->funcexpr))));
+        if (age_guc_enable_auto_column_conversion)
```
JoshInnis, line 626: Should the functionality be: if the column definition list doesn't match: check the flag, or if the flag is enabled don't check the column definition list?

src/backend/parser/cypher_analyze.c
```diff
@@ -19,7 +19,6 @@
 
 #include "postgres.h"
 
-#include "catalog/pg_type_d.h"
```
JoshInnis, line 22: I got a compile error. I had to add #include "catalog/pg_type.h" to get to work

src/include/parser/cypher_parse_node.h
```diff
@@ -26,6 +26,12 @@
 #define AGE_DEFAULT_ALIAS_PREFIX "_age_default_alias_"
 #define AGE_DEFAULT_VARNAME_PREFIX "_age_varname_"
 
+typedef struct extends_parsestate
```
JoshInnis, line 29: This needs documentation to explain what it is.

src/backend/parser/cypher_analyze.c
emotionbug: Regarding this logic, it's a little uncomfortable while using it, so I'll revise it as follows.
If the corresponding flag is set, it automatically matches regardless of the number of columns, otherwise, if the number of columns does not match, it will cause an error as before.

src/include/parser/cypher_parse_node.h
```diff
@@ -49,7 +54,6 @@ typedef struct cypher_parsestate
      * else). It is only used by transform_cypher_item_list.
      */
     bool exprHasAgg;
-    bool p_opt_match;
```
emotionbug, line 46: No longer used.

PR #192, opened 2022-03-08
Remove OIDS from `ag_label`

emotionbug:
Although OIDS is defined, the only part that is actually used is the logic of determining the existence of the label.
However, this is eliminated because it can be implemented with other fields such as label_id.

In the case of upgrading this, OIDS may be removed using 'ALTER TABLE ag_label SET WITHOUT OIDS'.

related : https://github.com/apache/incubator-age/issues/20

```pgsql
DROP index ag_label_oid_index;
ALTER TABLE ag_label SET WITHOUT OIDS;
```

### DIFF
--- a/age--1.0.0.sql
+++ b/age--1.0.0.sql
@@ -48,9 +48,7 @@ CREATE TABLE ag_label (
   id label_id,
   kind label_kind,
   relation regclass NOT NULL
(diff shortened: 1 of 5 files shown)


emotionbug:
@jrgemignani Rebased with master.

PR #194, opened 2022-03-09
Remove OIDS from `ag_graph`

emotionbug:
Since PG 12 does not recommend and not allow to use the use of Oids, so
removed Oids. Oid uses 4Byte, it uses INT type for compatibility. and,
create a sequence of "$TABLE_NAME$_$COLUMN_NAME$_seq" and calls it to
create a new ID.

And, due to the rule that a value less than 1 cannot exist, the case
where it does not exist is considered 0 (INVALID_AG_GRAPH_ID).

### DIFF
--- a/age--1.0.0.sql
+++ b/age--1.0.0.sql
@@ -24,12 +24,14 @@
 -- catalog tables
 --
 
+CREATE DOMAIN graphoid AS int NOT NULL CHECK (VALUE > 0);
+CREATE SEQUENCE ag_graph_id_seq START 1;
+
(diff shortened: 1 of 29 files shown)


deem0n:
Also, if you dump & restore, after that you'll get age crashes. You need to manually fix oids in the `ag_graph` & `ag_label` tables. I have a plsql procedure for that, but it will be unnecessary if we remove OIDs.

emotionbug:
@jrgemignani rebased master.

PR #212, opened 2022-04-29
Add access operator (`->`, `->>`) to Agtype

emotionbug:
Added the `->`, `->>` operator to agtype for approaching like Json(b).


JoshInnis:
The primary concern I have is that the cypher command itself will not use these operators as is. Cypher by default only uses Agtype, so I believe these operators and indices built on these operators will not be usable by cypher, rather only standard sql queries.

PR #219, opened 2022-05-23
Merge actions

vinzenz:
An attempt to add support for merge actions to AGE. This is currently WIP and I will need some help understanding the code a bit better.
Currently I am lost on what to do with the parsed code, to be able handling the on create/on match cases.

I somehow don't understand how everything works at this moment, so I will need some pointers or some explanation how to continue, if one of you is willing to do so, it would be much appreciated, for me it's a big missing piece to the puzzle on making it convenient to use, otherwise I will be forced to have a lot of round trips to the server for data insertion/updates.

Signed-off-by: Vinzenz Feenstra <evilissimo@gmail.com>

### DIFF
--- a/regress/expected/cypher_merge.out
+++ b/regress/expected/cypher_merge.out
@@ -815,6 +815,45 @@ SELECT * FROM cypher('cypher_merge', $$MATCH (n) DETACH DELETE n $$) AS (a agtyp
 (0 rows)
 
 /*
+ * test 31:
+ * MERGE + Actions
+ */
+
+-- Create a node with .created and .updated properties set to the same timestamp
+SELECT * FROM cypher('cypher_merge', $$MERGE (n {value: 1}) ON CREATE SET n.created = timestamp(), n.updated = timestamp() ON MATCH SET n.updated = timestamp() RETURN n.value $$) AS (a agtype);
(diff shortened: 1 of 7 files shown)


JoshInnis:
Hi @vinzenz,

JoshInnis:
Hi @vinzenz 

So a couple things. Merge actions is an issue that we need to address, so thank you for trying to get the ball rolling. :). BUT, you need to do more than change the expected out from the regression tests if you change the regression SQL, AND you need to throw an error if a user inputs a grammatically correct statement with MERGE actions that we cannot currently handle.

jrgemignani:
@vinzenz @JoshInnis What is the current status of this PR? The PR looks to need, at the very least, to be rebased against the current master branch. But, it is not known if there was a plan to go forward, or not at all.

vinzenz:
@jrgemignani well it looks like, i don't have the knowhow to finish it and not enough time to get it in, so i will close it.
It was marked as draft already anyway as it is not even nearly finished.


Closing

jrgemignani:
@vinzenz Thank you for the quick response and the attempt! I will try to see about getting MERGE actions added to our list of tasks.

PR #227, opened 2022-06-07
Fix and extend index tests

pdpotter:
Fix and extend index tests as mentioned in https://github.com/apache/incubator-age/issues/45#issuecomment-1121484835


pdpotter:
Some tests have already been fixed. Closing this PR and opening another one if further fixes are necessary.

Issue #318, opened 2022-10-06
PGXN Support.


ibrarahmad:
We need to add support for PGXN. This requires some additional files.

PR #319, opened 2022-10-06
Issue (#318): PGXN Support added.

ibrarahmad:
PGXN, the PostgreSQL Extension network, is a central distribution system for open-source PostgreSQL extension libraries. A new file META.json is require to support that platform.

### DIFF
--- a/META.json
+++ b/META.json
@@ -1,0 +1,46 @@
+{
+   "name": "ApacheAGE",
+   "abstract": "Apache AGE is a PostgreSQL Extension that provides graph database functionality",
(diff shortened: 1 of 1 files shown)


META.json
```diff
@@ -0,0 +1,46 @@
...
+   },
+   "generated_by": "andrew.ko@bitnine.net",
+   "meta-spec": {
+      "version": "1.0.0",
+      "url": "http://pgxn.org/meta/spec.txt"
+   },
+   "tags": [
+      "graphdb",
+      "graph-database",
+      "agensgraph"
```
JoshInnis, line 44: We cannot exclude AgensGraph in AGE. AgensGraph is a copyright owned by Bitnine and Apache does not allow copyrighted tags in its software.

META.json
```diff
@@ -0,0 +1,46 @@
...
+         "abstract": "Apache AGE is a PostgreSQL Extension that provides graph database functionality",
+         "file": "age--1.1.0.sql",
+         "docfile": "README.md",
+         "version": "1.1.0"
+      }
+   },
+   "prereqs": {
+      "runtime": {
+         "requires": {
+            "PostgreSQL": "12.0.0"
```
JoshInnis, line 21: The master branch supports Postgres 11, only the AGE_PG12.1.0_ALPHA branch supports postgres 12.

META.json
```diff
@@ -0,0 +1,46 @@
...
+   "version": "1.1.0",
+   "maintainer": [
+      "andrew.ko@bitnine.net"
```
JoshInnis, line 7: Please change to the apache account of a PMC member.

META.json
```diff
@@ -0,0 +1,46 @@
+{
+   "name": "ApacheAGE",
```
JoshInnis, line 2: Can there be a space between Apache and AGE?

ibrarahmad: Not Possible, I tried that PGXN does not support.

META.json
JoshInnis: I am not sure how the PGXN system works, but if it works on a branch by branch basis, then yes, create a PR for that branch and alter this to PG 11

ibrarahmad: Yes, it is, I will create for that branch.

JoshInnis: Ok, switch this to use 11 and the PG 12 branch for that

PR #320, opened 2022-10-07
Naming rules for graphs

Johan511:
Following naming rules were implemented

1. Graph name must begin either with an alphabet or an underscore.
2. The name can also include numbers after the first character.
3. No spaces or other special characters except underscore should be allowed.

Error Code is ERRCODE_INVALID_NAME => 42602
function to check validity is in the same file, please suggest an alternative file where I can instead define them to maintain cleanliness.


jrgemignani:
Rejected.

While I appreciate your enthusiasm, this is an **open topic for discussion** and **nothing** has been decided yet. You are welcome to add your thoughts to the discussion.

Additionally, this task was going to be assigned to one of our interns, once this discussion was complete.

john

Issue #369, opened 2022-11-26
Crash on Edge MERGE after Edge MATCH


TropicalPenguin:
**Describe the bug**
Usually, the MERGE operation works. But when a merge operation follows a MATCH clause that includes an edge, the server terminates.

**How are you accessing AGE (Command line, driver, etc.)?**
- Command line
- Golang driver

**What data setup do we need to do?**
```pgsql
...
SELECT create_graph('test');
SELECT * from cypher('test', $$ CREATE (a {x:1})-[:foo]->(b {x:2}),(c {x:3}) $$) as (v0 agtype);
...
```

**What is the necessary configuration info needed?**
- N/A

**What is the command that caused the error?**
```pgsql
SELECT * from cypher('test', $$ MATCH ()-[a:foo]->(), (b {x:2}),(c {x:3}) MERGE (b)-[d:bar]->(c) RETURN d $$) as (v0 agtype);
```
Error:
```
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
```

**Expected behavior**
The relationship should be created if it doesn't exist.

**Environment (please complete the following information):**
- Version: 1.1.0, commit ed44f91cb2607605ec2e225248e9ca46e146c215; also fails in earlier releases.

**Additional context**
N/A

TropicalPenguin:
Bounty opened: https://app.bountysource.com/issues/114119874-crash-on-edge-merge-after-edge-match

jrgemignani:
Just curious, why are you matching something that is extraneous - **()-[a:foo]->()** to the merge? Because, if you remove that part, the match works correctly.

```
psql-11.5-5432-pgsql=# SELECT * from cypher('test', $$ MATCH (b {x:2}),(c {x:3}) MERGE (b)-[d:bar]->(c) RETURN d $$) as (v0 agtype);
                                                            v0
--------------------------------------------------------------------------------------------------------------------------
 {"id": 9570149208162305, "label": "bar", "end_id": 281474976710659, "start_id": 281474976710658, "properties": {}}::edge
(1 row)

psql-11.5-5432-pgsql=#
```

That aside, there appears to be an edge case that was overlooked with the combination of the two. That needs to be addressed and resolved.

TropicalPenguin:
[...]

PR #376, opened 2022-12-05
Fix null pointer on name compare

parvit:
Hi all,
this PR resolves issue #369, the fix is a pretty straightforward null pointer check. 

I've also included a dedicated regression script for the issue, please let me know if there is anything you think i should add to the PR.


jrgemignani:
Looks good. Just waiting on Travis CI.

Added @dehowef to double check.

PR #398, opened 2022-12-20
Fix agtype_btree_cmp test in C.UTF-8 locale

df7cb:
When running with LC_COLLATE=C.UTF-8, agtype_btree_cmp returns -32 instead of -1. This is a valid return code, so adjust the test to allow that.

```
--- /home/cbe/projects/postgresql/age/age.git/./regress/expected/agtype.out	2022-12-14 14:42:30.512908027 +0100 +++ /home/cbe/projects/postgresql/age/age.git/./regress/results/agtype.out	2022-12-20 11:24:07.437053056 +0100 @@ -2886,7 +2886,7 @@
 SELECT agtype_btree_cmp('"string"'::agtype, '"string "'::agtype);
  agtype_btree_cmp
 ------------------
-               -1
+              -32
 (1 row)

 SELECT agtype_btree_cmp(NULL, NULL);
```

### DIFF
--- a/regress/expected/agtype.out
+++ b/regress/expected/agtype.out
@@ -2883,7 +2883,7 @@ SELECT agtype_btree_cmp('"string"'::agtype, '"string"'::agtype);
                 0
 (1 row)
 
-SELECT agtype_btree_cmp('"string"'::agtype, '"string "'::agtype);
+SELECT sign(agtype_btree_cmp('"string"'::agtype, '"string "'::agtype)) AS agtype_btree_cmp;
  agtype_btree_cmp 
 ------------------
                -1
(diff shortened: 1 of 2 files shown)


jrgemignani:
@df7cb Thank you for the PR! However, this has already been addressed in the master and PG12 branches -

```
commit dd34ea51d8e22b2008c3c2af9da5968ec2e95eaa
Author: John Gemignani <jrgemignani@gmail.com>
Date:   Wed Dec 14 13:10:07 2022 -0800

    Fix compare_agtype_scalar_values returned result

    Fixed compare_agtype_scalar_values returned result. It used a
    function, varstr_cmp, that wasn't guaranteed to return 1, 0, or -1.
    It was only guaranteed to return >0, 0, or <0. This caused some
    builds to fail during the regression tests, as 1, 0, or -1 were
    expected.
```

df7cb:
Ah ok, I was using the last released version. Should have checked that, sorry!

jrgemignani:
No problem, it happens.

PR #400, opened 2022-12-21
Create Cycle Graph Function

skkk98:
Added Cycle Graph Function in the graph_generation.c file.

### DIFF
--- a/.vscode/c_cpp_properties.json
+++ b/.vscode/c_cpp_properties.json
@@ -1,0 +1,17 @@
+{
+    "configurations": [
+        {
+            "name": "Linux",
+            "includePath": [
+                "${workspaceFolder}/**",
+                "/home/kalyan/Documents/pgsql-11.17/**"
+            ],
+            "defines": [],
+            "compilerPath": "/usr/bin/gcc",
+            "cStandard": "gnu17",
+            "cppStandard": "gnu++17",
+            "intelliSenseMode": "linux-gcc-x64"
+        }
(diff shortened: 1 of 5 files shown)


.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,17 @@
...
+            ],
+            "defines": [],
+            "compilerPath": "/usr/bin/gcc",
+            "cStandard": "gnu17",
+            "cppStandard": "gnu++17",
+            "intelliSenseMode": "linux-gcc-x64"
+        }
+    ],
+    "version": 4
+}
```
GhostVaibhav, line 17: Always be sure, **not** to include the files with your environment details (unless required/specified)

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+    "C_Cpp.errorSquiggles": "Disabled",
+    "files.associations": {
+        "parsenodes.h": "c"
+    }
+}
```
GhostVaibhav, line 6: Same goes for this

.vscode/c_cpp_properties.json
skkk98: Thank you for rectifying it. Should I remove these files and create a new Pull Request?

GhostVaibhav: Just remove the files and commit them here, and the PR would be updated automatically. No need for creating a new PR.

PR #402, opened 2022-12-22
Changed age_split, age_replace, age_sin from STABLE TO IMMUTABLE

Zahid07:
Changed the 3 functions mentioned above to IMMUTABLE, as we are scanning the database to get the parameter of the functions but the functions themselves don't know that.

A **stable** function is a function that always returns the same result for the same input, as long as the state of the database remains the same. Meaning if we don't change the database the answer does not change.

An **immutable** function, which always returns the same result for the same input, regardless of the state of the database. Meaning the answer does not depend on the database rather on the function.

**The age_split Function:**

The age_split functions takes 2 strings as an input and performs and actions and returns the value. The function **does not** go to a database to lookup the answer. It just performs the action that is defined and returns the answer. This way for the same input we have the same answer.

**The age_replace Function:**

The age_replace functions takes 3 strings as an input and performs and actions and returns the value. The function **does not** go to a database to lookup the answer. It just performs the action that is defined and returns the answer. This way for the same input we have the same answer.

**The age_sin Function:**

The age_sin functions takes an integer as an input and performs and actions and returns the value. The function **does not** go to a database to lookup the answer. It just performs the action that is defined and returns the answer. This way for the same input we have the same answer. It is not like we have a database with all the values of **sin** in that database and we search that database for the answer of the given input.

Generally reasons for changing signature of these functions are:

1. These functions do not need to perform database lookup.
2. These functions neither perform database modification.
3. These functions are guaranteed to return the same constant results given the same arguments.


jrgemignani:
Please give a more clear and complete explanation why this is needed for these functions.

Zahid07:
> Please give a more clear and complete explanation why this is needed for these functions.

Changed the comment. Could you tell me if this is sufficient.

jrgemignani:
Please fix your commit log entry for the PR to be descriptive of the change - what was changed and why. Once this is corrected, let us know so that we can review the PR for merging.

Zahid07:
> Please fix your commit log entry for the PR to be descriptive of the change - what was changed and why. Once this is corrected, let us know so that we can review the PR for merging.

Done.

PR #405, opened 2022-12-22
Changed age_id, age_start_id and age_end_id function signatures to IMMUTABLE

markgomer:
Changed age_id(agtype), age_start_id(agtype) and age_end_id(agtype) from STABLE to IMMUTABLE.

These functions only process the arguments passed to them, not doing any database scan. So they can be safely classified as immutable, instead of stable.


jrgemignani:
The differences between STABLE and IMMUTABLE are small and have nothing to do with the database operations. Can you justify this change in the confines of STABLE versus IMMUTABLE?

markgomer:
The case with these functions is that, through the arguments (a vertex or an edge or null), they receive all the data that is needed to give the wanted results, only doing processings that analyze its contents.

By only processing the arguments passed to it, these functions will always return the same results based on these arguments, being classified as IMMUTABLE.

If something would happen along the way (inside the function) that could change the result, like a scan in a modified database, it could be classified as STABLE.

jrgemignani:
You are conflating STABLE with VOLATILE https://www.postgresql.org/docs/current/xfunc-volatility.html

markgomer:
Sorry, I actually meant that, as an example of a stable function, a scan in the database could change its result, because the read data could be different in other calls of the function with the same arguments.

That being said, these functions give results only based on the containers received as arguments (vertexes or edges), and nothing will act to change it in calls with the same arguments.

For example, if I pass an edge with an id of 1, age_id will always look inside that edge container (not the database) and output 1, and in a similar way with the other two functions.

jrgemignani:
@markgomer Please fix your commit log entry for the PR to be descriptive of the change - what was changed and why. Once this is corrected, let us know so that we can review the PR for merging.

markgomer:
@jrgemignani I have updated the commit entry log for your review. Please let me know if these changes are appropriate.

PR #406, opened 2022-12-22
Changed age_head, age_last and age_properties to IMMUTABLE

MatheusFarias03:
Changed age_head(agtype), age_last(agtype) and age_properties from STABLE to IMMUTABLE, since none of them require database lookup and will always return the same result given the same arguments.


jrgemignani:
I feel this is potentially risky as these are execution phase functions that operate on lists and values that may change underneath.

https://www.postgresql.org/docs/current/xfunc-volatility.html - 

"Labeling a function IMMUTABLE when it really isn't might allow it to be prematurely folded to a constant during planning, resulting in a stale value being re-used during subsequent uses of the plan. This is a hazard when using prepared statements or when using function languages that cache plans (such as PL/pgSQL)."

PR #407, opened 2022-12-22
agtype_typecast funcs changed to immutable

maazzzzz:
changed the following three function from stable to immutable

- agtype_typecast_int
- agtype_typecast_numeric
- agtype_typecast_float


jrgemignani:
Could you kindly fully explain why this is needed?

maazzzzz:
based on my current understanding, it would be better from a query planning/optimization standpoint. Since these functions fulfill the criteria required for immutable functions ; tables arent accessed inside them, they dont effect the db and always return the same result for a given input . I think they can be safely marked immutable.

PR #408, opened 2022-12-23
Changed volatility types of age_substring, age_left and age_right 

KamleshKumar427:
Changed volatility types of age_substring, age_left, and age_right to IMMUTABLE from STABLE. 
Since these 3 functions are only performing operations on their inputs parameters and will provide the same result whenever the same parameters are passed.


jrgemignani:
@KamleshKumar427 Your other assertions are correct, however, this one is not -

> They are neither looking up the database nor modifying the database anywhere in the definitions. Which makes these functions IMMUTABLE rather than STABLE.

You are conflating STABLE with VOLATILE. Please read https://www.postgresql.org/docs/current/xfunc-volatility.html

KamleshKumar427:
@jrgemignani you are Right!! I have mixed some statements while commenting.

PR #409, opened 2022-12-23
Changed Volatility of age_cos, age_tan and age_cot from STABLE to IMMUTABLE as they give same result for the same argument values.

Pratikkapsime:
Changes: -
Updated the volatility category of functions age_cos, age_tan, and age_cot from STABLE to IMMUTABLE.

Reasons for modification: -
 1. As these functions only depend on their argument, it doesn't modify the database.
 2. These functions don't do database lookup.
 3. Guaranteed to give the same result for the same argument values.


jrgemignani:
@Pratikkapsime Please fix your commit log entry for the PR to be descriptive of the change. Once this is corrected, let us know so that we can review the PR for merging.

Pratikkapsime:
@jrgemignani  I have updated. Please let me know if you want more changes.

PR #410, opened 2022-12-23
Changed volatility category of age_rtrim,age_ltrim,age_trim

fatimafatimaprogrammer:
### **Changes**

Updated volatility category of functions age_ltrim, age_rtrim, age_trim from STABLE to IMMUTABLE

### **Reasons**

1. These functions will not perform database lookup/scan.
2. These functions will neither do database modification.
3. These functions are guaranteed to return the same constant results given the same arguments.
4. They simply take the input (String), applies operation on it and returns the result

### **Description**

**age_ltrim [age left trim function]:**
eliminates left leading and trailing spaces on string should be immutable because this function cannot modify the database and is guaranteed to return the same results given the same arguments forever. This immutable category allows the optimizer to pre-evaluate the function when a query calls it with constant arguments.
**age_rtrim [age right trim function]:**
eliminates right leading and trailing spaces on string should be immutable because this function cannot modify the database and is guaranteed to return the same results given the same arguments forever. This immutable category allows the optimizer to pre-evaluate the function when a query calls it with constant arguments.
**age_trim [age trim function]:**
eliminates right leading and trailing spaces on string should be immutable because this function cannot modify the database and is guaranteed to return the same results given the same arguments forever. This immutable category allows the optimizer to pre-evaluate the function when a query calls it with constant arguments.

image


jrgemignani:
Please fix your commit log entry for the PR to be descriptive of the change. Simply stating "immutable" isn't enough. Once this is corrected, let us know so that we can review the PR for merging.

fatimafatimaprogrammer:
> @fatimafatimaprogrammer Please fix your commit log entry for the PR to be descriptive of the change - what the change is and why. Simply stating "immutable" isn't enough. Once this is corrected, let us know so that we can review the PR for merging.

I've Updated PR description in a new commit let me know if it needs any further changes

PR #411, opened 2022-12-23
Changed age_sqrt and age_floor functions volatililty categories

imranzaheer612:
Changed function volatility categories of **age_srt()** and **age_floor()** from STABLE to **IMMUTABLE**

Why age_srt() and age_floor() are Immutable?

- Given the same input, the result will never gonna change no matter what.
- Both are pure functions & the results only depends on their input values.
- Are mathematical functions that will always return the same constant value given the same arguments.

Why age_srt() and age_floor() are not stable?

- For a STABLE function “*given the same arguments result could change across SQL statements*”. But in this case, the result will never gonna change for the same arguments even across multiple SQL statements.


jrgemignani:
@imranzaheer612 Please fix your commit log entry for the PR to be descriptive of the change - what was changed and why. Once this is corrected, let us know so that we can review the PR for merging.

imranzaheer612:
> @imranzaheer612 Please fix your commit log entry for the PR to be descriptive of the change - what was changed and why. Once this is corrected, let us know so that we can review the PR for merging.

Commit log updated. Let me know if still need to change.

jrgemignani:
@imranzaheer612 I probably should have shown you what I meant -

image

Please make sure the first part is a reasonable, but short, title. Then add in the details in the extended description. Remember that we don't want the line length going beyond 80 characters. This is due to legacy terminals.

imranzaheer612:
@jrgemignani I updated that. You can take a look now.

PR #412, opened 2022-12-23
Changed volatility category of age_atan2, age_degrees, age_radians

Hunaid2000:
**Changes**
- Updated volatility category of functions age_atan2, age_degrees, age_radians from STABLE to IMMUTABLE

**Reasons**
- These math functions will not scan/lookup database
- They simply take the input (int, float or numeric), applies operation on it and returns the result
- They cannot modify database
- They are guaranteed to return the same results given the same arguments forever


jrgemignani:
@Hunaid2000 Please fix your commit log entry for the PR to be descriptive of the change - what the change is and why. Once this is corrected, let us know so that we can review the PR for merging.

PR #413, opened 2022-12-23
Updated age_type, age_exists, and age_label category to IMMUTABLE

muneebkhan4:
These functions satisfy the condition for Immutable functions because:

- They do not modify database
- Returns the same result for same arguments forever
- Their result will remain the same for whole Query. We can prepare the result of these functions beforehand and their result will remain the same. It will not harm the expected results.

The snapshotting behavior is used for SELECT commands within IMMUTABLE functions. It is generally unwise to select from database tables within an IMMUTABLE function at all, since the immutability will be broken if the table contents ever change. However, PostgreSQL does not enforce that you do not do that. In these functions, we will not be using select from database tables because they do not essential need to do any scan.


muneebkhan4 (commented):
These functions satisfy the condition for Immutable functions like it do not modify database, do not do database look up and returns the same result for same argument vales. They do not need to have database look up so, they should not be Stable.

jrgemignani:
Please read https://www.postgresql.org/docs/current/xfunc-volatility.html 
Specifically, the differences between IMMUTABLE and STABLE and then VOLATILE versus IMMUTABLE and STABLE. What you state as a justification is not correct.

muneebkhan4:
> Please read https://www.postgresql.org/docs/current/xfunc-volatility.html Specifically, the differences between IMMUTABLE and STABLE and then VOLATILE versus IMMUTABLE and STABLE. What you state as a justification is not correct.

I have gone through the documentations again. I have updated the reasoning. Please check it.

jrgemignani:
@muneebkhan4 I am not sure, due to the nature of AGE and cypher, that these specific functions meet all of the requirements for IMMUTABLE due to this -

> It is generally unwise to select from database tables within an IMMUTABLE function at all, since the immutability will be broken if the table contents ever change.

While it is true that they act on the values passed to them, what if those values are passed by a select, or subselect? Can you guarantee that this won't happen?

muneebkhan4:
> @muneebkhan4 I am not sure, due to the nature of AGE and cypher, that these specific functions meet all of the requirements for IMMUTABLE due to this -
> 
> > It is generally unwise to select from database tables within an IMMUTABLE function at all, since the immutability will be broken if the table contents ever change.
**We need to select from database tables within implementation of these functions. Given below are the functionalities that I understood from these functions and my updated decisions.**
> 
> While it is true that they act on the values passed to them, what if those values are passed by a select, or subselect? Can you guarantee that this won't happen?
**Yes, the values passed to them can be from select or subselect queries.** 

**Functionalities:**
**`age_type`** tells the type of the given (edge/ vertex) in function's argument. Because of following reasons:
**Labeling a function IMMUTABLE when it really isn't might allow it to be prematurely folded to a constant during planning, resulting in a stale value being re-used during subsequent uses of the plan. This is a hazard when using prepared statements or when using function languages that cache plans (such as PL/pgSQL).**
In the above function, the result depends on the edge provided and it can definitely change in future. So, it will not be good if we make this function IMMUTABLE.

**`age_exists`** (is not documented in apache age website in scalar functions section). I think it will tell if the given age_type in function's argument exists in our database or not.
Again, we cannot make it IMMUTABLE because its result depends on the result depends on the edge provided and it can definitely change in future.

**`age_label`** (is not documented in apache age website in scalar functions section). I think it tells the label associated with the given age_type(edge or vertex) in function's argument.
Again, we cannot make it IMMUTABLE because its result depends on the result depends on the edge provided and it can definitely change in future.

**Updated Decision:**
`age_type`, `age_label`, and `age_exists` all these functions are not good to be declared as IMMUTABLE because of the above reasons. The category that best suits these functions is **`Stable`**.

rafsun42:
To clarify the functionalities:
`age_type` returns the label of an edge. Throws error if its argument is not an ede. 

`age_exists` returns whether a property exists. For example, `CREATE (a {name: 'ABC'}) RETURN exists(a.name), exists(a.age)` returns true and false respectively.

muneebkhan4:
@jrgemignani what do you thinking about it.

muneebkhan4:
> To clarify the functionalities: `age_type` returns the label of an edge. Throws error if its argument is not an edge.
> 
> `age_exists` returns whether a property exists. For example, `CREATE (a {name: 'ABC'}) RETURN exists(a.name), exists(a.age)` returns true and false respectively.

Still I think functions should be `stable` because of the above reasons.

jrgemignani:
@rafsun42 Do you think these would be okay or not? Remember, the volatility refers to the output based on the input. Personally, I feel it is safer to leave most non-numeric or non-string based functions as STABLE.

rafsun42:
@jrgemignani @muneebkhan4 I agree with you. Immutable functions output same result for same arguments forever. But, properties and labels of an entity can change. So, these three functions should not be immutable. Also, since these are read-only functions, I believe stable is the best option for them.

PR #467, opened 2022-12-29
CHANGED: ABS, SIGN, LOG -> IMMUTABLE

aru-d-at:
The functions were marked as STABLE. However, they can be changed to IMMUTABLE.

REASON:
* The functions designated are not conducting a database lookup. Rather, the value passed in as an argument is used by the function to determine the type and value to return.
* No mutation occurs in the course of using the functions.
* Possible function arguments are taken into consideration during function implementation to be within a certain range. Thus, the return values are to be as expected during usage.


rafsun42:
Thank you! But these changes were already merged as a part of a separate PR.

PR #468, opened 2022-12-29
Update SET clause to support assigning a map to a variable

rafsun42:
For example, 'SET v = {..}' will remove all properties of v and set the provided map as its properties.


jrgemignani:
@rafsun42 It looks like you need to rebase this PR against the master branch.

PR #582, opened 2023-01-20
Added some comments for create_graph function

intrigued-rishi:


jrgemignani:
Thank you for adding in comments! My only issue is that the comments should ideally be before the line, not following it on the same line. The exception would be for structs and enums. Please correct those.

intrigued-rishi:
Corrected them

PR #587, opened 2023-01-23
Implement CI test for python driver

MuhammadTahaNaveed:
### What I did
- Implemented a CI workflow for python driver that ensures that any PR/commit that modifies files under age/drivers/python is passing the tests.
- Moved docker-compose.yaml file in age/drivers/golang to age/drivers, because this is required to instantiate a postgres instance needed for unit testing.
- Modified host address of connection params to localhost in 'test_age_py.py' to connect to postgres instance running in docker container.

### Additional Context
This workflow will only run if their is a modification in files under age/drivers/python.


MuhammadTahaNaveed:
Ok 😅. Is their any forum or something like that where Apache informs about these things ?

jrgemignani (approved):

jrgemignani:
> Ok 😅. Is their any forum or something like that where Apache informs about these things ? Btw, this workflow is not using travis, it is using github actions.

They informed us via email. Our options are Github actions or Jenkins. Jenkins supports ARM whereas Github actions doesn't.

PR #632, opened 2023-01-28
update title and add "more information portion" in readme of drivers/nodejs

abduu11aahh:
Synchronize the title of drivers/nodejs readme file with the titles of other drivers readme files.
And add "more information portion" in its readme.


jrgemignani:
@abduu11aahh Unfortunately, someone else fixed this #642. So, it is no longer necessary.

PR #633, opened 2023-01-28
Adding some informative comments for Label creation functions

intrigued-rishi:


jrgemignani:
@intrigued-rishi This is not to discourage anyone from adding comments. However, I don't see any of **these comments** as really being necessary. It is already pretty obvious what those functions do, just by their names. Meaning, this code is self documenting.

Can you justify the need for these comments?

PR #637, opened 2023-01-30
Added documentation/comments in index.ts file and enhanced readability.

JawadAliAbbasi:
Added comments for and within functions to explain their functionality.


jrgemignani:
@JawadAliAbbasi Can you justify why any of the added documentation is actually necessary? The code seems rather self explanatory. Additionally, your descriptions should strive to make it easier for the reviewer. Is this core code? Is this a driver? If so, which driver? etc.

PR #638, opened 2023-01-30
Implement plus-equal operator in SET clause

rafsun42:
Additional changes:
Null properties are removed from cypher maps in certain cases. A new property `cypher_map.keep_null` is added. It is set while transforming CREATE, MATCH and SET clause. Based on its value, transform_cypher_map tells executor whether to keep the null properties in maps that appear in the mentioned clauses.

Cases of null removal:
  - For CREATE and SET with equal operator, nulls are removed.

  - For MATCH, nulls are not removed.

  - For SET with plus-equal operator, nulls are kept during transformation, but removed after execution.

  - For any other cases, nulls are removed by default.


src/include/nodes/cypher_nodes.h
```diff
@@ -195,6 +195,7 @@ typedef struct cypher_map
     ExtensibleNode extensible;
     List *keyvals;
     int location;
+    int keep_null; // if false, keyvals with null value are removed
```
jrgemignani, line 198: Why is this an integer and not a Boolean? It is preferable to use a Boolean type and true/false instead of 1/0.

rafsun42: @jrgemignani My bad. Fixed it now.

jrgemignani: Using an integer in place of a Boolean can be okay in some instances. I was just wondering because it looked like it was only ever used as a Boolean.

jrgemignani (approved):
Looks good

PR #646, opened 2023-02-01
Fix WITH ignoring WHERE clause

### LINKED ISSUE #238: WITH ignores WHERE condition
**Describe the bug**
Using a WHERE after WITH after MATCH ignores condition

-pgsql inside docker container

Using the sample city/country data in age/regress/age_load/data/

pgsql 

MATCH  (m:City) - [:has_city] -> (b:Country{iso2:'BE'})   with b,m where m.name='x'  return m.name,b.iso2 $$)  as (  "m.name" agtype, "b" agtype);

returns 550 rows

**Expected behavior**
Should return 0 rows due to m.name='x' WHERE condition on WITH

**Environment :**
- Version: latest docker image apache/age:latest 

**Additional info**
With/Where Ref: https://neo4j.com/docs/cypher-manual/current/clauses/where/

### LINKED ISSUE #329: Server crash in attempt to return undefined variable from MATCH
**Describe the bug**
Using the sample commands/test program from https://github.com/apache/age/issues/328, if I replace the `RETURN c,deps` sub-clause with `RETURN c,d`

In that program, this query prompts the following error:
`panic: read tcp 127.0.0.1:53392->127.0.0.1:5434: read: connection reset by peer`

And if I try a command in psql soon afterwards:
```
WARNING:  terminating connection because of crash of another server process
DETAIL:  The postmaster has commanded this server process to roll back the current transaction and exit, because another server process exited abnormally and possibly corrupted shared memory.
HINT:  In a moment you should be able to reconnect to the database and repeat your command.
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Succeeded.
```

- Golang Driver, psql

See https://github.com/apache/age/issues/328

- N/A

```
MATCH (a:A) WHERE ID(a)=0 WITH a OPTIONAL MATCH (a)-[:incs]->(c)-[d:incs]-() WITH a,c,COUNT(d) AS deps WHERE deps<=1 RETURN c,d
```

**Expected behavior**
An error stating that the variable d is undefined (since only 'deps' is preserved by the WITH, not d itself), rather than closing the connection.

- Version: 1.0.0

**Additional context**
N/A

### LINKED ISSUE #351: why can't "WHERE" work with 'MATCH/WITH' ?
I tried the "WITH"  cluse as described by the doc [Filter on aggregate function results](https://age.apache.org/age-manual/master/clauses/with.html#filter-on-aggregate-function-results). But, I found that the condition presented by 'WHERE' can't work with 'WITH'.

Example Code:
`select * from cypher('agload_test_graph', $$ match (n) with n where n.name='A' return n $$) as (sp agtype)`

the `n.name='A'` doesn't work.

If "WITH" is removed, 'WHERE' can then work:
`select * from cypher('agload_test_graph', $$ match (n) where n.name='A' return n $$) as (sp agtype)`

### LINKED ISSUE #394: Filtering does not work inside with clause
- Command line

```pgsql
 WITH 1 as n 
 WHERE n > 1
 RETURN n 
$$) as (n agtype);
```

**Expected behavior**
It should return nothing, it is returning 1 row.

- Version: 1.1.0

### LINKED ISSUE #396: Accessing variables not included in WITH clause terminates the server
**Describe the bug**
Returning any variable not included in `with clause` terminates the server

Command line

```pgsql
...
SELECT * FROM cypher('new_issue',$$
     CREATE (v),(u),(w),
            (v)-[:hasFriend]->(u),
            (u)-[:hasFriend]->(w)
$$) as (a agtype);
...
```

```pgsql
SELECT * FROM cypher('new_issue',$$
      MATCH p=(v)-[*1..2]->(u) 
      WITH p,length(p) AS path_length 
      RETURN v,path_length
$$) as (a agtype,b agtype);
```
```
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
```

**Expected behavior**
I know that I should only be accessing the variables defined in the scope of WITH clause, but I don't think accessing others should terminate the server.

- Version: 1.1.0

MuhammadTahaNaveed:
- When transforming `cypher_with` into `cypher_return` clause, **where** node of `cypher_with` was lost since cypher_return has no where node and since the function transform_cypher_clause_with_where was only given the wrapper clause(which had transformed cypher_return from cypher_with) as parameter, the where node was lost.
- Changed the function header of transform_cypher_clause_with_where to explicitly accept where node as parameter.
- Added regression tests for cypher_with.

resolves #238, resolves #351, resolves #394 

### Update

Fix segmentation fault when referencing undefined variables in WITH clause.

resolves #329, resolves #396

### DIFF
--- a/regress/expected/cypher_with.out
+++ b/regress/expected/cypher_with.out
@@ -16,8 +16,12 @@
  * specific language governing permissions and limitations
  * under the License.
  */
+ \! cp -r regress/age_load/data regress/instance/data/age_load
 LOAD 'age';
 SET search_path TO ag_catalog;
+--
+-- Load data
+--
 SELECT create_graph('cypher_with');
 NOTICE:  graph "cypher_with" has been created
  create_graph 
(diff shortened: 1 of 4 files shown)


MuhammadTahaNaveed:
### Update

Fix segmentation fault when referencing undefined variables in WITH clause.

Now also Resolves
- #329
- #396

jrgemignani:
@MuhammadTahaNaveed Please review my comment above.

MuhammadTahaNaveed:
> @MuhammadTahaNaveed Please review my comment above.

Hi john,
Which one are you reffering to? I don't see any comment.

jrgemignani:
@MuhammadTahaNaveed Is there a way to get this PR to run the new CI tests? It still lists Travis CI.

jrgemignani:
@MuhammadTahaNaveed Is there a reason why you added an extra parameter to pass self->where instead of the way it was done?

MuhammadTahaNaveed:
> @MuhammadTahaNaveed Is there a way to get this PR to run the new CI tests? It still lists Travis CI.

Yes, sure.

> @MuhammadTahaNaveed Is there a reason why you added an extra parameter to pass self->where instead of the way it was done?

Yes. When `cypher_with` is converted into `cypher_return` in function `transform_cypher_with`, the where node of `cypher_with` gets lost because there is no node for where in `cypher_return` struct. 

So, in order to transform the where node of `cypher_with`, we need to explicitly pass it to `transform_cypher_clause_with_where` because the self passed to it is cypher_return (that was converted from cypher_with).

I hope this justifies the reason.

dehowef:
Looks good to me.

PR #647, opened 2023-02-02
[Label Inheritance] Added inheritance to create_vlabel and create_elabel functions 

MatheusFarias03:
Added inheritance functionality to `create_vlabel` so now it is possible to a vertex label to have a parent label other than `_ag_vertex_label`. To do this, it is necessary to have an existing parent label and then call the function as:
```sql
SELECT create_vlabel('graph_name', 'label_name', 'parent_label_name');
```
Lets say we create a parent label:
```sql
SELECT create_vlabel('demo', 'Book');
NOTICE:  VLabel "Book" has been created

 create_vlabel
---------------

(1 row)
```

And then create a label that inherits from this "Book" label:
```sql
SELECT create_vlabel('demo', 'Fantasy', 'Book');
NOTICE:  VLabel "Fantasy" has been created

 create_vlabel
---------------

(1 row)
```

Creating a vertex with the "Fanstasy" label:
```sql
SELECT * FROM cypher ('demo', $$
CREATE (v:Fantasy)
RETURN v
$$) as (vertex agtype);
                                 vertex
------------------------------------------------------------------------
 {"id": 1125899906842625, "label": "Fantasy", "properties": {}}::vertex
(1 row)
```

The parent label "Book" will also have this vertex:
```sql
SELECT * FROM demo."Book";
        id        | properties
------------------+------------
 1125899906842625 | {}
(1 row)

SELECT * FROM demo."Fantasy";
        id        | properties
------------------+------------
 1125899906842625 | {}
(1 row)
```

### DIFF
--- a/age--1.2.0.sql
+++ b/age--1.2.0.sql
@@ -102,7 +102,7 @@ CREATE FUNCTION ag_catalog.create_vlabel(graph_name name, label_name name, paren
     LANGUAGE c
 AS 'MODULE_PATHNAME';
 
-CREATE FUNCTION ag_catalog.create_elabel(graph_name name, label_name name)
+CREATE FUNCTION ag_catalog.create_elabel(graph_name name, label_name name, parent_list name[] = NULL)
     RETURNS void
     LANGUAGE c
 AS 'MODULE_PATHNAME';
--- a/age--1.2.0.sql
+++ b/age--1.2.0.sql
@@ -97,7 +97,7 @@ RETURNS void
 LANGUAGE c
 AS 'MODULE_PATHNAME';
 
-CREATE FUNCTION ag_catalog.create_vlabel(graph_name name, label_name name)
+CREATE FUNCTION ag_catalog.create_vlabel(graph_name name, label_name name, parent_label_name name = NULL)
     RETURNS void
     LANGUAGE c
 AS 'MODULE_PATHNAME';
--- a/age--1.2.0.sql
+++ b/age--1.2.0.sql
@@ -97,7 +97,7 @@ RETURNS void
 LANGUAGE c
 AS 'MODULE_PATHNAME';
 
-CREATE FUNCTION ag_catalog.create_vlabel(graph_name name, label_name name)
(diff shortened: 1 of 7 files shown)


MatheusFarias03:
Changed a little bit on how the inheritance works. It is passed to the `label_name` parameter as `child_name:parent_name` separated by a ` : ` . It then checks if the parent name doesn't exists and if the child name already exists.

```sql
SELECT create_vlabel('graph_name', 'child_name:parent_name');
NOTICE: VLabel "child_name" has been created
 child_name
---------------

(1 row)
```

MatheusFarias03:
The function can receive two or three arguments, and the third argument is set to NULL so that it is optional to inform the array of parent label it inherits from :
```sql
--- This is found at the age--1.2.0.sql file , line 100.

CREATE FUNCTION ag_catalog.create_vlabel(graph_name name, label_name name, parent_list name[] = NULL)
    RETURNS void
    LANGUAGE c
AS 'MODULE_PATHNAME';
```

Calling the function with just two arguments works fine, but when calling with the third argument as a array throws this error:
```sql
SELECT * FROM ag_catalog.create_vlabel('demo', 'Programmer', ARRAY['Person']);

--- NOTICE:  VLabel Programmer will inherit from Person
--- NOTICE:  merging multiple inherited definitions of column "id"
--- NOTICE:  merging multiple inherited definitions of column "properties"
--- ERROR:  column "id" inherits conflicting default values
--- HINT:  To resolve the conflict, specify a default explicitly.

```

MatheusFarias03:
`create_vlabel` and `create_elabel` functions are both practically done. We need to edit test case files under the `regress` folder. I believe when we run installcheck it is going to store the results in the `regress/results` folder. Then we copy the results we got to the `regress/expected` folder.

The `create_vlabel` and `create_elabel` functions both work like this: 
```sql
SELECT * FROM ag_catalog.create_vlabel('graph_name', 'child_label_name', ARRAY['parent1', 'parent2', 'parent3']);
```

PR #654, opened 2023-02-07
Added Comments on functions which needed comments in ag_label.c and label_commands.c

intrigued-rishi:


jrgemignani:
Closed as unnecessary. This is a repackage of #633 with additional comments that aren't necessary.

PR #667, opened 2023-02-12
Add function to detect cypher queries

moeed-k:
-Initial function that detects a cypher call in a FROM clause -Recognizes statements of the format 'SELECT * FROM cypher(), cypher().....'


uhayat:
Looks good.

PR #715, opened 2023-02-24
Changed events that triggers the workflow of drivers

Hunaid2000:
Made changes to events so that workflow only runs for the changed driver.


jrgemignani:
@Hunaid2000 Thank you, however, the driver tests need to run when a change is made to the core code. This is because core code changes can impact the drivers.

PR #725, opened 2023-03-03
Update regression tests for cypher_with

MuhammadTahaNaveed:
- Used a reasonable dataset for regression tests.


jrgemignani:
Built and ran PR locally; looks good.

PR #731, opened 2023-03-05
Update README.md

Zainab-Saad:
**Update Quickstart section**

- Add another query where creating a vertex with properties and label is demonstrated
- Add "values can be expressions" in the explanation of creating edge with its properties

**Modify explanation**
- Remove "values can be expressions" in the explanation of creating edge with its properties

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@
 
 </br>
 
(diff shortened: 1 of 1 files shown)


dehowef (commented):
Looks good, except the change at Line 265( 274 in the proposed README). That note may be a bit redundant. Can you remove it? @Zainab-Saad

Zainab-Saad:
@dehowef  I have removed the redundant statement.

Zainab-Saad:
@dehowef I have squashed the commits. Let me know if I need to do any further changes.
Thanks

PR #732, opened 2023-03-05
[AGE-Viewer-Desktop-Go] removed node modules, agensgraph and refactor code

sarthak-19:
Removed node modules as it should not be present with the source code in repositories and clean the old redundant code of agensgraph


sarthak-19:
[ ]  go.work should be used to manage nested modules of golang as there are two projects in the directory for the same  work.

PR #736, opened 2023-03-07
Remove check for scalar(vertex,edge or path) agtypes in unwind

MuhammadTahaNaveed:
- Removed check that was disabling unwind to be used on arrays with elements of type vertex,edge or path. 
- Added regression tests for unwind.

### DIFF
--- a/age--1.2.0.sql
+++ b/age--1.2.0.sql
@@ -4162,10 +4162,10 @@ IMMUTABLE
 PARALLEL SAFE
 AS 'MODULE_PATHNAME';
 
-CREATE FUNCTION ag_catalog.age_unnest(agtype, block_types boolean = false)
-    RETURNS SETOF agtype
-    LANGUAGE c
-    IMMUTABLE
+CREATE FUNCTION ag_catalog.age_unnest(agtype)
+RETURNS SETOF agtype
+LANGUAGE c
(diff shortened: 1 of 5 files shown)


MuhammadTahaNaveed:
Hi @emotionbug,

I did not modify those after seeing this comment. 

What do you think, should it be modified?

emotionbug:
I think that variable is for the constraint you removed.
Since you've already determined that the constraint is unnecessary and removed it, I think the comments and code around it should also be modified.

Because even if you don't, it's already doing nothing with the variable, and it's not what the comment says.

MuhammadTahaNaveed:
> I think that variable is for the constraint you removed.
> Since you've already determined that the constraint is unnecessary and removed it, I think the comments and code around it should also be modified.
> 
> Because even if you don't, it's already doing nothing with the variable, and it's not what the comment says.
> 

Ok thanks. I will update the PR.

jrgemignani:
@MuhammadTahaNaveed I'm not sure why your comments say Co-authored-by you?

MuhammadTahaNaveed:
> @MuhammadTahaNaveed I'm not sure why your **commit description** comments say Co-authored-by you?

@jrgemignani Fixed it.

PR #739, opened 2023-03-08
Uploading centos7 spec file for postgress age rpm

khurramlatif91:
Added spec file to create postgress age rpm on centos7

### DIFF
--- a/centos7.spec
+++ b/centos7.spec
@@ -12,7 +12,7 @@ Source0:        %{name}-%{version}.tar.gz
 BuildRoot:      %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
 
 %description
-This package provides extensions for Apache Age
+Apache AGE is an extension for PostgreSQL that enables users to leverage a graph database on top of the existing relational databases.
 
 BuildRequires:  gcc glibc glib-common readline readline-devel zlib zlib-devel flex bison  
 BuildRequires: centos-release-scl epel-release postgresql%{pgmajorversion} postgresql%{pgmajorversion}-devel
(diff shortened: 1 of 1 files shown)


centos7.spec
```diff
@@ -0,0 +1,37 @@
...
+Version:        1
+Release:        %{dist}
+License:        Apache
+Group:          Databases
+Url:            https://github.com/apache/age.git
+Source0:        %{name}-%{version}.tar.gz
+BuildRoot:      %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
+
+%description
+This package provides extensions for Apache Age
```
uhayat, line 15: Add a bit details in description like bellow:
Apache AGE is an extension for PostgreSQL that enables users to leverage a graph database on top of the existing relational databases.

centos7.spec
```diff
@@ -0,0 +1,37 @@
...
+%install
+cp -p age.so %{pginstdir}/lib/age.so
+cp -p age--1.1.1.sql %{pginstdir}/share/extension/age--1.1.1.sql
+cp -p age.control %{pginstdir}/share/extension/age.control
+
+%files
+
+%changelog
+* Fri Feb 17 2023 - Khurram Latif <Khurram.L@agedb.io>
+- Initial packaging for AEP
```
uhayat, line 37: Fix type, Use AGE instead of AEP

uhayat (changes requested):

centos7.spec
khurramlatif91: Updated description and changelog comment as per your suggestion

uhayat (approved):
Looks good, ready to merge.

PR #741, opened 2023-03-12
Moved create_empty_agtype() outside of loops

mohayu22:
The statement `props = create_empty_agtype()` was being executed inside loops, which was causing unnecessary overhead. It has been moved outside of the loops, so that it is only executed once.

### DIFF
--- a/src/backend/utils/graph_generation.c
+++ b/src/backend/utils/graph_generation.c
@@ -194,13 +194,14 @@ Datum create_complete_graph(PG_FUNCTION_ARGS)
 
     vtx_seq_id = get_relname_relid(vtx_seq_name_str, nsp_id);
     edge_seq_id = get_relname_relid(edge_seq_name_str, nsp_id);
+  
+    props = create_empty_agtype();
 
     /* Creating vertices*/
     for (i=(int64)1;i<=no_vertices;i++)
     {   
         vid = nextval_internal(vtx_seq_id, true);
         object_graph_id = make_graphid(vtx_label_id, vid);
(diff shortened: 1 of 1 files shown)


jrgemignani:
@mohayu22 Please explain, in detail, **why** this is necessary.

mohayu22:
All vertices and edges of complete graph are created with the same properties. However, we are currently generating new _props_  using `create_empty_agtype` for every vertex and edge, although we only need to generate it once. 

It may appear that if we used the same reference to _props_ for all the vertices and edges, then they will share the same allocated memory, but that is not the case. Later in the call stack, the _props_ variable is passed to the `PointerGetDatum` alias `AGTYPE_P_GET_DATUM` function, which casts it into a Datum value and it is then stored and used to create vertices and edges.

To optimize speed and memory, it is recommended that we call `create_empty_agtype` only once and reuse the same _props_ for all vertices and edges.

mohayu22:
In `insert_vertex_simple` and `insert_edge_simple` functions that are called inside the loops

jrgemignani:
@mohayu22 I'm sorry, I should be more direct. I want you to thoroughly explain why. Part of this is because, with allocated data, we need to be very careful about passing the same object around. The other part of this is, we need to see that you were thorough and the level of detail you used to look into this.

jrgemignani:
@mohayu22 Please respond to my last inquiry and update your branch as there are conflicts.

mohayu22:
@jrgemignani I will redo this PR

Issue #701, opened 2023-02-22
MATCH (A) MATCH (A {prop: 'val'} ) ... Fails when it should not.


dehowef:
**Describe the bug**

There is unintended behavior when a user tries to MATCH a node with property constraints after already MATCHing a node with the same variable name.

**How are you accessing AGE (Command line, driver, etc.)?**

Command Line

**What data setup do we need to do?**

Have AGE running.

**What is the necessary configuration info needed?**
- N/A

**What is the command that caused the error?**
```pgsql
SELECT * FROM cypher('my_graph', $$ MATCH (a) MATCH (a {prop:'val'}) RETURN a $$ ) AS (a agtype);

```
```
postgres=# SELECT * FROM cypher('my_graph', $$ MATCH (a) MATCH (a {prop:'val'}) RETURN a $$ ) AS (a agtype);
2023-02-22 01:29:05.230 PST [59866] ERROR:  missing FROM-clause entry for table "a" at character 34
2023-02-22 01:29:05.230 PST [59866] STATEMENT:  SELECT * FROM cypher('pilots', $$ MATCH (a) MATCH (a {s:'as'}) RETURN a $$ ) AS (a agtype);
ERROR:  missing FROM-clause entry for table "a"
LINE 1: SELECT * FROM cypher('my_graph', $$ MATCH (a) MATCH (a {prop:'val'...
```

**Expected behavior**

The program should detect the variable and return the according result. The RTE is not being detected and the program is throwing an error.

**Environment (please complete the following information):**
- Version: PG11 Version 1.1.X, Fedora OS 36

**Additional context**

dehowef:
The issue seems to lie in transform_cypher_node() in cypher_clause.c
--

Because of how previously transformed match clauses are stored in the RTE, they are undetectable by create_property_constraints(). All of previous clauses are pooled into an RTE with the alias "_', and the find_rte() logic responsible detecting previously declared variables cannot produce property_constraints.

jrgemignani:
@dehowef I checked and this issue goes back to, at least, 1.0.0, that was as far back as I went. It is also in PG12 and PG13 versions. So, it isn't anything new and was likely just overlooked as a case.

PR #747, opened 2023-03-16
Fix property constraints against resolved variables (#701)

jrgemignani:
Fixed an issue where an already resolved variable, when used for
a property constraint, errored out. See #701 for more details.

This also fixed a regression test that was erroring out, but was
overlooked.

Adjusted and added additional regression tests.

Co-authored-by: Dehowe Feng <dehowefeng@gmail.com>

### DIFF
--- a/regress/expected/cypher_match.out
+++ b/regress/expected/cypher_match.out
@@ -1451,10 +1451,232 @@ $$) as (n agtype);
 (1 row)
 
 --
+-- Regression tests to check previous clause variable refs
+--
+-- set up initial state and show what we're working with
+SELECT * FROM cypher('cypher_match', $$
+    CREATE (a {age: 4}) RETURN a $$) as (a agtype);
+                                   a                                    
+------------------------------------------------------------------------
+ {"id": 281474976710665, "label": "", "properties": {"age": 4}}::vertex
+(1 row)
+
+SELECT * FROM cypher('cypher_match', $$
+	CREATE (b {age: 6}) RETURN b $$) as (b agtype);
+                                   b                                    
+------------------------------------------------------------------------
+ {"id": 281474976710666, "label": "", "properties": {"age": 6}}::vertex
+(1 row)
+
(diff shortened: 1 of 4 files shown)


dehowef:
John and I worked closely on this issue to resolve the limitations of the match clause in reflexive pattern matching. If anyone wants to take a look and review what was done, it would be much appreciated

MuhammadTahaNaveed:
Hi @dehowef @jrgemignani,

I was wondering why the query mentioned in issue still errors out. 
```
  MATCH (a {prop: “val”})-[]- (a {prop2: “val2”})
$$) as (a agtype);
```
I tried this query in regression tests
```
SELECT * FROM cypher('cypher_match', $$
	MATCH (a {i: 1})-[]-(a {j: 2}) RETURN a $$) as (a agtype);
ERROR:  variable a already exists
LINE 2:  MATCH (a {i: 1})-[]-(a {j: 2}) RETURN a $$) as (a agtype);
```

jrgemignani:
@MuhammadTahaNaveed Actually, that's a good point and a bit of an oversight on my part. My PR fixes the clause to clause previously resolved variables, not within the same clause. I will try to add that in tomorrow and update this PR.

dehowef:
@MuhammadTahaNaveed The issues with MATCH occur in several sections of the code-- we have chosen to address them incrementally for maximum transparency and visibility for the project.

I feel like this method as several advantages, being that each commit will be itemized so that in the event that something is updated, we can go back to the commit where a certain change was implemented to easily view and address it. Thanks!

jrgemignani:
@dehowef @MuhammadTahaNaveed I'm reviewing my patch to make sure it covers all the cases adequately.

jrgemignani:
@dehowef @MuhammadTahaNaveed I have updated the PR to point to the correct issue this PR resolves. If you can verify that it works against that issue? Sorry, for the confusion.

PR #748, opened 2023-03-16
[GDBMS_cypher] randomUUID() and jsonb_exists() implementation

asadimtiazmalik:
- [ ] I have added the randomUUID() function. Please review my PR and suggest any changes that I can make to the code. Thank you :)


asadimtiazmalik:
I have added exists function to the `cypher_func.c`. Please review and suggest any changes necessary. 

**Use Case:**
`MATCH (n) WHERE exists(n.surname) RETURN n.name AS name, n.surname AS surname`

Issue #724, opened 2023-03-03
MATCH property constraints fail when a variable is referenced more than once.


dehowef:
**Describe the bug**
Property Constraints throw an error when they are applied to a variable mentioned more than once.

**How are you accessing AGE (Command line, driver, etc.)?**
Commandline through postgresql.

**What data setup do we need to do?**
None

**What is the necessary configuration info needed?**
- Install Postgresql
- Install AGE

**What is the command that caused the error?**
```pgsql
SELECT * from cypher('my_graph_name', $$
  MATCH (a {prop: “val”})-[]- (a {prop2: “val2”})
$$) as (a agtype);
```
```
postgres=# SELECT * FROM cypher('pilots', $$ MATCH(a{prop:"val"})-[]-(a{prop2:"val2"}) RETURN a $$ ) AS (a agtype);
2023-03-03 09:05:16.444 PST [145400] ERROR:  variable a already exists at character 59
2023-03-03 09:05:16.444 PST [145400] STATEMENT:  SELECT * FROM cypher('pilots', $$ MATCH(a{prop:"val"})-[]-(a{prop2:"val2"}) RETURN a $$ ) AS (a agtype);
ERROR:  variable a already exists
LINE 1: ... FROM cypher('pilots', $$ MATCH(a{prop:"val"})-[]-(a{prop2:"...
```

**Expected behavior**
The property constraint must be evaluated.

**Environment (please complete the following information):**
- Version: 1.2.0

Zainab-Saad:
This probably is not the bug rather the MATCH clause using same variable for the two vertices should be used to match _self-loop_ on a vertex. `MATCH(a)-[]-(a)` and `MATCH(a {prop: "value"})-[]-(a)` would work fine as they are used for matching the _self-loop_ but `MATCH(a {prop :"value1")-[]-(a {prop : "value2"})` or `MATCH(a {prop1 : "value1"})-[]-(a {prop2 : "value2"})` would through an error below as they are not matching any self-loop (because the properties are different for two vertices)

> ERROR: variable a already exists

dehowef:
[...]

PR #751, opened 2023-03-17
Fix property constraints against resolved variables (#724)

jrgemignani:
Fixed an issue where an already resolved variable, when used for
a property constraint, errored out. See issue #724 for more details.

This is the second part of the fix for the match property constraint
bug regarding variable reuse. The prior fix addressed clause-to-clause,
this fix addresses within a clause.
 
Adjusted and added additional regression tests.

Co-authored-by: Dehowe Feng <dehowefeng@gmail.com>


dehowef:
I have reviewed these changes with John and have tested them and they look good and align with the expected behavior that AGE should have

jrgemignani:
@dehowef Updated PR to include label checking and additional regression tests.

dehowef:
Looked over the additional regression tests, and they look good to me.

PR #752, opened 2023-03-18
Apache Age Swift Driver: Methods added to check for the graph and Readme updated

FahadZaheerfzr:
Added a method to check if the graph already exists otherwise create one. Added this method on start-up. Moreover added some descriptions in the Readme file

### DIFF
--- a/drivers/age-swift/Sources/age-swift/age.swift
+++ b/drivers/age-swift/Sources/age-swift/age.swift
@@ -20,28 +20,28 @@
 import PostgresClientKit
 
 
-func setUpAge(connection:Connection){
+func setUpAge(connection:Connection, graphName: String){
     do{
-        try connection.prepareStatement(text: "Load 'age';").execute()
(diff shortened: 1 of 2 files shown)


drivers/age-swift/Sources/age-swift/age.swift
```diff
@@ -102,15 +127,17 @@ class Age{
...
-    
-    
-}
+        let rows = fetchSQL(statement: statement, connection: self.connection!)
+        
+        for column in rows{
+            print(column)
+        }
+    } 
+}
```
dehowef, line 143: Hi, your PR looks good, but can you correct this no newline at end of file error? Thank you

dehowef (commented):
Looks good overall

drivers/age-swift/Sources/age-swift/age.swift
FahadZaheerfzr: Added a new line

PR #754, opened 2023-03-20
[AGEHA] Add support for qualified name in Cypher query check

moeed-k:
Added support for qualified names in the isCypherQuery function (i.e. checks for ag_catalog.cypher() calls).


uhayat (approved):
Looks good, ready to merge

PR #756, opened 2023-03-21
[AGEHA] Implemented function to extract cypher functions

KamleshKumar427:
Implemented function to extract cypher functions

List* extractCypherFuncs(List* ParsedCyphertreeList);

returns List which contains all the cypher functions used in a query. 
This function is called only when the query is cypher.


uhayat (approved):
Looks good, ready to merge

PR #787, opened 2023-04-02
[ AGE Inheritance ] Fixed VLE inheritence

Zahid07:
Fixed the ` getChildren` to return only the children of the current edge label. 
We were getting all the children in the graph, but now only the one which we are currently searching.

### DIFF
--- a/src/backend/utils/adt/age_global_graph.c
+++ b/src/backend/utils/adt/age_global_graph.c
@@ -25,6 +25,7 @@
 #include "utils/rel.h"
 #include "utils/snapmgr.h"
 #include "commands/label_commands.h"
+#include "catalog/pg_inherits.h"
 
 #include "utils/age_global_graph.h"
 #include "utils/agtype.h"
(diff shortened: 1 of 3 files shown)


Zahid07:
We are getting conflicts because I changed the getChildren Function so that it caters the current Edge Label, and returns the children according to that Edge Label Name.

Zahid07:
Changed the getChildren to cater the following situation.

A inherits B
B inherits C.

When we do the query to get all labels connected to A, we were getting only A and B
But now we are getting A, B and C.

//Creating a graph
SELECT create_graph('Testing1');

//Creating Edge Labels with inheritance
SELECT create_elabel('Testing1', 'A');
SELECT create_elabel('Testing1', 'B', ARRAY['A']);
SELECT create_elabel('Testing1', 'C', ARRAY['B']);

//Creating Vertex Labels
SELECT create_vlabel('Testing1', 'Head');
SELECT create_vlabel('Testing1', 'Node');
SELECT create_vlabel('Testing1', 'Tail');

//Adding data
SELECT * FROM cypher('Testing1', $$
    CREATE (:Head {id: 1})-[:A]->(:Node {id: 2})-[:B]->(:Node {id: 3})-[:C]->(:Tail {id: 4})
 $$) AS (a agtype);

//Query
SELECT * FROM cypher('Testing1', $$
    MATCH (a)-[:A*1]->(b)
    RETURN a.id, b.id
$$) AS (a_id agtype, b_id agtype);

Result:
 a_id | b_id
------+------
 1    | 2
 2    | 3

Intended Result:
 a_id | b_id
------+------
 1    | 2
 2    | 3
 3    | 4

After changing the getChildren function, we get our intended result which is:
 a_id | b_id
------+------
 1    | 2
 2    | 3
 3    | 4

PR #792, opened 2023-04-06
[AGEHA] Fix and Optimize isWriteQuery and extractCypherFuncs

moeed-k:
Fixed incompatibility between isWriteQuery and extractCypherFuncs.
Also merged both functions into one for optimization. Now the List of clauses is only traversed once.


uhayat:
Looks good, ready to merge.

PR #803, opened 2023-04-12
[Viewer Desktop Go] Remove dist/assets and build/bin exe files

AhmarZaidi:
Removed the `build/bin` and `frontend/dist` directories as it is standard practice to not include them.
Removed the `exe` build files.


saadjameel:
@aked21 Please merge I've reviewed

PR #805, opened 2023-04-12
[Age-Viewer-Desktop-Go] : Node modules removed, repository cleaned

sarthak-19:
Removed node modules as it should not be present with the source code in repositories and codebase should be clean


saadjameel:
@aked21 Merge please, discussed in meeting about this

Issue #810, opened 2023-04-14
Remove Unused DSN in python driver unittest


Munmud:
When users trying to unittest, They get confused where to put the connection as there are two place to put the information as shown bellow
```python
DSN = "host=127.0.0.1 port=5432 dbname=postgres user=postgres password=agens"
TEST_HOST = "127.0.0.1"
TEST_PORT = 5432
TEST_DB = "postgres"
TEST_USER = "postgres"
TEST_PASSWORD = "agens"
TEST_GRAPH_NAME = "test_graph"
```

DSN is never used in the code. So deleting it might be a good choice for not to confuse users

Munmud:
Made a PR #811 for this

PR #811, opened 2023-04-14
Feat : Remove Unused DSN in python driver unittest

Munmud:
Made this pr as mentioned in issue #810 

DSN is never used in the code. So deleting it might be a good choice for not to confuse users.


dehowef (commented):
I think this change unnecessary

dehowef:
hey this isn't a feature

dehowef:
Also it's not a feature-- so the feature label (FEAT) is not the best here

Issue #812, opened 2023-04-14
Add optional argument for python driver unitTest


Munmud:
Changing source code for connection string is not user friendly, For making connection string, passing optional args like database name, username, password, host, port, graph name will be better when compailing in that case.

For example :
```cmd
python test_age_py.py \
-host "127.0.0.1" \
-db "postgres" \
-u "postgres" \
-pass "agens" \
-port 5432 \
-gn "test_graph"
```

PR #814, opened 2023-04-14
Add optional argument for python driver unitTest

Munmud:
As mentioned in issue : #812 

Example :
```
python test_age_py.py \
-host "127.0.0.1" \
-db "postgres" \
-u "postgres" \
-pass "agens" \
-port 5432 \
-gn "test_graph"
```

### Must add
- `-db` for database name
- `-u` for username
-  `-pass` for password

### Optional argument (If not match with default)
- `-host` for hostname. Default is "127.0.0.1"
- `port` for port no. Default is 5432
- `-gn` for graph name. Default is "test_graph"

### DIFF
--- a/.github/workflows/python-driver.yaml
+++ b/.github/workflows/python-driver.yaml
@@ -55,5 +55,5 @@ jobs:
 
     - name: Test
       run: |
-        python -m unittest -v test_age_py.py
+        python test_age_py.py -db "postgres" -u "postgres" -pass "agens"
         python -m unittest -v test_agtypes.py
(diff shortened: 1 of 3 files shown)


dehowef (commented):
Hey @Munmud can you tell me a little about the intent and rationale behind making these changes in the driver? Thank you!

Munmud:
Unit testing ensures that all code meets quality standards before it's deployed. Currently its required to change the credential by changing the source code for unittest. But its not always possible, For example when using it in a docker, Now I am only be able to unitTest if I build a docker postgres image by matching all the credential as used in soruce code. [See for example](https://github.com/apache/age/blob/master/drivers/docker-compose.yml). So thats why I thought why not pass all the information through args while compailing

PR #815, opened 2023-04-14
[Label Inheritance] Fix issue with excessive NOTICE messages being displayed

markgomer:
This PR is part of work for the [Label Inheritance Project 
](https://github.com/orgs/apache/projects/176)

* Previously, the system was displaying an excessive number of NOTICE messages, which was making it difficult for users to read other important information. This commit addresses this issue by adding a filter to remove any redundant or unnecessary NOTICE  
messages. As a result, users will now have a much clearer view of the system's status and will be able to easily identify any potential issues.  
* Changes on regress tests.  
* Removed variable for inheritance argument check.  
* Added function to issue ALTER TABLE ... INHERIT command through SPI  
interface.  

Co-authored-by: Matheus Matsumoto <matheusfarias519@gmail.com>

### DIFF
--- a/age--1.2.0.sql
+++ b/age--1.2.0.sql
@@ -102,7 +102,7 @@ CREATE FUNCTION ag_catalog.create_vlabel(graph_name name, label_name name, paren
     LANGUAGE c
 AS 'MODULE_PATHNAME';
 
-CREATE FUNCTION ag_catalog.create_elabel(graph_name name, label_name name)
+CREATE FUNCTION ag_catalog.create_elabel(graph_name name, label_name name, parent_list name[] = NULL)
     RETURNS void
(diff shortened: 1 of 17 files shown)


rafsun42:
@markgomer @MatheusFarias03 

The label redesign project is implementing label inheritance using openCypher spec instead of label commands. So, to create two children of `Person`, the query would be like- `CREATE (x:Person:Student), (y:Person:Admin)`. This will result in following tables: Person, Student, Admin, PersonStudent, PersonAdmin. The last two tables are intersections. Using intersection table will allow us to make queries like `MATCH (x:Person:Student)` faster.

The label redesign project may have conflicts with label inheritance because of having a different architecture.

PR #819, opened 2023-04-18
[AGEHA] Add cypher query routing to PgPool-II

moeed-k:
-Added cypher query routing to PgPool-II, such that write queries go to the primary node in load balancing/replication mode and read queries go to secondary nodes.


uhayat (approved):
Looks Good

PR #821, opened 2023-04-18
Make some changes in Makefile to remove compilation error.

moiz697:
Removed the Makefile.global entry and added path to include files. You need to export PGSRC to the PostgreSQL path.


ibrarahmad (approved):
LGTM

PR #822, opened 2023-04-19
Age Swift Driver: Rollback and Commit functions added

FahadZaheerfzr:
Added commit function for Age Class so the user can commit the changes after Insertion or Update. Also added rollback to allow the user to rollback a query


dehowef (approved):
Looks good

Issue #791, opened 2023-04-06
Segmentation fault when SET n.value = rand() or SET n.value = pi()


mohayu22:
**Describe the bug:**
The server crashes with Segmentation Fault error whenever the _value_ property of a Node or Edge is set to `rand()` or `pi()`
```
Program received signal SIGSEGV, Segmentation fault.

0x00005650823d9abb in datumTransfer (value=4592508940796821504, 
    typByVal=typByVal@entry=false, typLen=-1) at datum.c:192
192		if (!typByVal && typLen == -1 &&
```

**How to produce the error?**
```
SELECT * from cypher('demo', $$
  MATCH (n)
  SET n.value = rand()
  RETURN n
$$) as (result agtype);
```
And
```
SELECT * from cypher('demo', $$
  MATCH (n)
  SET n.value = pi()
  RETURN n
$$) as (result agtype);
```
**Important:**
```
/*
* Check for openCypher functions that are directly mapped to PG
* functions. We may want to find a better way to do this, as there
* could be many.
*/

if (pg_strcasecmp(name, "rand") == 0)
    funcname = SystemFuncName("random");
else if (pg_strcasecmp(name, "pi") == 0)
    funcname = SystemFuncName("pi");
else if (pg_strcasecmp(name, "count") == 0)
    funcname = SystemFuncName("count");
```
**Additional context:**
Using `rand() + 0` or `pi() + 0` resolves the issue and allows the program to function properly.

**Environment:**
- PostgreSQL 11.18 
- Apache Age (master branch)

jrgemignani:
@mohayu22 Thank you for finding this. I have a feeling this is a casting issue -

```
psql-11.5-5432-pgsql=# SELECT * from cypher('graph1', $$
  MATCH (n) SET n.value = rand()::integer RETURN n
$$) as (result agtype);
                                            result
----------------------------------------------------------------------------------------------
 {"id": 844424930131969, "label": "zero", "properties": {"name": "Zero", "value": 0}}::vertex
(1 row)

psql-11.5-5432-pgsql=# SELECT * from cypher('graph1', $$
  MATCH (n) SET n.value = rand()::float RETURN n
$$) as (result agtype);
                                                    result
--------------------------------------------------------------------------------------------------------------
 {"id": 844424930131969, "label": "zero", "properties": {"name": "Zero", "value": 0.880863124504685}}::vertex
(1 row)

psql-11.5-5432-pgsql=# SELECT * from cypher('graph1', $$
[...]

PR #823, opened 2023-04-20
Introduce age_pi and age_rand functions and related updates

mohayu22:
Resolved issue #791 

- Added `age_pi` and `age_rand`
- Added new regression tests

**Problem:**
In order to maintain compatibility with the agtype data type, all values used as properties of vertices and edges must be of the agtype type. When setting a property, such as `SET n.i = 3.1415`, the float value is first converted into agtype using the `float_to_agtype` function before being passed to postgres. Similarly, other conversion functions like `integer_to_agtype` and `string_to_agtype` are used as needed.

When a function like `pi` is used with a typecast, such as `SET n.i = pi()::int`, the function is executed, typecasted, and converted to agtype before being passed to postgres.

However, when using functions like `rand` or `pi` without a typecast, such as `SET n.i = pi()`, the function is passed directly to postgres and returns a float value that is not automatically converted into agtype. This can result in a segmentation fault.

**Solution:**
One potential solution is creating wrapper functions for the `drandom` and `dpi` ("d" stands for "double precision") functions of postgres, which would convert their output into agtype before passing it to postgres. 

**Changes:**
This pull request aims to resolve this problem by introducing the `age_rand` and `age_pi` functions. The creation of these functions involves adding related SQL code in the _age--1.2.0.sql_ file, and their C implementation in the _agtype.c_ file. 

The `age_pi` function is marked as `IMMUTABLE` but `age_rand` is not.

The C implementation uses the postgre `drandom` and `dpi` functions, which are located in the _float.c_ file. The `DirectFunctionCall1` macro is used to call these functions. The float return value of these functions is then converted to agtype.

**Grammer File:**
In _cypher_gram.c_, function node (fnode) is created for both age functions and those that directly mapped to postgres functions. However, after adding `age_rand` and `age_pi` functions, the code related to mapped functions `random` and `pi` is removed.

**Regression Tests:**
Regression tests have been added for these new functions. For testing `rand`, `n.i < 1 AND n.i >= 0` logic is used which would return `true` to pass.


jrgemignani:
@mohayu22 Please update your commit description, the big box, to give a description of what was done and why. Make sure to conform to the archaic :) 80 columns per line of text.

image

mohayu22:
@jrgemignani Done

jrgemignani:
@mohayu22 It still shows the same for the commit description,...

mohayu22:
I have updated the PR description.

jrgemignani:
image

mohayu22:
@jrgemignani I have changed the commit messages.

PR #832, opened 2023-04-23
[Dynamic window HEIGHT & WIDTH]

Huzaiifaaaa:
1-Set window height dynamically-> Gets and sets current device height and width(uses https://github.com/lxn/win)
2-Removed "agensgraph.c"-> not required for "age_viewergo"(Might have dependency issue with other branches).


dehowef:
Outdated

PR #835, opened 2023-04-23
Update Makefile and add Flex/Bison files for a sample project

carlasanches:
This PR updates the Makefile to compile psql code and adds Flex/Bison files for a sample project called "match". Additionally, this PR includes the necessary rules to generate and compile project files.


dehowef:
Outdated

PR #836, opened 2023-04-24
[Ageviewer_go_Desktop] Added functionalities to read query response dynamically

KamleshKumar427:
After this PR the response of any query can be obtained in the form of JSON, from the backend dynamically.


saadjameel:
@aked21 Please merge, discussed and reviewed

PR #846, opened 2023-04-25
Added [HELP] button on main screen(Without Dynamic Height & Width)

Huzaiifaaaa:
- Added help button on main screen(needs o be linked to proper documentation/community)
- Removed Dynamic Height & Width(as in PL #832 ) 

image


dehowef:
Outdated

PR #850, opened 2023-04-25
Fixed indentation of vertexExp

rrrokhtar:
Removed extra space at the samples of python driver that was causing "**Unexpected indentation**" at vertexExp definition in [drivers/python/samples/apache-age-agtypes.ipynb](https://github.com/apache/age/blob/master/drivers/python/samples/apache-age-agtypes.ipynb) at cell no.3


jrgemignani:
@rrrokhtar Please fill in the big 'Add an optional extended description...' box with some details. It doesn't have to be much, but it can't be blank. I know it states it as optional, but for the AGE project, it isn't optional.

image

rrrokhtar:
Updated

PR #853, opened 2023-04-26
Refactored imports and added docstrings to functions

hammadsaleemm:
I removed the line from .age import * since it is already imported in the next line. Instead of importing all modules using from .age import * and from .models import *, I explicitly imported the necessary functions/classes directly. I also added docstrings to all functions and modules to explain their purpose, inputs, and outputs, and added type annotations to all functions to improve code readability and maintainability.

I also renamed rawPrinter to RawPrinter to follow Python naming conventions and made it a separate class RawPrinter that inherits from ResultHandler to make it more extensible. Finally, I updated the connect function to return an Age object instead of None.

### DIFF
--- a/drivers/python/age/__init__.py
+++ b/drivers/python/age/__init__.py
@@ -13,22 +13,40 @@
 # specific language governing permissions and limitations
 # under the License.
 
-from . import age
-from .age import *
-from .models import *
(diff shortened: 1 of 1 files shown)


hammadsaleemm:
> @hammadsaleemm There is a python test case that is failing on this commit -- can you look into why it is failing?

I'll look into it

dehowef:
Never followed up. Closing

PR #863, opened 2023-04-27
Ignored .vscode folder

humzakt:
Added the .vscode folder in gitignore because whenever i setup launch.json in vscode debugger, it adds that to the changes as well. I believe this would help all others using vscode as a debugger

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ build.sh
 .DS_Store
 *.tokens
 *.interp
+.vscode


rrrokhtar:
There is a trick I'm using to can exclude/ignore some files in git locally even if it is not ignored on the remote git
Open terminal in the project directory
```sh
echo '/.vscode/*' >> .git/info/exclude
```
You can treat that exclude file as a local .gitginore file which you can add whatever you want there

jrgemignani:
@humzakt Your local repository is for you to manage. Please don't add items to the AGE repository to manage yours for you. See @rrrokhtar comment. 

Unless you can justify this PR, please close it.

Issue #870, opened 2023-04-28
Odd behavior in context of orderability of different agtypes.


CapnSpek:
**Describe the bug**
While evaluating the inequality between an edge and a path, it appears that the orderability is totally bypassed and true/false is returned based on the sign of the inequality.

**How are you accessing AGE (Command line, driver, etc.)?**
Command line

**What data setup do we need to do?**
Apache AGE, and any graph with atleast 2 vertices and 1 edge.
```pgsql
SELECT * from cypher('my_graph_name', $$
  CREATE (:Person {name: 'John'})-[:Knows]->(:Person {name: 'Jonh'})
$$) as (a agtype);
```

**What is the necessary configuration info needed?**
Any graph with atleast 2 vertices and 1 edge.

**What is the command that caused the error?**

In regards to orderability between different agtypes: - 
https://age.apache.org/age-manual/master/intro/comparability.html

I was testing the orderability between path and an edge, and I found this odd behavior: -

Comparing a path with an edge with p > e inequality: -
```pgsql
test=# SELECT *
FROM cypher('test', $$
        WITH [{id: 0, label: "label_name_1", properties: {i: 0}}::vertex,
            {id: 2, start_id: 0, end_id: 1, label: "edge_label", properties: {i: 0}}::edge,
           {id: 1, label: "label_name_2", properties: {}}::vertex
           ]::path as p
MATCH (n)-[e]-(m) RETURN p>e
$$) AS (e agtype);
  e   
------
 true
 true
 true
 true
(4 rows)
```
Fair enough, same result holds when we compare it with p < e inequality
```pgsql
test=# SELECT *
FROM cypher('test', $$
        WITH [{id: 0, label: "label_name_1", properties: {i: 0}}::vertex,
            {id: 2, start_id: 0, end_id: 1, label: "edge_label", properties: {i: 0}}::edge,
           {id: 1, label: "label_name_2", properties: {}}::vertex
           ]::path as p
MATCH (n)-[e]-(m) RETURN p<e
$$) AS (e agtype);
   e   
-------
 false
 false
 false
 false
(4 rows)
```
But look, when we change the inequality to e > p
```pgsql
test=# SELECT *
FROM cypher('test', $$
        WITH [{id: 0, label: "label_name_1", properties: {i: 0}}::vertex,
            {id: 2, start_id: 0, end_id: 1, label: "edge_label", properties: {i: 0}}::edge,
           {id: 1, label: "label_name_2", properties: {}}::vertex
           ]::path as p
[...]

PR #872, opened 2023-04-28
[AGE] Update agtype_util.c in reference to issue #870

CapnSpek:
In reference to issue https://github.com/apache/age/issues/870 This would resolve the case of 
edge > path, and path > edge both being true in case of inequalities.

Regarding why I picked this specific order of priorities, it is mainly based on the documentation https://age.apache.org/age-manual/master/intro/comparability.html (even though the order remains ambiguous there as well,) and the answers received in reference to this stackoverflow question. https://stackoverflow.com/questions/75845310/clarification-regarding-apache-age-documentation-orderability-between-different

However, if required, the priorities are up for debate.

### DIFF
--- a/src/backend/utils/adt/agtype_util.c
+++ b/src/backend/utils/adt/agtype_util.c
@@ -197,20 +197,24 @@ uint32 get_agtype_length(const agtype_container *agtc, int index)
  */
 static int get_type_sort_priority(enum agtype_value_type type)
 {
-    if (type == AGTV_OBJECT)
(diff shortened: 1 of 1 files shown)


jrgemignani:
From https://github.com/apache/age/issues/870

> @CapnSpek Then it needs to be discussed as to what the priority should be for those types. I would look to Neo4j for guidance.

This needs to be discussed first, then a PR can be created - not before.

PR #881, opened 2023-04-30
[Ageviewer_go_Desktop] Refactor Inline Styles

Hasan-Iqtedar:
I have separated inline styles into constant objects for better readability and reusability. Also removed unnecessary imports and added the license comments which were missing.


dehowef:
Outdated

Issue #871, opened 2023-04-28
Using CALL with a function and a RETURN with an undefined variable does not output correct error


aru-d-at:
**Description**
CALL function with a RETURN does not output correct error message

**Mode:**
- Command line

**Data Setup:**
- None

**Configuration:**
- Default AGE 1.3.0 installation with PostgreSQL 12

**Command**
```pgsql
SELECT * FROM cypher('test', $$
    CALL toInteger('1')
    RETURN v
$$) as (v agtype);
```
```
ERROR:  Procedure call inside a query does not support naming results implicitly
LINE 1: SELECT * FROM cypher('test', $$
                                      ^
HINT:  Name explicitly using `YIELD` instead
```

**Expected behavior**
Should throw syntax error at `RETURN v` instead of crashing

Similar to
```
ERROR:  could not find rte for v
LINE 3: RETURN v
               ^
```

**Environment:**
- Version: Tested using 1.3.0

dehowef:
Hi, thanks for catching this~ can you go ahead and modify the code to throw the appropiate error? thx!

aru-d-at:
@dehowef Sure, working on it.

aru-d-at:
@dehowef Hi, could you please check the PR. Thanks.

jrgemignani:
@aru-d-at I agree that the crash is not an acceptable result. But, neither is `ERROR:  could not find rte for v`. That is an error stating that something is wrong with the code; that something was transformed or built incorrectly. Errors like that need to be fixed as well. So, while you may be able to stop it from crashing, you still haven't resolved the core issue.

> Expected behavior
> Should throw syntax error at RETURN v instead of crashing
> 
> Similar to
> 
> ```
> ERROR:  could not find rte for v
> LINE 3: RETURN v
>                ^
> ```

aru-d-at:
@jrgemignani So, after further investigating the issue, I have found that the crash by itself is unrelated to the `CALL` clause. Rather, it seems using AGE 1.3.0 on a Database that had previously used AGE 1.1.0 results in crashing the database. Most queries seem to cause a crash.

When I opened the issue, my database was in a a similar state (AGE 1.1.0 -> 1.3.0) and it just so happens the query used at the time had the `CALL` clause. For this issue, the title as well as the error output needs to be edited. So I will do that first.

In regards to the crash, it can be either be that, 1.3.0 is incompatible with 1.1.0 based Databases or that it is indeed a bug. I may need to open a separate issue for this.

[...]

PR #887, opened 2023-05-02
Updated Error message for `CALL` with `RETURN`

aru-d-at:
- ADDED Syntax Error on `CALL func() RETURN rte`

- Adjusted `CALL` Regression Test expected ERROR message

The error message reflects missing `rte` declaration when running solo `CALL` function

Fixes issue #871 

### Checks:
- Regression tests:
Installcheck


jrgemignani:
@aru-d-at Please see my comment in the issue thread. This PR needs to address the root cause of the crash and produce an acceptable message; either a syntax error or something about invalid usage. But `ERROR: could not find rte` is not acceptable as it is a problem in and of itself.

PR #888, opened 2023-05-02
Moved create_empty_agtype out of loops to avoid repetition

mohayu22:
`create_empty_agtype` can be classified as a pure function since it does not modify the database, but simply creates an empty agtype.

To avoid repeated calls inside the loop, `create_empty_agtype` was declared outside the loop and its output was assigned to the `props` variable which is then used inside the loop. 

This approach ensures that the function is only called once and improves the efficiency of the code.


jrgemignani:
@mohayu22 Please adjust your commit description to the archaic 80 characters per line standard :)

image

mohayu22:
@jrgemignani Done

PR #889, opened 2023-05-03
Fix issue #878 - unrecognized node type error

### LINKED ISSUE #878: Unexpected Error: unrecognized node type: 107
When running the following query:
```cypher
MERGE ({n0:(null IS NULL)})
```

I encountered the following error:
```
unrecognized node type: 107
```

I believe the query mentioned above is semantically and syntactically correct and thus no error should be thrown here.

### Steps to reproduce
Run the following query and observe it throws an error:
```cypher
MERGE ({n0:(null IS NULL)})
```

### Expected behavior

### Actual behavior
The query fails with the error message `unrecognized node type: 107`.

MuhammadTahaNaveed:
- The error was in the way expr was getting transformed in case of T_NullTest.
- The original expr was also getting tranformed, causing the nodeType of that expr unrecognizable next time the expr is used.
- Fixed by creating a new NullTest expr.
- Added regression tests as well.

resolves #878

### DIFF
--- a/src/backend/parser/cypher_expr.c
+++ b/src/backend/parser/cypher_expr.c
@@ -152,12 +152,15 @@ static Node *transform_cypher_expr_recurse(cypher_parsestate *cpstate,
     case T_NullTest:
     {
         NullTest *n = (NullTest *)expr;
+        NullTest *tranformed_expr = makeNode(NullTest);
 
-        n->arg = (Expr *)transform_cypher_expr_recurse(cpstate,
-                                                       (Node *)n->arg);
-        n->argisrow = type_is_rowtype(exprType((Node *)n->arg));
+        tranformed_expr->arg = (Expr *)transform_cypher_expr_recurse(cpstate,
+                                                         (Node *)n->arg);
+        tranformed_expr->nulltesttype = n->nulltesttype;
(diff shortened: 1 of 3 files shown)


src/backend/parser/cypher_expr.c
```diff
@@ -152,12 +152,15 @@ static Node *transform_cypher_expr_recurse(cypher_parsestate *cpstate,
     case T_NullTest:
     {
         NullTest *n = (NullTest *)expr;
+        NullTest *tranformed_expr = makeNode(NullTest);
```
jrgemignani, line 155: Is there a particular reason for creating a new node as opposed to using the passed node?

jrgemignani:
@MuhammadTahaNaveed Added a question, otherwise it looks good.

src/backend/parser/cypher_expr.c
jrgemignani: Actually, what about using the transformed node and just checking to see if it was already transformed?

MuhammadTahaNaveed:
@jrgemignani where? I could not see it.

jrgemignani:
@MuhammadTahaNaveed 
image

jrgemignani (approved):

Issue #275, opened 2022-09-06
Barbell Graph


JoshInnis:
the barbell graph is two complete graphs connected by a bridge path

Syntax:
```
ag_catalog.age_create_tadpole_graph(graph_name Name,
                                    m int,
                                    n int, 
                                    vertex_label_name Name DEFAULT = NULL,
                                    vertex_properties agtype DEFAULT = NULL,
                                    edge_label_name Name DEAULT = NULL,
                                    edge_properties agtype DEFAULT = NULL,
                                    bidirectional bool DEFAULT = true)

```

Input:

- graph_name - Name of the graph to be created
- m - number of vertices in both complete graphs
- n - number of vertices in the bridge path.
- vertex_label_name - Name of the label to assign each vertex to.
- vertex_properties - Property values to assign each vertex. Default is NULL
- edge_label_name - Name of the label to assign each edge to.
- edge_properties - Property values to assign each edge. Default is NULL
- bidirectional - Bidirectional True or False. Default True.

https://en.wikipedia.org/wiki/Barbell_graph

markgomer:
Working on this enhancement.

jrgemignani:
FYI: I did not close this issue, this was done by an automated task. 

Please do not have automated tasks close issues under someone else's name without notifying others that you are doing so as this causes unnecessary confusion. @aked21

Issue #271, opened 2022-09-06
Complete Graph


JoshInnis:
A complete graph on n nodes means that all pairs of distinct nodes have an edge connecting them.

Syntax: 
```
ag_catalog.age_create_complete_graph(graph_name Name, Nodes int,
                                                                vertex_label_name Name DEFAULT = NULL,
                                                                vertex_properties agtype DEFAULT = NULL,
                                                                edge_label_name Name DEAULT = NULL,
                                                                edge_properties agtype DEFAULT = NULL,
                                                                bidirectional bool DEFAULT = TRUE);

```
Input:
- graph_name - Graph Name
- nodes - Number of Nodes
- vertex_label_name - Name of the label to assign each vertex to.
- vertex_properties  - Property values to assign each vertex. Default is NULL
- edge_label_name - Name of the label to assign each edge to.
- edge_properties  - Property values to assign each edge. Default is NULL
- bidirectional - Bidirectional True or False. Default True.

Returns Void

https://en.wikipedia.org/wiki/Complete_graph

thegautamkumarjaiswal:
I would like to contribute and develop this algorithm using python.

JoshInnis:
ag_catalog.age_create_complete_graph(graph_name Name, nodes int)

susano0:
Currently working on this functionality.

PR #892, opened 2023-05-04
[Graph Generation Algorithms] Add properties on creating a complete graph:

markgomer:
This work is part of the [Graph Generation Algorithms][1] project, issues [271][2] and [275][3].

* change hint on top of create_complete_graph function to match the new arguments;
* add some comments;
* move same vertex/edge label check up in order to prevent creation of an empty graph;
* change on age_create_barbell_graph to be compatible with the new arguments on create_complete_graph;
* change age--1.2.0.sql to receive agtype properties as arguments of create_complete_graph;
* change and add regress tests.

[1]: https://github.com/orgs/apache/projects/103/views/1
[2]: https://github.com/apache/age/issues/271
[3]: https://github.com/apache/age/issues/275


markgomer:
My idea of implementing this appeared just because it's in the description of the issue #271.

But I think that, since the purpose of the function is to create a new set of nodes in a graph, it would be beneficial for them to have something like an initial structure in common.

A use case I thought would be on creating a new group with a set of individuals on a social media.
Maybe they could have an initial Status of newcomer in this group, with 0 reputation and Default privacy settings.

In the same line, the connection between them would start with a strength of 1, and if they are best friends as false.

We would start that graph with the command:

```opencypher
SELECT create_complete_graph('social_media', 20, 
'group_connection', '{"strength":"1", "best_friends":"true"}',
'that_new_group', '{"Status":"Newcomer", "Reputation":"0", "Privacy_settings":"Default"});
```

So that would serve as a baseline or starting point for all users, making the initialization of the graph more straightforward.

PR #896, opened 2023-05-05
[Ageviewer-go] Retrieve metadata of all the graphs within a connected database.

KamleshKumar427:
Metadata functionalities changed to retrieve data from all graphs, rather than passing through from the front end, so that We can show all the graphs and corresponding nodes and vertices at the front end.


saadjameel:
@aked21  please merge

KamleshKumar427:
@aked21, would you kindly merge the pending PRs at your earliest convenience? We would greatly appreciate it, as this will enable us to proceed with other tasks that are contingent upon the code being updated. Thanks

saadjameel:
@aked21 Can you please merge, need to do more PRs.

PR #900, opened 2023-05-06
Python Driver: Updated the module import statement for sample notebook

FahadZaheerfzr:
The sample notebook was importing a module that didn't exist. I updated that module


dehowef:
Looks good~

PR #903, opened 2023-05-08
Fix issue #898 - Crash on property access of node in same MATCH

### LINKED ISSUE #898: Crash on property access of node in same `MATCH`
I found a bug using my cypher fuzzer.

When running the following query against an empty database:
```
CREATE () WITH * MATCH (x{n0:x.n1}) RETURN 0
```

The postgres instance crashes and goes into recovery mode.

### Steps to reproduce

Spin up a local instance of **apache/age:PG13_latest**: `docker run -e POSTGRES_PASSWORD=123 --rm --name age apache/age:PG13_latest`

Get a shell in the docker container: `docker exec -it age /bin/bash`

Connect to postgres: `su postgres -c psql`

Run the following queries:
```psql
LOAD 'age';
-
-
SELECT create_graph('graph');
-
    CREATE () WITH * MATCH (x{n0:x.n1}) RETURN 0
$$) as (v agtype);
```

### Expected behavior

### Actual behavior
The database crashes

MuhammadTahaNaveed:
- Transformed entity was not added in cpstate entities before tranforming props due to which self referencing props were not resolving. Added them to cpstate before transforming props.
- Added test cases.

Resolves #898

### DIFF
--- a/regress/expected/cypher_match.out
+++ b/regress/expected/cypher_match.out
@@ -1787,6 +1787,61 @@ ERROR:  duplicate edge variable 'b' within a clause
 LINE 1: ...pher('cypher_match', $$ MATCH p=(a)-[b:knows]->()-[b]->(a) R...
                                                              ^
 --
+-- self referencing property constraints (issue #898)
+--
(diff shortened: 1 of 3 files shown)


dehowef:
Hey Taha, these changes look good to me. Thanks~

Issue #899, opened 2023-05-05
`SET` of property to `true OR true` causes crash


DominicWuest:
I found a bug using my cypher fuzzer.

When running the following query against an empty database:
```
CREATE (x) SET x.n0 = (true OR true)
```

The postgres instance crashes and goes into recovery mode.

I encountered this issue when testing queries against the **apache/age:PG13_latest** docker image.

### Steps to reproduce

Spin up a local instance of **apache/age:PG13_latest**: `docker run -e POSTGRES_PASSWORD=123 --rm --name age apache/age:PG13_latest`

Get a shell in the docker container: `docker exec -it age /bin/bash`

Connect to postgres: `su postgres -c psql`

Run the following queries:
```psql
LOAD 'age';
-
SET search_path = ag_catalog, "$user", public;
-
SELECT create_graph('graph');
-
SELECT * FROM cypher('graph',$$
    CREATE (x) SET x.n0 = (true OR true)
$$) as (v agtype);
```

### Expected behavior
The query should run successfully

### Actual behavior
The database crashes

DominicWuest:
`gdb` stacktrace:

```
#0  0x000055766192b3d8 in datumTransfer ()
#1  0x00005576617bdef4 in SPI_datumTransfer ()
#2  0x00007f1581ec5610 in plpgsql_exec_function ()
   from /usr/lib/postgresql/13/lib/plpgsql.so
#3  0x00007f1581ed0dd3 in plpgsql_call_handler ()
   from /usr/lib/postgresql/13/lib/plpgsql.so
#4  0x00005576617832e2 in ?? ()
#5  0x00007f158ae35644 in ExecEvalExprSwitchContext (isNull=0x7ffe97658227, 
    econtext=0x557662aa4360, state=0x557662aa4518)
    at /usr/include/postgresql/13/server/executor/executor.h:322
#6  ExecProject (projInfo=0x557662aa4510)
    at /usr/include/postgresql/13/server/executor/executor.h:356
#7  exec_cypher_create (node=<optimized out>)
    at src/backend/executor/cypher_create.c:248
#8  0x00005576617904e0 in ExecScan ()
#9  0x00007f158ae36efb in ExecProcNode (node=0x557662a607f8)
    at /usr/include/postgresql/13/server/executor/executor.h:248
#10 exec_cypher_set (node=0x557662a5ffc8)
    at src/backend/executor/cypher_set.c:627
#11 0x000055766178742d in standard_ExecutorRun ()
[...]

PR #904, opened 2023-05-09
Fix SET bug, issue #899

jrgemignani:
This patch fixes a bug in SET and likely all clauses that use the agtype_volatile_wrapper.

The original wrapper was written such that it could only handle an AGTYPE input. Additionally, it was called regardless of the input type. This would cause crashes on any non-AGTYPE types given as input to the function. For example, a PG boolean.

The fix was to replace this wrapper with a C function that takes "any" type as input and then creates AGTYPE return values for non-AGTYPE types. AGTYPE values would be passed per usual.

Added regression tests.


dehowef (approved):
Looks good to me.

Issue #824, opened 2023-04-20
Typecast bool to integer and vice versa


Zainab-Saad:
**Is your feature request related to a problem? Please describe.**
The typecasting from bool data type to integer and vice versa is not supported
The query:
`agedb=# SELECT * FROM cypher('graph', $$
RETURN 1::bool
$$) AS (result agtype);`
outputs: 
`ERROR:  typecast 'bool' not supported
LINE 1: SELECT * FROM cypher('graph', $$`

and 
`agedb=# SELECT * FROM cypher('graph', $$
RETURN true::integer
$$) AS (result agtype);`
output:
`ERROR:  typecast expression must be a number or a string`

**Describe the solution you'd like**
To the point I have debugged, issue lies with transform_cypher_typecast function in backend/parser/cypher_expr.c and thus some modifications need to be made there

**Describe alternatives you've considered**
N/A

**Additional context**
This feature of typecasting bool to integer and vice versa is supported in postgres, thus it would be worth adding this feature to AGE as well.

jrgemignani:
@Zainab-Saad This looks like something that could be useful to add. I'm not sure about which  of the these **::bool** or **::boolean** would be best to use, though. Also, you may want to consider adding this to the "other" casts to PG types (like ::pg_bigint and ::pg_float8)

Zainab-Saad:
This PR works with `agtype_to_bool` which is not what I meant in this issue.
@jrgemignani As already brought to your notice, I am already working on a PR regarding the typecasting using the operator ::
Moreover, I am working on adding this to other casts like ::pg_bigint and ::pg_float8

jrgemignani:
@Zainab-Saad Will you be creating a PR for this?

Zainab-Saad:
@jrgemignani  Yes. I am working on it

PR #905, opened 2023-05-09
Typecast bool to integer and vice versa

Zainab-Saad:
Implemented the enhancement raised in issue #824 

- Added, modified code for typecast bool to integer and vice versa
- Added, modified regression tests

### DIFF
--- a/regress/expected/agtype.out
+++ b/regress/expected/agtype.out
@@ -2163,6 +2163,9 @@ SELECT agtype_to_int4(agtype_in('null'));
                
 (1 row)
 
+--
+-- Test agtype to integer cast
+--
 SELECT agtype_to_int4(agtype_in('1'));
  agtype_to_int4 
 ----------------
@@ -2186,6 +2189,8 @@ SELECT agtype_to_int4(agtype_in('"string"'));
 ERROR:  invalid input syntax for integer: "string"
 SELECT agtype_to_int4(agtype_in('[1, 2, 3]'));
 ERROR:  cannot cast agtype array to type int
+SELECT agtype_to_int4(agtype_in('{"int":1}'));
+ERROR:  cannot cast agtype object to type int
 --
 -- Test agtype to int[]
 --
--- a/regress/expected/agtype.out
+++ b/regress/expected/agtype.out
@@ -2077,11 +2077,15 @@ SELECT agtype_to_bool(agtype_in('false'));
  f
 (1 row)
 
+SELECT agtype_to_bool(agtype_in('1'));
+ agtype_to_bool 
+----------------
+ t
+(1 row)
+
 -- These should all fail
 SELECT agtype_to_bool(agtype_in('null'));
 ERROR:  cannot cast agtype null to type boolean
-SELECT agtype_to_bool(agtype_in('1'));
-ERROR:  cannot cast agtype integer to type boolean
 SELECT agtype_to_bool(agtype_in('1.0'));
 ERROR:  cannot cast agtype float to type boolean
 SELECT agtype_to_bool(agtype_in('"string"'));
--- a/regress/expected/agtype.out
+++ b/regress/expected/agtype.out
@@ -2139,6 +2139,27 @@ SELECT agtype_to_int8(agtype_in('false'));
 (1 row)
 
 --
+-- Test boolean to integer cast
+--
+SELECT agtype_to_int4(agtype_in('true'));
+ agtype_to_int4 
+----------------
+              1
+(1 row)
+
+SELECT agtype_to_int4(agtype_in('false'));
+ agtype_to_int4 
+----------------
+              0
+(1 row)
+
+SELECT agtype_to_int4(agtype_in('null'));
+ agtype_to_int4 
+----------------
+               
+(1 row)
+
+--
 -- Test agtype to int[]
 --
 SELECT agtype_to_int4_array(agtype_in('[1,2,3]'));
--- a/regress/expected/agtype.out
+++ b/regress/expected/agtype.out
@@ -2163,6 +2163,29 @@ SELECT agtype_to_int4(agtype_in('null'));
                
 (1 row)
 
+SELECT agtype_to_int4(agtype_in('1'));
+ agtype_to_int4 
+----------------
+              1
+(1 row)
+
+SELECT agtype_to_int4(agtype_in('1.45'));
+ agtype_to_int4 
+----------------
+              1
+(1 row)
+
+SELECT agtype_to_int4(agtype_in('1.444::numeric'));
+ agtype_to_int4 
+----------------
+              1
+(1 row)
+
+-- These should all fail
+SELECT agtype_to_int4(agtype_in('"string"'));
+ERROR:  invalid input syntax for integer: "string"
+SELECT agtype_to_int4(agtype_in('[1, 2, 3]'));
+ERROR:  cannot cast agtype array to type int
 --
 -- Test agtype to int[]
 --
--- a/regress/expected/expr.out
+++ b/regress/expected/expr.out
@@ -1147,6 +1147,14 @@ $$) AS (i bigint);
  1
 (1 row)
 
+SELECT * FROM cypher('type_coercion', $$
+	RETURN true
+$$) AS (i int);
+ i 
+---
+ 1
+(1 row)
+
 --Invalid String Format
 SELECT * FROM cypher('type_coercion', $$
 	RETURN '1.0'
@@ -1250,6 +1258,22 @@ $$) AS r(result agtype);
  result 
 --------
  3
+(1 row)
+
+SELECT * FROM cypher('expr', $$
+RETURN true::int
+$$) AS r(result agtype);
+ result 
+--------
+ 1
+(1 row)
+
+SELECT * FROM cypher('expr', $$
+RETURN false::int
+$$) AS r(result agtype);
+ result 
+--------
+ 0
 (1 row)
 
 SELECT * FROM cypher('expr', $$
--- a/regress/expected/expr.out
+++ b/regress/expected/expr.out
@@ -1370,6 +1370,14 @@ SELECT * FROM cypher('expr', $$
 RETURN 'infinity'::float::int
 $$) AS r(result agtype);
 ERROR:  bigint out of range
+SELECT * FROM cypher('expr', $$
+RETURN ''::int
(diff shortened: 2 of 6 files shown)


Zainab-Saad:
Before adding to the regression tests, here are a few edge cases:
`SELECT * FROM cypher('test', $$
RETURN ""::int
$$) AS (result agtype);
ERROR:  invalid input syntax for integer: ""`

`SELECT * FROM cypher('test', $$
RETURN "falze"::int
$$) AS (result agtype);
ERROR:  invalid input syntax for integer: "falze"`

`SELECT * FROM cypher('test', $$
RETURN ""::bool
$$) AS (result agtype);
ERROR:  cannot cast agtype string to type boolean`

`SELECT * FROM cypher('test', $$
RETURN "falze"::bool
$$) AS (result agtype);
ERROR:  cannot cast agtype string to type boolean`

@jrgemignani Do I add these as well?

Zainab-Saad:
@jrgemignani In the above tests that I have shared, are these good to add? Moreover, do I add regression tests for float::bool as well? (I didnt add that before because this PR dealt with integer <-> booleans)

jrgemignani:
@Zainab-Saad Those are good to add. To answer your question I'll reiterate these 2 comments -

> We want to attempt to cover as many possible ways a user could use a feature, both good and bad, and put them in as regression tests.

> The goal is to show that an addition, or modification, works when used properly and that it gives the appropriate error messages when it isn't. And, most importantly, that it doesn't crash on user input.

Creating regression tests can be hard. You need to think of how a user, well intentioned or not, might use a feature or modification.

Zainab-Saad:
@jrgemignani Yes, I understand that. Moreover, I have written a few more regression tests and commenting them here so that I can make a commit to add only relevant regression tests:
1. `SELECT * FROM cypher('test', $$
RETURN ""     
$$) AS (result int);
ERROR:  invalid input syntax for integer: ""`

2. `SELECT * FROM cypher('test', $$
RETURN "falze"
$$) AS (result int);
ERROR:  invalid input syntax for integer: "falze"`

3. `SELECT * FROM cypher('test', $$
RETURN tRuE   
$$) AS (result int);`

 result 
- - - - - - 
      1
(1 row)

4. `SELECT * FROM cypher('test', $$
RETURN ""         
$$) AS (result bool);
ERROR:  cannot cast agtype string to type boolean`

5. `SELECT * FROM cypher('test', $$
RETURN "falze"
$$) AS (result bool);
ERROR:  cannot cast agtype string to type boolean`

These are the tests for *type coercion*

jrgemignani:
@Zainab-Saad I will **generally** never say one has too many regression tests. And PRs are works in progress, so don't worry about making many adjustments to them. It is good to get used to anyway.

Zainab-Saad:
@jrgemignani Done with adding regression tests for edge cases

jrgemignani:
@Zainab-Saad If you could also add more to agtype.sql please :)

Zainab-Saad:
@jrgemignani Added test cases for agtype_to_int4 (added tests for float, numeric which were missing as well and checked for bad input). agtype_to_bool was fine already

regress/expected/agtype.out
```diff
@@ -2138,6 +2142,50 @@ SELECT agtype_to_int8(agtype_in('false'));
...
+----------------
+              0
+(1 row)
+
+SELECT agtype_to_int4(agtype_in('null'));
+ agtype_to_int4 
+----------------
+               
+(1 row)
+
```
jrgemignani, line 2165: Might want to add a comment stating that these are for agtype to integer casting

regress/expected/agtype.out
```diff
@@ -2138,6 +2142,50 @@ SELECT agtype_to_int8(agtype_in('false'));
...
+ agtype_to_int4 
+----------------
+              1
+(1 row)
+
+-- These should all fail
+SELECT agtype_to_int4(agtype_in('"string"'));
+ERROR:  invalid input syntax for integer: "string"
+SELECT agtype_to_int4(agtype_in('[1, 2, 3]'));
+ERROR:  cannot cast agtype array to type int
```
jrgemignani, line 2188: You may also want to include objects in your tests, fx -

`SELECT agtype_to_int4(agtype_in('{name: "Jason", age: 34, gpa: 3.4}'));` 

or just numbers and characters -

`SELECT agtype_to_int4(1);`
`SELECT agtype_to_int4(3.1415);`
`SELECT agtype_to_int4(true);`
`SELECT agtype_to_int4('1');`

These are just examples.

jrgemignani (changes requested):

regress/expected/agtype.out
Zainab-Saad: `SELECT agtype_to_int4(1);` and `SELECT agtype_to_int4(true);` termiantes the connection to server. The other two give error.

jrgemignani: If your changes cause the crash, or it is related to the area you are working in, I would say that you should address it with your PR. You can always state that you fixed issues with the code that caused crashes or whatever the case may be.

Zainab-Saad: @jrgemignani I have verified, the crash is not because of my changes but existed before. The server crashes for agtype_to_int2, agtype_to_int4, agtype_to_int8 for this input. 
I am looking for a solution to fix this, but I believe that dealing this bug here would make this PR very cluttered since there are many functions that need changes.
I can resolve this bug in another PR if that's fine to deal with these edge cases there.

jrgemignani: Okay. Create an issue for the crash and note that the regression tests that caused the failures will be addressed in the forthcoming patch.

regress/expected/agtype.out
Zainab-Saad: @jrgemignani  Done

jrgemignani (approved):

Zainab-Saad:
@jrgemignani I see that in the merged commit, regression tests fail as well. Is there a need to fix this?

jrgemignani:
@Zainab-Saad I reran the test and it was successful. I also pulled down the latest master and it built successfully.

Zainab-Saad:
@jrgemignani Okay. And about the JDBC test, is it acceptable that a driver test fails?

jrgemignani:
@Zainab-Saad Btw, for future reference, you don't need this much detail for your extended commit description -

```
commit 87f860151fc17f7d124ddf3f409f2b24acd928de
Author: Zainab Saad <105385638+Zainab-Saad@users.noreply.github.com>
Date:   Wed May 10 22:33:42 2023 +0500

    Typecast bool to integer and vice versa (#905)

    * Typecast boolean to integer

    - Modify agtype_typecast_int and agtype_to_int4 in agtype.c

    * Add regression tests for typecast boolean to integer

    - Add regression tests: sql script, expected results in agtype.sql, agtype.out
    - Add regression tests: sql script, expected results in expr.sql, expr.out

    * Typecast integer to boolean

    - Modify transform_cypher_typecast in cypher_expr.c
    - Modify agtype_to_bool in agtype.c

    * Add regression tests for typecast integer to boolean

    - Add/Modify regression tests in agtype.sql, agtype.out
    - Add/Modify regression tests in expr.sql, expr.out

    * Add regression tests for edge cases

    - Add regression tests in expr.sql, expr.out
    - These regression tests check the edge cases for typecasting

    * Add regression tests for edge cases - II

    - Add regression tests in agtype.sql, agtype.out
    - Add edge cases for agtype_to_int4

    * Fix & Add regression tests

    - Fix comments in agtype.sql, agtype.out
    - Add regression test for edge case
```

Something like this would be more concise -

```
commit 87f860151fc17f7d124ddf3f409f2b24acd928de
Author: Zainab Saad <105385638+Zainab-Saad@users.noreply.github.com>
Date:   Wed May 10 22:33:42 2023 +0500

    Typecast bool to integer and vice versa (#905)

    * Typecast boolean to integer

    - Modify agtype_typecast_int and agtype_to_int4

    * Typecast integer to boolean

    - Modify transform_cypher_typecast
    - Modify agtype_to_bool

    - Add regression tests
```

jrgemignani:
> @jrgemignani Okay. And about the JDBC test, is it acceptable that a driver test fails?

For whatever reason, that test sometimes fails and sometimes doesn't. It is likely a resource issue. It isn't okay, but re-running it usually fixed it. I would worry if the actual build fails or some other test fails. But, only if rerunning the test isn't successful.

Zainab-Saad:
@dehowef Thanks for pointing that out. 
Should the pull request revert implicit casting from integer to booleans only or from booleans to integers too? 
Because at present, there is implicit casting both ways.

Zainab-Saad:
@dehowef Moreover, the query results as:
`SELECT * FROM cypher('test', $$
RETURN TRUE AND 1
$$) AS (result agtype);
ERROR:  cannot cast agtype integer to type boolean`
without the patch.

dehowef:
Hey @Zainab-Saad , I think that that implicit casting and working around PG's Boolean type is tricky in general. If it requires removal of implicit boolean casting one way fixes this case, it may be wise to perform additional tests to see if any other edge cases may fail. Thank you~, let me know if there is anything else I can add to be more clear or helpful

Zainab-Saad:
@dehowef I have seen that a possible fix of this issue (eliminating int to bool implicit casting while retaining the explicit :: casting) is to define a new function `agtype_typecast_bool` which will be used for ::boolean casting while inside the `agtype_to_bool`, we simply remove the casting from int to bool.
Moreover, I am looking into edge cases where implicit casting from bool to int/int8 can cause issues (did not find an edge case yet). The code base initially did not have `agtype_typecast_bool` defined (there is this function for other data types like `agtype_typecast_int`, `agtype_typecast_numeric` etc.) and I was wondering if there was a specific reason for this exclusion that I should be aware of.
Thanks

dehowef:
A lot of the boolean checking and processing is done on the Postgres side of things, so being able to implicitly cast values into the boolean type without actually ensuring it's validity would cause errors like this. That is why we never included an implicit cast in the first place. We have discussed it and the boolean implementation in general, though.

dehowef:
Yeah, I think explicit casting is fine-- implicit operations is where the errors usually pop up

Zainab-Saad:
@dehowef I have created PR #923 to address this issue. Kindly take a look.
Thanks

dehowef:
Will do~ Thank you for your contributions, @Zainab-Saad

Issue #908, opened 2023-05-10
NULLIF() function


WendelLana:
**Is your feature request related to a problem? Please describe.**
I was trying to create a conditional statement using the `NULLIF()` function inside a Cypher query, but I received an error message stating that the function does not exist.

**Describe the solution you'd like**
A similar function to the PG `NULLIF()` function

**Describe alternatives you've considered**
Using `CASE` would be an alternative to `NULLIF`

**Additional context**
I'm working on a PR for this, if there's no problem

farakh-shahid:
Yes i have checked this We have alternative CASE .
We can achieve similar functionality  using the 'CASE' expression in cypher.

Try this :
```
MATCH  (n:node)
RETURN CASE
WHEN n.property = 'value' THEN NULL
ELSE n.property
END AS result;
```

This will is  n.property has a value of 'value' then it will return null for this  else it will return n.property.

PR #910, opened 2023-05-10
Added NULLIF() function

WendelLana:
PR for issue #908.

Added NULLIF() function that takes 2 arguments of "any" type (converting non-AGTYPE types to AGTYPE values) and return NULL if the two specified expressions are equal, if not returns the first expression.

Added regression tests.

### DIFF
--- a/age--1.3.0.sql
+++ b/age--1.3.0.sql
@@ -365,7 +365,7 @@ AS 'MODULE_PATHNAME';
 CREATE FUNCTION ag_catalog._label_name(graph_oid oid, graphid)
 RETURNS cstring
 LANGUAGE c
-STABLE
+IMMUTABLE
 PARALLEL SAFE
 AS 'MODULE_PATHNAME';
 
@@ -3104,7 +3104,7 @@ AS 'MODULE_PATHNAME';
 CREATE FUNCTION ag_catalog._agtype_build_vertex(graphid, cstring, agtype)
 RETURNS agtype
 LANGUAGE c
-STABLE
+IMMUTABLE
 CALLED ON NULL INPUT
 PARALLEL SAFE
 AS 'MODULE_PATHNAME';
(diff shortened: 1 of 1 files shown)


jrgemignani:
@WendelLana Is this PR necessary? There is a response in issue #908 stating how this could be done without this - 

image

Additionally, if it is necessary, you need to bring your PR up-to-date with the master branch -

image

WendelLana:
@jrgemignani In my issue I've said using CASE would be an alternative to NULLIF too, but I find that the NULLIF() function is less verbose and simpler to use, even when nested. This is also part of a project ([Conditional Statement Improvements](https://github.com/orgs/AGEDB-INC/projects/15)) that you can find [here](https://github.com/orgs/AGEDB-INC/projects/15/views/1?pane=issue&itemId=18128198).

jrgemignani:
@WendelLana Are AGDB-INC projects separate from AGE? Also, your build still fails.

WendelLana:
@rafsun42 I don't believe so. However, there is an issue on the Neo4j repository about adding this function, and it's also easier to use and comprehend compared to a command like this:
`COALESCE(CASE columnname WHEN '' THEN NULL ELSE columnname END, 'STRING') `
or nested CASE clauses.

rafsun42:
@WendelLana Could you elaborate about the issue you are referring to?

Could you also give a use case of `nullif` function?

WendelLana:
I'm referring to this [issue](https://github.com/neo4j/neo4j/issues/1058) that shows one use case. I believe we could use it in some scenarios, for instance, when dividing by zero. Consider the following Cypher query:
````
MATCH (n)-[]->(m)
RETURN n.value / nullif (m.value, 0)
````

Additionally, I find this command:
````
RETURN NULLIF(n.value, m.value)
````
to be better and less verbose than:
````
RETURN 
    CASE n.value = m.value THEN NULL
    ELSE n.value
````
I know it may seem like a very specific case, but I genuinely think having a function like Postgres' `NULLIF` is beneficial as it promotes more integration between SQL and Cypher. This permits a simpler logic when writing a Cypher query using SQL-like logic.

Issue #921, opened 2023-05-14
Python Driver : Reference to author name and author email may not be correct


Munmud:
In [Setup.py](https://github.com/apache/age/blob/a43002ae39132781c535031d426d04bc5464851e/drivers/python/setup.py#L28) `author` and `author_email` is referencing to specific person. So it might not be correct.

hammadsaleemm:
Hey @Munmud , ig youre right the author and author_email fields in the setup.py file are currently set to specific values ('rhizome' and 'rhizome.ai@gmail.com', respectively). It is possible that these values may not be correct or up-to-date.

Munmud:
Resolved the issue with pr

PR #922, opened 2023-05-14
Changed author name and email

Munmud:
Updated author name and email address as Referenced in issue #921 
Previously it was referencing to specific person. So it might not be correct.


dehowef:
@rhizome-ai  was the original developer of the driver, and those are likely his credentials. I think this change would be good.

PR #923, opened 2023-05-14
Remove implicit casting

Zainab-Saad:
- Fixed&Added code, regression tests to deal an edge case  highlighted 
[here](https://github.com/apache/age/pull/905#issuecomment-1544287220)

### DIFF
--- a/age--1.2.0.sql
+++ b/age--1.2.0.sql
@@ -4039,6 +4039,13 @@ IMMUTABLE
 PARALLEL SAFE
 AS 'MODULE_PATHNAME';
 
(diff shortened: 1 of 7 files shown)


Zainab-Saad:
@dehowef I have fixed the code for the edge case that you described. This PR at present only removes implicit casting from int to bool and not the other way around. Doing this fixes the issue of being able to use int as an operand in logical operations. 
Moreover, I have looked into the regression tests and did not find an edge case where implicit casting from bool to int could cause issue.

dehowef (commented):
This is fantastic! Looks good Zainab~

PR #925, opened 2023-05-14
Update README.md

Hamza-Mushtaque:
Updated details about Postgres Installation on Ubuntu. As some essential commands were missing. Those changes are already live in Apache Age documentation Setup part.
But were not here in README.md file of his repository which results in inconsistency, plus error if one tries to follow the installation guide present on this page.


dehowef:
I think that it may be redundant to list a install command for every version of postgres.

PR #930, opened 2023-05-15
Added MATCH, WITH, WHERE, RETURN command functionality

im-aditya30:
**Instructions**
- Requires PG15
- Create following variables: 
```
 export PATH={location of pgsql-15}/bin:$PATH
 export PGDATA={location of pgsql-15}/bin/data
```
- Create a database and start the pgsql server
- Clone the repo and run the following command: 
```
flex -b -Cfe -p -p -o 'cypher.c' cypher.l
bison -d cypher.y
make
```
- Run `./agesql`
- Type `/` and then type the query to view the result

**Example**
On running `MATCH (n:Person {gender: 'M'}) RETURN n;`
The output is as follows: 
```
Match clause parsed successfully.
 Match = true
Labels: n
Node label parsed: Person.
Node pattern parsed successfully.
String expression parsed: M.
Map entry parsed successfully.
Property gender
Map literal entry parsed successfully.
Nonempty map literal.
Node properties parsed successfully.
No WHERE clause.
No WITH clause.
Identifier expression parsed: n
Return item parsed successfully.
Return item parsed successfully.
No ORDER BY clause.
No SKIP clause.
No LIMIT clause.
Return clause parsed successfully.
Return = true
```

### DIFF
--- a/cypher.l
+++ b/cypher.l
@@ -37,7 +37,9 @@
 
 [0-9]+ { yylval.int_val = atoi(yytext); return INTEGER; }
 [a-zA-Z][a-zA-Z0-9_]* { yylval.str_val = strdup(yytext); return IDENTIFIER; }
-"([^\"]|\.)*" { yylval.str_val = strdup(yytext); return STRING; }
+\"([^\"]|\.)*\" { yylval.str_val = strdup(yytext); return STRING; }
+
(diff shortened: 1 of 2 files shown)


dehowef:
Outdated

PR #931, opened 2023-05-16
Fix issue #876 - column reference "_age_default_alias_0" is ambiguous

### LINKED ISSUE #876: Unexpected Error: column reference "_age_default_alias_0" is ambiguous
When running the following query:
```cypher
MATCH ({n0:0}) MATCH ()-->() MATCH ({n1:0})--() RETURN 0 AS n2
```

I encountered the following error:
```
column reference "_age_default_alias_0" is ambiguous
```

I believe the query mentioned above is semantically and syntactically correct and thus no error should be thrown here.

### Steps to reproduce
Run the following query and observe it throws an error:
```cypher
MATCH ({n0:0}) MATCH ()-->() MATCH ({n1:0})--() RETURN 0 AS n2
```

### Expected behavior

### Actual behavior
The query fails with the error message `column reference "_age_default_alias_0" is ambiguous`.

MuhammadTahaNaveed:
- The problem was with get_next_default_alias function which was not generating unique aliases.
- Every clause transformed as a subquery has its own cpstate which is being freed after it is tranformed. The root cpstate is the one that has the default_alias_number initialized. So we need to reach the root cpstate to get the next correct default alias number.
- This change also fixed some previously incorrect results (verified using neo4j).
- Added additional regression tests.

Resolves #876

### DIFF
--- a/src/backend/parser/cypher_parse_node.c
+++ b/src/backend/parser/cypher_parse_node.c
@@ -131,8 +131,21 @@ RangeTblEntry *find_rte(cypher_parsestate *cpstate, char *varname)
  */
 char *get_next_default_alias(cypher_parsestate *cpstate)
 {
+    ParseState *pstate = (ParseState *)cpstate;
+    cypher_parsestate *parent_cpstate = (cypher_parsestate *)pstate->parentParseState;
     char *alias_name;
     int nlen = 0;
+
+    /*
+     * Every clause transformed as a subquery has its own cpstate which is being
+     * freed after it is tranformed. The root cpstate is the one that has the
+     * default alias number initialized. So we need to reach the root cpstate to
+     * get the next correct default alias number.
+     */
+    if (parent_cpstate)
+    {
+        return get_next_default_alias(parent_cpstate);
+    }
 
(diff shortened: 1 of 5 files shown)


src/backend/parser/cypher_parse_node.c
```diff
@@ -131,9 +131,20 @@ RangeTblEntry *find_rte(cypher_parsestate *cpstate, char *varname)
...
     char *alias_name;
     int nlen = 0;
 
+    /*
+     * Every clause transformed as a subquery has its own cpstate which is being
+     * freed after it is tranformed. The root cpstate is the one that has the
+     * default alias number initialized. So we need to reach the root cpstate to
+     * get the next correct default alias number.
+     */
+    if (parent_cpstate)
```
jrgemignani, line 145: Remember to add {} for all if/else

jrgemignani (changes requested):
Minor coding issue.

src/backend/parser/cypher_parse_node.c
MuhammadTahaNaveed: Fixed it

PR #937, opened 2023-05-19
[Python driver] adding connect_timeout

Mghrabi:
* Added `connect_timeout` to files in samples and to `test_age_py.py` to quit long-time failed connections
* Modified the connection instructions based on the changes

Reason for making the PR:
I went through an issue while trying to connect to a database and it failed for some reason (using a jupyter notebook), the cell that runs the connection request was running forever and without yielding any output as the connect_timeout is not set. By adding the connect_timeout long-time failed connections quit

Note: (FROM MY UNDERSTANDING) I think adding timeout is a good practice, especially for local connections which shouldn't take a long time to connect, that's why I made this PR anyway since it targets samples and test files.


Mghrabi:
Reason for making the PR:
I went through an issue while trying to connect to a database and it failed for some reason (using a jupyter notebook), the cell that runs the connection request was running forever and without yielding any output as the `connect_timeout` is not set. By adding the `connect_timeout` long-time failed connections quit

* Note: (FROM MY UNDERSTANDING) I think adding `timeout` is a good practice, especially for local connections which shouldn't take a long time to connect, that's why I made this PR anyway as it targets **samples** and **test** files.

dehowef (commented):
Hey, I think it might be better to let the users kill the program if it hangs rather than have it terminate early on them. Thank you for your contribution though.

Mghrabi:
Hey, thank you for your time, I think if it is terminated early on them, they will have control over that as the `connect_timeout` is visible to them while making the connection and the property value is easy to change or even remove, so from that perspective it should be ok, which means if it didn't turn out to be helpful it wouldn't hurt. However, it could be helpful in situations like the one I mentioned above where a failed connection doesn't or takes so long before erroring out which could be confusing. 
Thanks again for your valuable time.

PR #942, opened 2023-05-21
AGE Swift Driver: Readme updated

FahadZaheerfzr:
Add description of Apache Age Swift Driver

### DIFF
--- a/drivers/age-swift/README.md
+++ b/drivers/age-swift/README.md
@@ -1,20 +1,40 @@
-# Licensed to the Apache Software Foundation (ASF) under one
-# or more contributor license agreements.  See the NOTICE file
-# distributed with this work for additional information
-# regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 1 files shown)


dehowef (changes requested):
Hey Fahad, Your additions are great, but I don't think removing the licensing information from the readme file here is a good idea. You may want to check with @eyab on the License info requirements for each file.

eyab:
@FahadZaheerfzr We can not remove the licenses. The license needs to remain on that page.

FahadZaheerfzr:
Ah I saw the readme of other drivers and noticed there were no License. No worries I will update

dehowef:
@FahadZaheerfzr That's a good thing to point out actually, we might need to add licensing info to those READMEs as well. @eyab thoughts?

FahadZaheerfzr:
The requested changes are made and pushed @dehowef please check

dehowef (approved):

PR #952, opened 2023-05-29
Allow using distinct timestamps within the same query

mohayu22:
While the STABLE flag ensured that the timestamp() value remained constant throughout the query due to caching, it restricted the flexibility of using multiple timestamps in single query.

By removing the STABLE volatility flag from the age_timestamp() function, we allow the usage of two different timestamps within a single query.


dehowef:
Some regression tests and examples of the previous unwanted behavior and new, preferred behavior would be helpful here.

PR #956, opened 2023-05-31
Added License in Python Driver README.md

FahadZaheerfzr:
Added the license which was missing in the Python Driver readme file


dehowef:
Good work~

PR #957, opened 2023-06-01
Fix MERGE crashes due to incorrect label type

jrgemignani:
Fixed a crash due to MERGE not checking if the label retrieved was of the correct type.

NOTE: This fix is temporary until a better solution can be
      implemented.

Added error messaging for improper usage of the add_volatile_wrapper function. Previously, it was causing crashes when used by MERGE.


jrgemignani:
@MuhammadTahaNaveed There are items flagged in this PR that could be worked on, if you want to.

PR #958, opened 2023-06-01
Fix memory leaks in MERGE (merge_edge & merge_vertex)

jrgemignani:
Fixed memory leaks in MERGE. Specifically in the merge_edge and merge_vertex functions. These functions will now throw error messages rather than attempt to store values in the wrong locations. Additionally, added some code to make it easier to debug the source of an incorrect tuple position.

Fixed bad regression tests.

Updated code formatting.


dehowef:
Looks good.

Issue #897, opened 2023-05-05
Complex `MERGE` causes crash


DominicWuest:
I found a bug using my cypher fuzzer.

When running the following query against an empty database:
```
MERGE n0 = (:A)-[:B{n1:0}]->(x:C)-[:D]->(x)-[:E]->(x)<-[:F{n2:0}]-(:G{n3:0})-[:H{n4:0}]->(y:I{n5:0})<-[:J]-(y)<-[:K]-(:L{n6:0})
```

The postgres instance crashes and goes into recovery mode.

I encountered this issue when testing queries against the **apache/age:PG13_latest** docker image.

### Steps to reproduce

Spin up a local instance of **apache/age:PG13_latest**: `docker run -e POSTGRES_PASSWORD=123 --rm --name age apache/age:PG13_latest`

Get a shell in the docker container: `docker exec -it age /bin/bash`

Connect to postgres: `su postgres -c psql`

Run the following queries:
```psql
LOAD 'age';
-
SET search_path = ag_catalog, "$user", public;
-
SELECT create_graph('graph');
-
SELECT * FROM cypher('graph',$$
    MERGE n0 = (:A)-[:B{n1:0}]->(x:C)-[:D]->(x)-[:E]->(x)<-[:F{n2:0}]-(:G{n3:0})-[:H{n4:0}]->(y:I{n5:0})<-[:J]-(y)<-[:K]-(:L{n6:0})
$$) as (v agtype);
```

### Expected behavior
The query should run successfully

### Actual behavior
The database crashes

DominicWuest:
`gdb` stacktrace:

```
#0  0x00007f158d9c2d16 in epoll_wait (epfd=4, events=0x5576629cb6b0, 
    maxevents=1, timeout=-1) at ../sysdeps/unix/sysv/linux/epoll_wait.c:30
#1  0x00005576618ce67c in WaitEventSetWait ()
#2  0x00005576617cae30 in secure_read ()
#3  0x00005576617d2947 in ?? ()
#4  0x00005576617d37e5 in pq_getbyte ()
#5  0x00005576618f1579 in PostgresMain ()
#6  0x00005576618794a5 in ?? ()
#7  0x000055766187a3f4 in PostmasterMain ()
#8  0x00005576615fc046 in main ()
```

jrgemignani:
@DominicWuest Actually, I take that back. I was able to get it to crash.

dehowef:
[...]

PR #961, opened 2023-06-02
Fix complex MERGE causes crash (#897)

jrgemignani:
Fixed the complex MERGE crash, which was due to storing variables in tuple positions that did not exist. This happened on MERGE commands where there wasn't a RETURN clause.

Added logic to catch these and handle appropriately. Meaning, discarding the variable results, instead of storing them.

Fixing this issue has highlighted the need to fix variable reuse in the MERGE transform itself. This is because MERGE preprocesses the path before handing it off to anything else.

Added regression tests.

### DIFF
--- a/regress/sql/cypher_merge.sql
+++ b/regress/sql/cypher_merge.sql
@@ -479,6 +479,22 @@ SELECT * FROM cypher('cypher_merge', $$ MATCH (n:node) RETURN n $$) AS (n agtype
 SELECT * FROM cypher('cypher_merge', $$ MERGE (n:node {name: 'Jason'}) SET n.name = 'Lisa', n.age = 23, n.gender = 'Female' RETURN n $$) AS (n agtype);
 SELECT * FROM cypher('cypher_merge', $$ MATCH (n:node) RETURN n $$) AS (n agtype);
 
+--
+-- Complex MERGE w/wo RETURN values
+--
+-- These should each create a path, if it doesn't already exist.
+-- TODO Until the issue with variable reuse of 'x' in MERGE is corrected,
+--      these commands will each create a new path.
+SELECT * FROM cypher('cypher_merge', $$ MERGE ()-[:B]->(x:C)-[:E]->(x:C)<-[f:F]-(y:I) $$) AS (x agtype);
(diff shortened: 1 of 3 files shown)


dehowef:
I looked over it, and good job preventing the crash. If @muhammadshoaib could take a look too, that'd be great

jrgemignani:
Added @MuhammadTahaNaveed as well.

regress/sql/cypher_merge.sql
```diff
@@ -479,6 +479,22 @@ SELECT * FROM cypher('cypher_merge', $$ MATCH (n:node) RETURN n $$) AS (n agtype
...
 SELECT * FROM cypher('cypher_merge', $$ MATCH (n:node) RETURN n $$) AS (n agtype);
 
+--
+-- Complex MERGE w/wo RETURN values
+--
+-- These should each create a path, if it doesn't already exist.
+-- TODO Until the issue with variable reuse of 'x' in MERGE is corrected,
+--      these commands will each create a new path.
+SELECT * FROM cypher('cypher_merge', $$ MERGE ()-[:B]->(x:C)-[:E]->(x:C)<-[f:F]-(y:I) $$) AS (x agtype);
+SELECT * FROM cypher('cypher_merge', $$ MERGE ()-[:B]->(x:C)-[:E]->(x:C)<-[f:F]-(y:I) RETURN x $$) AS (x agtype);
```
MuhammadTahaNaveed, line 489: These queries error out in neo4j saying `Can't create node x with labels or properties here.` pointing towards second reference of x. Shoudn't these fail in age as well ?

MuhammadTahaNaveed (commented):
@jrgemignani  just a small question.

regress/sql/cypher_merge.sql
MuhammadTahaNaveed: okay thanks... my confusion is clear on this thing now...
Marking it as resolved...

jrgemignani: It was good to bring it up because it highlights that MERGE is not done yet. We still have work to do with it.

PR #987, opened 2023-06-13
Fix issue #945 unaliased match clause giving incorrect results in count

### LINKED ISSUE #945: If an alias is not specified, the count result is incorrect.
**Describe the bug**
If an alias is not specified, the count result is incorrect.

- anywhere. I did in Command line.

```pgsql
SELECT * from cypher('movie_graph', $$
CREATE (a:Part {part_num: '123'}), 
       (b:Part {part_num: '345'}), 
       (c:Part {part_num: '456'}), 
       (d:Part {part_num: '789'})
$$) as (a agtype);
```
```
moviedb=# SELECT * from cypher('movie_graph', $$ match (a:Part) return a $$) as (result agtype);
                                        result                                        
--------------------------------------------------------------------------------------
 {"id": 1125899906842625, "label": "Part", "properties": {"part_num": "123"}}::vertex
 {"id": 1125899906842626, "label": "Part", "properties": {"part_num": "345"}}::vertex
 {"id": 1125899906842627, "label": "Part", "properties": {"part_num": "456"}}::vertex
 {"id": 1125899906842628, "label": "Part", "properties": {"part_num": "789"}}::vertex
(4 rows)
```

When an alias is not specified
```
moviedb=# SELECT * from cypher('movie_graph', $$ match (:Part) return count(*) $$) as (result agtype);   
 result 
--------
 1  <-------- incorrect count result
(1 row)
```
When an alias and a label is specified
```
moviedb=# SELECT * from cypher('movie_graph', $$ match (a:Part) return count(*) $$) as (result agtype); 
 result 
--------
 4
(1 row)
```

When only an alias is specified
```
moviedb=# SELECT * from cypher('movie_graph', $$ match (a) return count(*) $$) as (result agtype);     
 result 
--------
 4
(1 row)
```

**Expected behavior**

Same as agensgraph or neo4j

On Neo4j
```
neo4j@ldbcgraph> match (:Part) return count(*) ;
+----------+
| count(*) |
+----------+
| 4        |
+----------+
```

On AgensGraph
```
sampledb=# match (:Part) return count(*) ;
 count 
-------
 4
```
**Additional context**

Just so you know, agensgraph and neo4j give accurate results.

maazzzzz:
unaliased match clause would give incorrect results in count because the node wasn't being added to the join tree. Fixed it by adding unaliased match clauses with valid label names to join tree.
- added entry for unaliased nodes with label names in macro 
- added regression test for proposed change  

resolves #945

### DIFF
--- a/regress/expected/cypher_match.out
+++ b/regress/expected/cypher_match.out
@@ -2120,6 +2120,43 @@ SELECT * FROM cypher('cypher_match', $$MATCH ({n0:0}) MATCH ()-[]->() MATCH ({n1
 (0 rows)
 
 --
+-- issue#945 : un-aliased match clauses should be included in output node
+--
+SELECT * from cypher('cypher_match', $$ CREATE (a:label1 {id: '1'}), (b:label1 {id: '2'})$$) as (a agtype);
+ a 
+---
+(0 rows)
+
+SELECT * from cypher('cypher_match', $$ CREATE (a:label2 {id: '1'}), (b:label2 {id: '2'})$$) as (a agtype);
+ a 
+---
+(0 rows)
+
+SELECT * from cypher('cypher_match', $$ MATCH (:label1) RETURN count(*) $$) as (result agtype);
+ result 
+--------
+ 2
+(1 row)
+
+SELECT * from cypher('cypher_match', $$ MATCH (:label1),(:label2) RETURN count(*) $$) as (result agtype);
+ result 
+--------
+ 4
+(1 row)
+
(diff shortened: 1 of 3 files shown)


MuhammadTahaNaveed:
@jrgemignani can you please review this one. I think this was missed.

MuhammadTahaNaveed:
@maazzzzz Does it produce correct result for the below query?
```pgsql
MATCH () RETURN count(*)
```

maazzzzz:
@MuhammadTahaNaveed checked and it doesn't, missed that case when testing. I'll try looking into it if I get the time but if anyone else could look into it, thatd be great.

MuhammadTahaNaveed:
Issue addressed by this PR has already been resolved by PR #1288

PR #994, opened 2023-06-15
[AGE agtype_util.c] Fix issue #870 regarding orderability

CapnSpek:
Fixed issue #870 
 Odd behavior in context of orderability of different agtypes.

Implemented the solution suggested here:
https://github.com/apache/age/issues/870#issuecomment-1545592359

### DIFF
--- a/src/backend/utils/adt/agtype_util.c
+++ b/src/backend/utils/adt/agtype_util.c
@@ -197,20 +197,42 @@ uint32 get_agtype_length(const agtype_container *agtc, int index)
  */
 static int get_type_sort_priority(enum agtype_value_type type)
 {
-    if (type == AGTV_OBJECT)
+    if (type == AGTV_PATH)
+    {
         return 0;
-    if (type == AGTV_VERTEX)
+    }
+    if (type == AGTV_EDGE)
+    {
         return 1;
-    if (type == AGTV_ARRAY)
+    }
+    if (type == AGTV_VERTEX)
+    {
         return 2;
-    if (type == AGTV_STRING)
+    }
+    if (type == AGTV_OBJECT)
+    {
         return 3;
-    if (type == AGTV_BOOL)
+    }
+    if (type == AGTV_ARRAY)
+    {
         return 4;
-    if (type == AGTV_NUMERIC || type == AGTV_INTEGER || type == AGTV_FLOAT)
+    }
+    if (type == AGTV_STRING)
+    {
         return 5;
-    if (type == AGTV_NULL)
+    }
+    if (type == AGTV_BOOL)
+    {
         return 6;
+    }
+    if (type == AGTV_NUMERIC || type == AGTV_INTEGER || type == AGTV_FLOAT)
+    {
+        return 7;
+    }
+    if (type == AGTV_NULL)
+    {
+        return 8;
+    }
     return -1;
 }
 
@@ -356,6 +378,18 @@ int compare_agtype_containers_orderability(agtype_container *a,
                 break;
             }
 
+            /* Correction step because AGTV_ARRAY might be there just because of the container type */
+            /* Case 1: left side is assigned to an array, right is an object */
+            if(va.type == AGTV_ARRAY && vb.type == AGTV_OBJECT)
+            {
+                ra = agtype_iterator_next(&ita, &va, false);
+            }
+            /* Case 2: left side is an object, right side is assigned to an array */
+            else if(va.type == AGTV_OBJECT && vb.type == AGTV_ARRAY)
+            {
+                rb = agtype_iterator_next(&itb, &vb, false);
+            }
+            
             Assert(va.type != vb.type);
             Assert(va.type != AGTV_BINARY);
             Assert(vb.type != AGTV_BINARY);
--- a/regress/expected/agtype.out
+++ b/regress/expected/agtype.out
@@ -1523,7 +1523,7 @@ SELECT agtype_in('{"bool":true}') < agtype_in('{"bool":true, "null": null}');
 (1 row)
 
 -- Comparisons between types
--- Object < List < String < Boolean < Integer = Float = Numeric < Null
+-- Path < Edge < Vertex < Object < List < String < Boolean < Integer = Float = Numeric < Null
 SELECT agtype_in('1') < agtype_in('null');
  ?column? 
 ----------
@@ -1590,6 +1590,24 @@ SELECT agtype_in('{"bool":true, "integer":1}') < agtype_in('{"bool":true, "integ
  t
 (1 row)
 
+SELECT agtype_in('{"id":0, "label": "v", "properties":{"i":0}}::vertex') < agtype_in('{"bool":true, "i":0}');
+ ?column? 
+----------
+ t
+(1 row)
+
+SELECT agtype_in('{"id":2, "start_id":0, "end_id":1, "label": "e", "properties":{"i":0}}::edge') < agtype_in('{"id":0, "label": "v", "properties":{"i":0}}::vertex');
+ ?column? 
+----------
+ t
+(1 row)
+
+SELECT agtype_in('[{"id": 0, "label": "v", "properties": {"i": 0}}::vertex, {"id": 2, "start_id": 0, "end_id": 1, "label": "e", "properties": {"i": 0}}::edge, {"id": 1, "label": "v", "properties": {"i": 0}}::vertex]::path') < agtype_in('{"id":2, "start_id":0, "end_id":1, "label": "e", "properties":{"i":0}}::edge');
+ ?column? 
+----------
+ t
+(1 row)
+
 SELECT agtype_in('1::numeric') < agtype_in('null');
  ?column? 
 ----------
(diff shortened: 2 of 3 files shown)


CapnSpek:
Created the PR.

rafsun42:
@CapnSpek  Are you following these [documentation](https://age.apache.org/age-manual/master/intro/comparability.html#orderability-between-different-agtypes) for orders:

> Edge
> Path
> Map
> Vertex
> Edge
> Array
> String
> Bool
> Numeric, Integer, Float
> NULL

CapnSpek:
@rafsun42 the issue only arose when I noticed the ambiguity in the position of edge in the documentation, regarding which I created an issue here https://github.com/apache/age-website/issues/143

Then after discovering the odd behaviour I created the issue #870 

In which I and John discussed how the original purpose of having orderability was "orderability was chosen over comparability in order to allow for the sorting, grouping, of output."

Based on this information, I extrapolated the currently present priority order (which did not include edge or path at all btw)
From

Object, Vertex, Array String, Bool, Numbers, Null

To

Object, (Path, Edge), Vertex, Array, String, Bool, Numbers, Null.

I reached on this decision because if the aim was orderability and grouping, then, it correctly orders types as:-

Objects, Collection of graph elements (path), Graph elements (edges and vertices), Array, String, Bool, Numbers, Null.

Because it only adds some priorities in between the previously defined priorities, and does not change the structure that previously was there, this change is backwards compatible, and fixes the bug adequately by defining the previously left ambiguous priorities.

I'll come up with regression tests soon.

rafsun42:
@CapnSpek Sounds good. May I suggest the following priority order:
```
```

rafsun42:
@CapnSpek  Sounds good. May I suggest the following order:
```
(Path, Edge, Vertex), (Object, Array), String, Bool, Numbers, Null.
```
I believe `object` refers to map. So, overall order would be: graph elements > composite types (map, list) > primitive types.

CapnSpek:
My only issue with it is that we're changing the original priorities that are currently part of the code and it can have some unwanted effects. In the other solution, only some priorities are added in between the existing ones.

I feel maybe having minimal changes to code can make it more backwards compatible and avoid any unwanted drawbacks.

What do you think?

rafsun42:
@CapnSpek I am okay with your argument. I am just curious about the unwanted effects. What are the side effects other than regression tests giving result in different order? For example, do we use the `get_type_sort_priority` function internally somewhere that changing it would break some other functionalities?

CapnSpek:
The function `get_type_sort_priority` is not used anywhere other than the function `compare_agtype_containers_orderability` which is the function that decides between 2 given agtypes (of the same type or different type) which one should be prioritized above other. 

The function `get_type_sort_priority` is only used when deciding the orderability between different agtypes.
All other cases (of same agtypes) are handled with the function, or by other function `compare_agtype_scalar_values`.

I changed the `get_type_sort_priority` function to the suggested priority, reinstalled AGE, and ran the test cases. All the test cases have been passed.

It also solves all the problems that were given in issue #870 .

I think it is a good solution. Should I proceed to writing new regression tests?
@rafsun42

jrgemignani:
@CapnSpek All PRs should have [as complete as possible] regression tests for the work added, modified, or removed. You don't need to ask :)

dehowef:
Thank you for the putting in the work with your thorough explanation and documentation of this issue @CapnSpek . What you are saying makes sense to me.

CapnSpek:
While going through the regression tests relating to different agtypes (given in /regress/sql/agtype.sql )
I noticed all the tests are written in the following format: -

`SELECT agtype_in( [some agtype ] ) < [other agtype;`

I tried searching for a way to write an edge, or a vertex, or path that way, but I could not find anything.

So, I wanted to ask, would it be fine if I write the test case in the form of a cypher query?

CapnSpek:
I ran into a problem while testing it: -

Even after changing the priorities in the `get_type_sort_priority`, everything is good except for every query, object is the smallest agtype there is.

While debugging, I discovered this issue does not arise from that function, but instead from `compare_agtype_containers_orderability` function (agtype_util.c: line 231)

```
int compare_agtype_containers_orderability(agtype_container *a,
                                           agtype_container *b)
{
    agtype_iterator *ita;
    agtype_iterator *itb;
    int res = 0;

    ita = agtype_iterator_init(a);
    itb = agtype_iterator_init(b);

```

The function takes 2 agtype_container values as parameters (the two values to be compared).

Then it initializes an iterator for both, and declares the result variable in which the result of the comparison will be stored, then returned.

The issue takes place in the do-while loop: -

```
do
    {
        agtype_value va;
        agtype_value vb;
        agtype_iterator_token ra;
        agtype_iterator_token rb;

        ra = agtype_iterator_next(&ita, &va, false);
        rb = agtype_iterator_next(&itb, &vb, false);
```
The bottom two lines of the code above set both `va.type` & `vb.type` to `AGTV_ARRAY` in case of all agtypes except object, in which they set it to `AGTV_OBJECT`

The wrong `AGTV_ARRAY` assignment to all values only takes place in the first iteration of the loop. In the second iteration it corrects itself to the correct type.
The second iteration only takes place if both values have same agtype, thus this correction step is skipped when `AGTV_OBJECT` and `AGTV_ARRAY` are compared, and decision is readily made that `AGTV_ARRAY > AGTV_OBJECT`.

I believe it might be because `agtype_container` is an `ARRAY` or `OBJECT` type struct as written in its comment in agtype.h: line 224: -

```
/*
 * An agtype array or object node, within an agtype Datum.
 *
 * An array has one child for each element, stored in array order.
 *
 * An object has two children for each key/value pair.  The keys all appear
 * first, in key sort order; then the values appear, in an order matching the
 * key order.  This arrangement keeps the keys compact in memory, making a
 * search for a particular key more cache-friendly.
 */
typedef struct agtype_container
{
    uint32 header; /* number of elements or key/value pairs, and flags */
    agtentry children[FLEXIBLE_ARRAY_MEMBER];

    /* the data for each child node follows. */
} agtype_container;
```

Solution: -
Because it is a very special case (when both Agtypes differ, one is an array, other is an object)

The correction can be made into a special case below line 370 as follows

```
if(va.type == AGTV_ARRAY && vb.type == AGTV_OBJECT)
            {
                ra = agtype_iterator_next(&ita, &va, false);
            }
            else if(va.type == AGTV_OBJECT && vb.type == AGTV_ARRAY)
            {
                rb = agtype_iterator_next(&itb, &vb, false);
            }
```
That fixes the problem without any side effects.

CapnSpek:
@jrgemignani Can you take a look at this PR?

src/backend/utils/adt/agtype_util.c
```diff
@@ -197,20 +197,24 @@ uint32 get_agtype_length(const agtype_container *agtc, int index)
  */
 static int get_type_sort_priority(enum agtype_value_type type)
 {
-    if (type == AGTV_OBJECT)
+    if (type == AGTV_PATH)
```
jrgemignani, line 200: As we are updating a function, please make sure that it [the function] conforms to our standards. In this case, {} for all if/else/for/while/do/etc. So, for all of these ifs and anything else in this function. You don't need to worry about functions that you don't touch.

src/backend/utils/adt/agtype_util.c
```diff
@@ -356,6 +360,16 @@ int compare_agtype_containers_orderability(agtype_container *a,
...
 
+            /*Correction step because AGTV_ARRAY might be there just because of the container type*/
+            if(va.type == AGTV_ARRAY && vb.type == AGTV_OBJECT)
+            {
+                ra = agtype_iterator_next(&ita, &va, false);
+            }
+            else if(va.type == AGTV_OBJECT && vb.type == AGTV_ARRAY)
+            {
+                rb = agtype_iterator_next(&itb, &vb, false);
+            }
```
jrgemignani, line 371: What about a final else? What if the if and else if don't apply?

jrgemignani (changes requested):

jrgemignani:
@CapnSpek Reviewed and requesting changes

src/backend/utils/adt/agtype_util.c
CapnSpek: Got it, I will update this right away.

src/backend/utils/adt/agtype_util.c
CapnSpek: In the final case, the rest of the function executes as it is. Should I replace the if, else if with 2 ifs?
Because they refer to very special cases. There are 2 ifs only because its possible that the first one is array and second one is object, or second one is array and first one is object. Which would be the preferred code structure in this situation?

jrgemignani: So long as the final else isn't necessary, then leaving it out is okay. I just wanted to make sure you checked that.

src/backend/utils/adt/agtype_util.c
CapnSpek: Done.

src/backend/utils/adt/agtype_util.c
CapnSpek: I have added comments to explain the absence of last else.

jrgemignani:
@CapnSpek I would like to see some more regression tests for these changes. In particular, regression tests showing sorting/order by.

jrgemignani (changes requested):
Just add some more regression tests for sorting/order by.

CapnSpek:
So I wrote a test where a graph is created, then some vertices are added to that graph with a property having all the different agtypes for all the different vertices, and then I query for the vertices of that graph while also ordering them, and then the graph is dropped.

However, it seems it changes the graph_id for the creation of a graph in the next set of regression tests (catalog.out), so I would be modifying that as well here.

jrgemignani:
@CapnSpek 

> So I wrote a test where a graph is created, then some vertices are added to that graph with a property having all the different agtypes for all the different vertices, and then I query for the vertices of that graph while also ordering them, and then the graph is dropped.
> 
> However, it seems it changes the graph_id for the creation of a graph in the next set of regression tests (catalog.out), so I would be modifying that as well here.

You can add your new regression tests to the end of the sql file. Just make sure to state, in comments, what the regression tests are for, what they will do, and their expected results.

jrgemignani (approved):

jrgemignani:
@CapnSpek Looks good, adding @rafsun42 to get his opinion too.

rafsun42 (approved):
Looks good to me as well. Thanks for adding the tests.

CapnSpek:
Thank you. 
@jrgemignani Please merge.

dehowef:
I gave my review a week ago, but now that @rafsun42 and @jrgemignani have approved, I'll go ahead and merge it in.

PR #997, opened 2023-06-17
Fix MERGE variable reuse

jrgemignani:
Fixed the MERGE clause to allow for correct variable reuse in both the transform and execution phase.

Fixed an incorrect usage in MATCH where a variable was compared with pg_strcasecmp instead of strcmp.

Refactored some of the code using the volatile wrapper.

Verified and corrected old regression tests that were in error.

Added regression tests.


dehowef:
Looks good to me.

rafsun42 (approved):

dehowef (approved):

PR #1001, opened 2023-06-20
Fix issue #989: Impossible to create array with > 100 items

jrgemignani:
Fixed issue 989: Impossible to create an object with an array field of more than 100 elements.

This issue is due to the function agtype_build_list taking elements as arguments. As the elements can be anything, expressions too, they need to be processed by the transform phase and resolved in the execution phase. As PG has a limitation of 100 function arguments, this restricts the size of arrays built through agtype_build_list.

The fix was to break up large lists into segments of 100 items or less and then use the concatenation operator to join them in the execution phase.

Added regression tests.


dehowef (approved):
Creative solution. Looks good to me

MuhammadTahaNaveed (approved):
Tested it with SET and nested cases as well. Looks good to me 👍🏻 .

jrgemignani:
@MuhammadTahaNaveed Could you provide me with some of those tests? I will add them to the PR.

MuhammadTahaNaveed:
> @MuhammadTahaNaveed Could you provide me (email) with some of those tests? I will add them to the PR.
> 

Yeah sure.

jrgemignani:
Added Taha's additional regression tests.

dehowef:
Additional tests look good.

PR #1020, opened 2023-07-10
Added Python Git Ignore file

Munmud:
### Changes
- Added a gitignore file specifically designed for Python projects

### Source 
- [Official python gitignore](https://github.com/github/gitignore/blob/main/Python.gitignore)

### Why This change is required
The gitignore file ensures that certain files and directories related to Python development are ignored by Git, preventing them from being tracked and committed to the repository.

The gitignore file includes patterns to ignore commonly generated files and directories such as "pycache", "*.pyc", and "venv/". These entries are crucial for maintaining a clean and organized repository without cluttering it with unnecessary files.

By adding this gitignore file, it becomes easier to manage the repository and avoid accidentally committing files that should be ignored. It promotes better collaboration and reduces the chance of including sensitive or irrelevant information in the project history.

The addition of this gitignore file demonstrates a good practice for Python projects, ensuring that only the necessary files and directories are tracked by Git.

### DIFF
--- a/drivers/python/.gitignore
+++ b/drivers/python/.gitignore
@@ -1,0 +1,161 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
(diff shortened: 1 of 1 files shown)


dehowef:
I am rejecting this solely on the basis that **we do not allow ChatGPT written descriptions**

PR #1016, opened 2023-07-06
Added the toFloatList() function

M4rcxs:
- Basically the same function of OpenCypher toFloatList() https://neo4j.com/docs/cypher-manual/current/functions/list/#functions-tofloatlist
- toFloatList() converts a list of values and returns a list of floating point values. If any values are not convertible to floating point they will be null in the list returned.
- Returns a list containing the converted elements; depending on the input value a converted value is either a floating point value or null.
- Also added the regressions tests.

PR #1025, opened 2023-07-11
Added the toIntegerList() function

M4rcxs:
- This function is inspired by [toIntegerList()](https://neo4j.com/docs/cypher-manual/current/functions/list/#functions-tointegerlist) in OpenCypher.
- The function **toIntegerList()** takes a list of values and produces a new list containing only the integer values. If any values cannot be converted to an integer, they will be represented as null in the returned list.
- The result of the **toIntegerList()** function is a list that contains the converted elements. The converted values in the list can be either integer values or null, depending on the input value.
- Also added the regression tests.


dehowef:
@M4rcxs , this fails the regression build. Could you look into why this is occurring? Thanks

M4rcxs:
> @M4rcxs , this fails the regression build. Could you look into why this is occurring? Thanks

@dehowef The error was in the expected part of the regression test, but I fixed it!

dehowef:
Looks good now~, please refer to the message I commented on #1016 Thank you~

PR #1034, opened 2023-07-13
Fix issue #1033 - property access from entity in array

### LINKED ISSUE #1033: Property access from edge/vertex in array gives unexpected error
**Describe the bug**
Property access from edge/vertex in array gives unexpected error.

- terminal

```pgsql
    CREATE (n)-[:KNOWS {n:'hello'}]->({n:'hello'})
$$) as (a agtype);
```

```pgsql
    MATCH ()<-[x *]-()
    RETURN x[0].n
$$) as (a agtype);
```
```
ERROR:  container must be an array or object
```

**Expected behavior**
should have executed successfully and returned `hello`

- 1.3.0 PG12

MuhammadTahaNaveed:
- Issue was with agtype_access_operator. It was overseeing the possiilty of scalar types as element in array/object.
- Added regression tests.

resolves #1033

### DIFF
--- a/regress/expected/agtype.out
+++ b/regress/expected/agtype.out
@@ -2590,6 +2590,18 @@ SELECT agtype_access_operator(
  {"id": 1, "label": "label_name", "end_id": 3, "start_id": 2, "properties": {"key": "value"}}::edge
 (1 row)
 
+SELECT agtype_access_operator(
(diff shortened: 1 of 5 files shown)


jrgemignani:
Ooops. We both put out a PR for the same thing, lol

PR #1036, opened 2023-07-13
Add REGTYPEOID to toString function

jrgemignani:
Added the `REGTYPEOID ` to the `toString ` function. This will help in debugging by allowing `pg_typeof ` to work with AGE.

For example -

```
psql-13.9-5432-pgsql=# SELECT * FROM cypher('graph',$$
     MATCH ()<-[x *]-() RETURN toString(pg_catalog.pg_typeof(x))
$$) as (a agtype);
    a
----------
 "agtype"
 "agtype"
 "agtype"
(3 rows)
```

Added regression tests.


rafsun42 (commented):
Looks good!

There is one suggestion. It might be useful to have a function that returns the actual type of `agtype` as well (i.e. `edge`, `vertex` etc.). Do we have such functions?

dehowef (commented):
Looks good to me.

PR #1038, opened 2023-07-14
Added more trigonometry functions

titoausten:
- Added `cosec()` function: returns the reciprocal (not inverse) of sin() i.e `cosec() = 1 / sin()`.
- Added `acosec()` function: returns `sin^-1(1/x)`.
- Added `asec()` function: returns `cos^-1(1/x)`.
- Added regression tests for the functions implemented.

### DIFF
--- a/age--1.3.0.sql
+++ b/age--1.3.0.sql
@@ -3685,6 +3685,13 @@ IMMUTABLE
 PARALLEL SAFE
 AS 'MODULE_PATHNAME';
 
+CREATE FUNCTION ag_catalog.age_cosec(variadic "any")
+RETURNS agtype
+LANGUAGE c
+IMMUTABLE
+PARALLEL SAFE
+AS 'MODULE_PATHNAME';
+
 CREATE FUNCTION ag_catalog.age_asin(variadic "any")
 RETURNS agtype
 LANGUAGE c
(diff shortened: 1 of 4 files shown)


jrgemignani:
@titoausten Can you justify these additions?

titoausten:
> @titoausten Can you justify these additions?

This PR adds more functions to the current set of trigonometric and arc functions in **AGE** i.e cos, sin, tan, cot, acos, asin, atan and atan2.

https://www.britannica.com/science/trigonometry

https://en.m.wikipedia.org/wiki/Trigonometric_functions

titoausten:
> @titoausten Can you justify these additions?
> 
> Hint:
> 
> PG_FUNCTION_INFO_V1(age_asin);

I would appreciate more details on what to justify please.

titoausten:
> I understand your desire to add additional functions. However, one could ask, where would it stop and how useful it would actually be? In other words, is it necessary?
> 
> I always tell others that our goal is to **fully support the openCypher specifications first**, then possibly more later. As Neo4j created that spec. and is representative of Cypher, look at Neo4j when the openCypher specification is not clear or ambiguous. After that, look to PostgreSQL, as many users of AGE may be PostgreSQL users.
> 
> https://neo4j.com/docs/cypher-manual/current/functions/
> https://www.postgresql.org/docs/current/functions-math.html
> 
> If neither include those functions, I am hard pressed to add them to AGE. Meaning, there needs to be a good justification for adding them, especially if they are trivial to implement. 
> 
> For example,
> 
> The secant of an angle, **sec(a)**, is **1/cos(a)**
> The square of a value, **sqr(3)**, is **3^2**
> 
> These, in my opinion, are unnecessary.
> 

Well understood. Thanks for clarity. The functions created or added should be fully supported by Neo4j and/or PostgreSQL.

PR #1039, opened 2023-07-14
Fix issue 395: ERROR: container is not an agtype array

jrgemignani:
Fixed issue 395 -

 ERROR: container is not an agtype array

This was due to an incorrect order in checking for AGTV_NULL in the collect aggregate transfer function: age_collect_aggtransfn

Added regression tests.


dehowef:
Looks good to me~

PR #1042, opened 2023-07-17
Added agesql python file

Munmud:
### Changes
- Added `agesql.py` file to combine all sql command used in python driver

### Reason for changes
- Improves re usability
- If any command changes with future changes, we don't need to update in all the function.


dehowef:
I appreciate the effort, but having the SQL/Cypher commands written out is more readable for what is actually being sent. Separating this out into another file is unnecessary.

PR #1037, opened 2023-07-13
Fix issue 1033: access operator on vle edge lists

jrgemignani:
This patch fixes an issue where the edge lists generated by the vle function couldn't be resolved by the agtype_access_operator.

Added regression tests.

PR #1050, opened 2023-07-17
Fix issue 1033: access operator on vle edge lists (#1037)

jrgemignani:
This patch fixes an issue where the edge lists generated by the vle function couldn't be resolved by the agtype_access_operator.

Added regression tests.

A big thanks to Taha for showing that this function could be simplified.


dehowef:
Looks good to me~

PR #1063, opened 2023-07-19
Update files, format, and version information

jrgemignani:
Updated the format and version information in files.

Remove the Dockerfile link from the root directory as it is no longer necessary.


dehowef:
Looks good.

PR #1064, opened 2023-07-20
Add checks for array functions to recognize and decode VPC

### LINKED ISSUE #1044: Array functions not recognizing VPC
**Describe the bug**
I encountered an error when trying to get size of vle edge array in where clause which should have executed without error.

- terminal

```pgsql
    CREATE (n)-[:KNOWS {n:'hello'}]->({n:'hello'})
$$) as (a agtype);
```

```pgsql
SELECT * FROM cypher('graph', $$
    MATCH (m)-[v *]->(b) 
    WHERE size(v) >1
    RETURN b
$$) AS (a agtype);
```
```
ERROR:  size() unsupported argument
```

**Expected behavior**
Should have returned
```
 {"id": 281474976710658, "label": "", "properties": {"n": "hello"}}::vertex
```

- latest (PG13 master)

**Additional context**
same error exists in all array functions i.e. they dont recognize VPC.

MuhammadTahaNaveed:
- Added for array functions size, head, last, isEmpty, reverse, agtype_in_operator and agtype_access_slice.

- Fixed a bug where object input to reverse would terminate the server instead of erroring out.

- Added regression tests.

resolves #1044 

**If I am missing any array function other than above listed, please let me know.**


jrgemignani (approved):

src/backend/utils/adt/agtype.c
```diff
@@ -6250,7 +6406,45 @@ Datum age_reverse(PG_FUNCTION_ARGS)
         /* get the agtype argument */
         agt_arg = DATUM_GET_AGTYPE_P(arg);
 
-        if (!AGT_ROOT_IS_SCALAR(agt_arg))
+        /* if it is a binary container, check for a VLE vpc */
+        if (AGT_ROOT_IS_BINARY(agt_arg))
```
rafsun42, line 6410: Can this expression be generalized into a function since it repeats in many places?
```c
        /* if it is a binary container, check for a VLE vpc */
        if (AGT_ROOT_IS_BINARY(agt_arg))
        {
            if (AGT_ROOT_BINARY_FLAGS(agt_arg) == AGT_FBINARY_TYPE_VLE_PATH)
            {
                /* retrieve an array of edges from the vpc */
                agt_arg = agtype_value_to_agtype(
                                          agtv_materialize_vle_edges(agt_arg));
            }
            else
            {
                ereport(ERROR,
                        (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
                        errmsg("binary container must be a VLE vpc")));
            }
        }
```

MuhammadTahaNaveed: actually yes, it can be. Thanks for pointing out. 👍🏻

MuhammadTahaNaveed: Updated!

MuhammadTahaNaveed:
Updated the PR!

jrgemignani:
@MuhammadTahaNaveed 

Do we want a function `agtype *extract_array_if_vpc(agtype *agt_arg)` that returns an `agtype`, instead of an `agtype_value`?

If it is a vpc array, wouldn't it be better to just return it as an `agtype_value ` as we will need to eventually get it in that form anyway?

I'm mentioning this because the `agtype_value_to_agtype ` function is really expensive to use. So, we should be considering ways to avoid converting back and forth. That was the whole idea behind the vpc.

MuhammadTahaNaveed:
> I'm mentioning this because the `agtype_value_to_agtype ` function is really expensive to use. So, we should be considering ways to avoid converting back and forth, if at all possible. That was the whole idea behind the vpc.

oh okay, didn't know that. The current logic of these functions process agtype_container from arg passed as agtype. I think we will have to write extra logic to process `agtype_value AGTV_ARRAY` which is returned by `agtv_materialize_vle_edges`.

rafsun42 (approved):

MuhammadTahaNaveed:
Updated the functions to process agtype_value returned by agtv_materialize_vle_edges. Currently looking if we can simplify agtype_in_operator.

MuhammadTahaNaveed:
Updated.

jrgemignani:
@MuhammadTahaNaveed Looks like this got overlooked. Can you update this to the latest master, plz :)

MuhammadTahaNaveed:
> @MuhammadTahaNaveed Looks like this got overlooked. Can you update this to the latest master, plz :)

@jrgemignani Done.

jrgemignani (approved):

PR #1077, opened 2023-07-22
Cypher randomUUID() function implemented PG13

Allison-E:
* Works the same way as the [cypher randomUUID() function](https://neo4j.com/docs/cypher-manual/current/functions/scalar/#functions-randomuuid).
* Works only on PostgreSQL 13 or higher since the gen_random_uuid() function in PostgreSQL was introduced in version 13.
* Added regression tests for it.


rafsun42:
@Allison-E 

randomUUID is not in open cypher spec.

Also, it can be done like this-
```
(pg-15.4) agedev=# SELECT * from cypher('graph', $$
    RETURN pg_catalog.gen_random_uuid()    
$$) as (result text);
                result                
--------------------------------------
 1d99c524-5746-400c-a94f-c5f89840b2b1

```

PR #1079, opened 2023-07-23
Improved testing of python driver

Munmud:
### Changes
- Added test on test1 `multicolumn returns`
- Deleted unnecessary print
- Added more test on `single column returns`


dehowef:
Outdated

PR #1083, opened 2023-07-24
Added cbrt() function

titoausten:
- Same as PostgreSQL's [cbrt()](https://www.postgresql.org/docs/current/functions-math.html#) mathematical function.
- Added regression tests.


rafsun42:
@titoausten 

This can be done like this-
```
(pg-15.4) agedev=# SELECT * from cypher('graph', $$
    RETURN (pg_catalog.cbrt(27::pg_float8))           
$$) as (result real);
 result 
--------
      3
(1 row)
```

Do you think we should add a function that PG already provides?

titoausten:
> @titoausten 
> 
> This can be done like this-
> ```
> (pg-15.4) agedev=# SELECT * from cypher('graph', $$
>     RETURN (pg_catalog.cbrt(27::pg_float8))           
> $$) as (result real);
>  result 
> --------
>       3
> (1 row)
> ```
> 
> Do you think we should add a function that PG already provides?

Yes, It makes the query clear and concise.

rafsun42:
@titoausten 

How about we find a way to implicity cast types and qualify the function name (with pg_catalog, for example) so the queries can be written concisely using PG's functions?
```
SELECT * from cypher('graph', $$
    RETURN cbrt(27)           
$$) as (result real);
```

This will allow us to use any PG functions while avoiding unnecessary reimplementations.

titoausten:
> @titoausten 
> 
> How about we find a way to implicity cast types and qualify the function name (with pg_catalog, for example) so the queries can be written concisely using PG's functions?
> ```
> SELECT * from cypher('graph', $$
>     RETURN cbrt(27)           
> $$) as (result real);
> ```
> 
> This will allow us to use any PG functions while avoiding unnecessary reimplementations.

This seems fine.

jrgemignani:
Atm, unnecessary.

PR #1084, opened 2023-07-24
Added the toStringList() function - Master

M4rcxs:
- This function is inspired by the "tostringlist" function in OpenCypher.https://neo4j.com/docs/cypher-manual/current/functions/list/#functions-tostringlist
- toStringList() converts a list of values and returns a list of string values. If any values are not convertible to string they will be null in the list returned.
- A list containing the converted elements; depending on the input value a converted value is either a string value or null.
- Also added the regression tests


jrgemignani:
@M4rcxs Please go back and correct all of toStringList, and related PRs. Please make sure you are compiling these locally with the correct flags.

```
src/backend/utils/adt/agtype.c: In function ‘age_tostringlist’:
src/backend/utils/adt/agtype.c:6278:13: warning: format ‘%d’ expects argument of type ‘int’, but argument 3 has type ‘int64’ [-Wformat=]
             sprintf(buffer, "%d", elem->val.int_value);
             ^
src/backend/utils/adt/agtype.c:6213:11: warning: variable ‘float_num’ set but not used [-Wunused-but-set-variable]
     float float_num;
           ^
```

PR #1085, opened 2023-07-25
Added the toStringList() function PG11

M4rcxs:
- This function is inspired by the "tostringlist" function in OpenCypher.https://neo4j.com/docs/cypher-manual/current/functions/list/#functions-tostringlist
- toStringList() converts a list of values and returns a list of string values. If any values are not convertible to string they will be null in the list returned.
- A list containing the converted elements; depending on the input value a converted value is either a string value or null.
- Also added the regression tests


dehowef:
Looks good.

PR #1089, opened 2023-07-25
TEST PR for CI

jrgemignani:
Temporary fix for the CI jobs. For some strange reason, the
DockerHub **latest** build is not working for CI connections. This
is a temporary patch to allow them to run correctly while we look
into the issue.

### DIFF
--- a/.github/workflows/jdbc-driver.yaml
+++ b/.github/workflows/jdbc-driver.yaml
@@ -27,13 +27,13 @@ jobs:
       run: |
         if [[ "$GITHUB_EVENT_NAME" == "push" ]]; then
           if [[ "$GITHUB_REF" == "refs/heads/master" ]]; then
-            echo "TAG=latest" >> $GITHUB_ENV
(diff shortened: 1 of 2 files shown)


MuhammadTahaNaveed:
You have to change for pull event as well.

.github/workflows/jdbc-driver.yaml
```diff
@@ -26,8 +26,8 @@ jobs:
     - name: Set tag based on branch
       run: |
         if [[ "$GITHUB_EVENT_NAME" == "push" ]]; then
-          if [[ "$GITHUB_REF" == "refs/heads/master" ]]; then
-            echo "TAG=latest" >> $GITHUB_ENV
+          if [[ "$GITHUB_REF" == "refs/heads/latest" ]]; then
```
MuhammadTahaNaveed, line 29: it should remain master

jrgemignani: Yeah, I just fixed that.

PR #1114, opened 2023-07-31
Update CIs to trigger only for PG12 branch

MuhammadTahaNaveed:
- Removed duplicate installcheck.yml
- Removed master as trigger branch
- Removed tags as it is not needed for this branch


.github/workflows/installcheck.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
         run: |
           make PG_CONFIG=$HOME/pg12/bin/pg_config installcheck
         continue-on-error: true
-
+
```
MuhammadTahaNaveed, line 45: extra space

jrgemignani (approved):

PR #1119, opened 2023-08-01
Added delete Graph function in python driver

Munmud:
### Changes
- Added delete graph function in python driver

### DIFF
--- a/drivers/python/age/age.py
+++ b/drivers/python/age/age.py
@@ -187,6 +187,13 @@ class Age:
         checkGraphCreated(self.connection, graph)
         self.graphName = graph
         return self
+    
+    def deleteGraph(self, graph:str = None):
+        if (graph == None):
+            graph = self.graphName
+        if (graph == None):
+            raise GraphNotSet
+        deleteGraph(self.connection, self.graphName)
 
     def commit(self):
         self.connection.commit()


jrgemignani:
@Munmud Can you justify why this is necessary? I'm not saying that it isn't, but you need to justify why it is necessary.

dehowef:
There is already a deletegraph function @Munmud

Munmud:
@dehowef  There is no deletegraph function in [AGE class](https://github.com/Munmud/age/blob/c874f647b47f97cecc1d13341e2d684e5f07d3e2/drivers/python/age/age.py#L170). Here I am  using the function in AGE class just like [setUpAge()](https://github.com/Munmud/age/blob/c874f647b47f97cecc1d13341e2d684e5f07d3e2/drivers/python/age/age.py#L178C9-L178C17), 

The reason I am adding this is because it reduces extrawork of importing deletegraph() function when I am using only an instance of AGE class object which already has all the functionality

For example, 
### Without this changes (separate import for deleteGraph)
```py 
from age import Age, deleteGraph
```

### With this changes (only one import)
```py
from age import Age
DSN = "host=localhost port=5432 dbname=postgres user=moontasir password=pass"
graphName = 'bitnine_global_inic'
c = Age()
c.connect(graph=graphName, dsn=DSN)
c.deleteGraph()
```

PR #1125, opened 2023-08-08
Fix issue 1124: Segmentation fault when using specific tables

jrgemignani:
This is a fix for issue 1124 -

Segmentation fault when using specific names for tables.

This issue is due to an oversight of some very specific types of tables that are parameterless functions. The logic to walk through these nodes didn't expect them and would crash due to them.

This fix adds in code to skip over these types of tables.

Added regression tests.


dehowef:
Looks good to me.

MuhammadTahaNaveed (approved):
Looks good 👍🏻

PR #1133, opened 2023-08-11
Fix issue 1000: Indexes created on WHERE

jrgemignani:
This fixes the issue with applying indexes on the WHERE clause.

It does this by modifying the volatility flag for 3 functions -

    _agtype_build_vertex
    _agtype_build_edge
    _label_name

All three were verified to met the specifications for IMMUTABLE.


dehowef:
These changes look good to me.

PR #1135, opened 2023-08-11
PR to merge master branch into PG14 branch

jrgemignani:
**PLEASE DO NOT MERGE** until thoroughly reviewed. Additionally, this merge needs to be appropriately squashed before merging.

PR for moving the PG14 branch up to the current master branch.

It contains all of the latest work in the master branch, which is currently at PostgreSQL version 14. When this PR is merged PG14 will be the latest and represent PostgreSQL version 14 as the master moves to PostgreSQL version 15.

### DIFF
--- a/src/backend/utils/adt/agtype.c
+++ b/src/backend/utils/adt/agtype.c
@@ -31,6 +31,7 @@
 #include "postgres.h"
 
 #include <math.h>
+#include <float.h>
 
 #include "access/genam.h"
 #include "access/heapam.h"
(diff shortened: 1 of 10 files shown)


src/backend/utils/adt/agtype.c
```diff
@@ -2377,7 +2378,9 @@ static agtype_value* agtype_build_map_as_agtype_value(FunctionCallInfo fcinfo)
     nargs = extract_variadic_args(fcinfo, 0, true, &args, &types, &nulls);
 
     if (nargs < 0)
-        PG_RETURN_NULL();
+    {
+        return NULL;
```
rafsun42, line 2382: Should this be `PG_RETURN_NULL()`? This macro sets `fcinfo->isnull = true;` before returning `0`.

dehowef: Good Question

dehowef (commented):
Looks good to me, but @rafsun42 does raise a valid question. After reviewing that change, I'd like to know as well why the line in question was changed to just return NULL.

dehowef (approved):

PR #1143, opened 2023-08-15
Fix memory leak issues with age_load

rafsun42:
In this patch, all agtype_values, that are allocated while processing the csv file, are free'd.


jrgemignani:
PG15 is not the branch for version 15, yet. Just use the master for now.

Issue #1159, opened 2023-08-18
Server terminates for SET plus-equal when the original properties has a non-scalar value


Zainab-Saad:
**Describe the bug**
Updating properties of a vertex/edge (having a non-scalar value) with SET plus-equal terminates the server 

**How are you accessing AGE (Command line, driver, etc.)?**
Command Line

**What data setup do we need to do?**
Create 2 vertices with non-scalar property values
```
SELECT * FROM cypher('graph', $$
CREATE (n: A {map: {a : 1}}), (m: B {lst: [1, 2]})
$$) AS (result agtype);
```

**What is the command that caused the error?**
This query terminated the server (same for vertex with label `B`)
```
SELECT * FROM cypher('graph', $$
MATCH (n: A) SET n += {bool: true} RETURN n
$$) AS (result agtype);
```

**Expected behavior**
Should update the properties as shown [here](https://github.com/apache/age/blob/master/regress/expected/cypher_set.out) 

**Environment (please complete the following information):**
AGE master (This problem exists for other branches too)

**Additional context**
SET plus-equal works fine for scalar property values
```
SELECT * FROM cypher('graph', $$
CREATE (n: C {num : 1}) RETURN n
$$) AS (result agtype);
```

```
SELECT * FROM cypher('graph', $$
MATCH (n: C) SET n += {bool: true} RETURN n
$$) AS (result agtype);
                                         result                                         
----------------------------------------------------------------------------------------
 {"id": 1407374883553281, "label": "C", "properties": {"num": 1, "bool": true}}::vertex
(1 row)
```

rafsun42:
@Zainab-Saad Are you working on it?

Zainab-Saad:
@rafsun42 I have created a PR for this, could you review that?

rafsun42:
@Zainab-Saad This query didn't terminate on my machine (Master, PG15). What PG version are you using?
```
SELECT * FROM cypher('graph', $$
MATCH (n: A) SET n += {bool: true} RETURN n
$$) AS (result agtype);
```

Zainab-Saad:
@rafsun42 version is PG 15.1. with the data setup I have provided, this query terminated.

rafsun42:
@Zainab-Saad 
Is this the expected result for your data setup? Could you put this on the "Expected behavior" section in case anyone else wants to reproduce the issue.
```
agedev=# SELECT * FROM cypher('graph', $$
MATCH (n: A) SET n += {bool: true} RETURN n
[...]

PR #1160, opened 2023-08-18
Fix Issue # 1159: Server terminates for SET plus-equal

Zainab-Saad:
This PR is a fix for Issue #1159

### DIFF
--- a/regress/expected/cypher_set.out
+++ b/regress/expected/cypher_set.out
@@ -787,6 +787,11 @@ SELECT * FROM cypher('cypher_set_1', $$ CREATE (a:Robert {name:'Robert', role:'m
 ---
 (0 rows)
 
+SELECT * FROM cypher('cypher_set_1',$$ CREATE (a: VertexA {map: {a: 1, b: {c: 2, d: []}, c: [{d: -100, e: []}]}, list: [1, "string", [{a: []}, [[1, 2]]]], bool: true, num: -1.9::numeric, str: "string"})$$) as (a agtype);
+ a 
+---
+(0 rows)
+
 -- test copying properties between entities
 SELECT * FROM cypher('cypher_set_1', $$
     MATCH (at {name: 'Andy'}), (pn {name: 'Peter'})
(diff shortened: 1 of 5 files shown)


rafsun42:
@Zainab-Saad 

1. Could this be solved without using the function `agtype_value_to_agtype`?
2. Does this issue also appear in other part of the codebase? It seems the root of the issue is pushing an object type with push_agtype_value is prevented by `Assert` function. Without assertion check, the issue disappears without any patch.

Zainab-Saad:
@rafsun42 
1. `agtype_value_to_agtype` is used to initialize the iterator
2. This is what I thought at first to remove the Assertion, but keeping this assertion makes more sense to me because if the value is an object it should not be pushed right away as we might be missing other checks.
For example, the assertions for WAGT_KEY, WAGT_ELEM etc. So any non-scalar should be iterated when making an `agtype_value`.
The approach I have used here for copying the properties is the same used for appending new properties into the `parsed_agtype_value` and the later one did not produce any bug for non-scalar values because it treats non-scalar values as `AGTV_BINARY` which in `push_agtype_value` is not directly passed to `push_agtype_value_scalar` to be pushed into agtype_value rather is iterated through and sent to `push_agtype_value_scalar` for pushing to agtype_value.

rafsun42:
@Zainab-Saad 
Appending new properties uses this approach because `new properties` is already in `agtype` format. However, `original properties` is not. Can it be done without converting it to an `agtype` format? This can be expensive.

Zainab-Saad:
@rafsun42 
Mostly such implementations are done with `agtype_value` as iterator needs this format.
Is writing our own such implementation with `agtype` (without iterator) an option since I have not found any such case yet in the codebase.

rafsun42:
@Zainab-Saad 
`MATCH (n) CREATE (x {prop: n}) RETURN x`
You can look into how n (which is a vertex json object) gets inserted into the x's properties?

Zainab-Saad:
@rafsun42 
Are you referring to `agtype_build_map_nonull`/`agtype_build_map` ? Because this is the function that would make the properties for the vertex to be created.
These functions utilize `datum_to_agtype` which given a datum converts in `agtype_value` format to be added to `agtype_in_state`

Zainab-Saad:
@rafsun42 
As far as I have researched, such implementations are done with the extra serialization step when the deserialized `agtype_value` is provided or in the other case `agtype`  structure itself is provided. `alter_property_value` is an example for this.
So for doing this directly with `agtype_value`, I have written a helper function that recursively copies the values and works with the non-scalar values i.e; objects and arrays as well (I have tested with the added tests in cypher_set.sql in this PR)
```
agtype_value *alter_properties(agtype_value *original_properties,
                               agtype *new_properties)
{
    ...
    parsed_agtype_value = push_agtype_value(&parse_state, WAGT_BEGIN_OBJECT,
                                            NULL);
    // Copy original properties.
    if (original_properties)
    {
        if (original_properties->type != AGTV_OBJECT)
        {
            ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
                            errmsg("a map is expected")));
        }

        copy_original_properties(parse_state, original_properties,
                                &parsed_agtype_value, true);
    }
...
}
```
This one might need some refactoring
```
/*
 * helper function to copy the original properties for SET plus-equal
 * given the original properties are in agtype_value format
 * this function helps skip the serialization step
 */
void copy_original_properties(agtype_parse_state* pstate, agtype_value* props,
                            agtype_value **copied_properties, bool is_top_level)
{
    int i = 0;

    check_stack_depth();

    if (props->type == AGTV_OBJECT)
    {
        if (!is_top_level)
        {
            *copied_properties = push_agtype_value(&pstate, WAGT_BEGIN_OBJECT,
                                                    NULL);
        }

        for(; i < props->val.object.num_pairs; i ++)
        {
            agtype_pair *pair = props->val.object.pairs + i;
            *copied_properties = push_agtype_value(&pstate, WAGT_KEY,
                                                    &pair->key);

            if (IS_A_AGTYPE_SCALAR(&pair->value))
            {
                *copied_properties = push_agtype_value(&pstate, WAGT_VALUE,
                                                    &pair->value);
            }
            else
            {
                copy_original_properties(pstate, &pair->value,
                                        copied_properties, false);
            }
        }

        if (!is_top_level)
        {
            *copied_properties = push_agtype_value(&pstate, WAGT_END_OBJECT,
                                                    NULL);
        }
    }
    else if (props->type == AGTV_ARRAY)
    {
        *copied_properties = push_agtype_value(&pstate, WAGT_BEGIN_ARRAY,
                                                    NULL);

        for (; i < props->val.array.num_elems; i++)
        {
            agtype_value elem = props->val.array.elems[i];

            if (IS_A_AGTYPE_SCALAR(&elem))
            {
                *copied_properties = push_agtype_value(&pstate, WAGT_ELEM,
                                                    &elem);
            }
            else
            {
                copy_original_properties(pstate, &elem,
                                        copied_properties, false);
            }
        }

        *copied_properties = push_agtype_value(&pstate, WAGT_END_ARRAY,
                                                    NULL);
    }
    else
    {
        ereport(ERROR, (errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
                            errmsg("invalid type provided")));
    }
}
```

Zainab-Saad:
@rafsun42 Sorry for the repeated mentions, is there any update on this PR? What do you think about the helper function I have added above, does it solve the issue?

rafsun42:
@Zainab-Saad If it passes your tests, it should solve the issue?

Only one suggestion, should we give `copy_original_properties` a more generic name (i.e. `copy_agtype_value`), since it can be used to copy any agtype_value object (not just properties in SET clause).

Zainab-Saad:
@rafsun42 
Thanks for the suggestion.
I have updated the PR

rafsun42 (approved):

rafsun42:
@Zainab-Saad Could you please create this PR for other branches as well?

PR #1203, opened 2023-08-30
Minor improvements to the golang driver

fervidnerd:
* Migration to Makefile, as it is the most common approach to building golang projects.
* Upgrade antlr parser to 1.13 and migrate imports
* Commit generated parser golang files (these are needed when importing this library for other projects). Otherwise we'll get compilation errors.


jrgemignani:
@fervidnerd Sorry for the delay, would it be possible to get this PR updated to the current master?

PR #1206, opened 2023-09-01
simplify nested condition

cmucheru:
Simplify nested function for readability


src/backend/catalog/ag_catalog.c
```diff
@@ -178,24 +177,19 @@ static void object_access(ObjectAccessType access, Oid class_id, Oid object_id,
      * The above applies to DROP TABLE command too.
      */
 
+    if (drop_arg->dropflags & PERFORM_DELETION_INTERNAL)
```
rafsun42, line 180: Does this change break for the case `class_id == RelationRelationId`?

rafsun42:
Why are the files in `bc` extension?

jrgemignani:
Why so many modified files?

src/backend/catalog/ag_catalog.c
cmucheru: No. It does not break.

rafsun42: Could you elaborate more? In this case, the function calls `delete_label` first before returning. But with your change, it returns without calling that function. Does it have side effect?

rafsun42:
@cmucheru The tests are failing at `catalog.sql`.

cmucheru:
@rafsun42 thankyou. I am unfamiliar with the tests. I will like to check my code again to pass the tests in catalog.sql .

PR #1207, opened 2023-09-05
Fix Issue#1159: Server terminates for SET plus-equal

Zainab-Saad:
This PR fixes the [issue](https://github.com/apache/age/issues/1159) with `alter_properties()` (helper function used in SET clause) where while copying the original properties into the new agtype object, server terminated for non-scalar values in properties. As a fix, a function `copy_agtype_value` is added to recursively copy the original properties and deals with non-scalar values (object, array) cases.


rafsun42 (approved):

rafsun42:
@Zainab-Saad Could you send one for PG15 as well?

PR #1227, opened 2023-09-11
Code Style Update for agtype_util.c

MatheusFarias03:
Updated some of the lines from agtype_util.c so that it follows Apache AGE's coding style. When the bodies of if/else statement contain a single line, the braces are omitted. For single line comments that used "/*...*/", the "//..." was used instead.

### DIFF
--- a/src/backend/utils/adt/agtype_util.c
+++ b/src/backend/utils/adt/agtype_util.c
@@ -109,7 +109,7 @@ agtype *agtype_value_to_agtype(agtype_value *val)
 
     if (IS_A_AGTYPE_SCALAR(val))
     {
-        /* Scalar value */
+        // Scalar value 
         agtype_parse_state *pstate = NULL;
         agtype_value *res;
         agtype_value scalar_array;
@@ -125,9 +125,8 @@ agtype *agtype_value_to_agtype(agtype_value *val)
         out = convert_to_agtype(res);
     }
     else if (val->type == AGTV_OBJECT || val->type == AGTV_ARRAY)
-    {
         out = convert_to_agtype(val);
-    }
+    
     else
     {
         Assert(val->type == AGTV_BINARY);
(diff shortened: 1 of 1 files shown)


jrgemignani:
@MatheusFarias03 Where did you see this? Our coding style for if/else is to always use braces regardless. Additionally, we can use either // or /**/.

MatheusFarias03:
On AGE's website, the [Code Convention](https://age.apache.org/contribution/guide) page. 

> If all the bodies of if/else statement contain a single line, omit braces.

> For multi-line comments, use C style multi-line comments. For single-line comments, use C++ style single-line comments. See below.

jrgemignani:
@MatheusFarias03 It does need to be changed. But, let me have @dehowef verify this is the correct location.

dehowef:
@MatheusFarias03 , yeah actually we do use braces regardless of how many lines an if/else statement is. This was something we decided awhile back. If you could change that page that would be great. I'm not sure why the page states this o_O

Thanks for bringing this to our attention

MatheusFarias03:
I'm right on it! The `.html` file is a bit confusing though, feels like everything is stuck together and it's pretty hard to see which thing is which. Gonna update that also.

jrgemignani:
@MatheusFarias03 If you can also modify the coding standards to state that all braces are on their own line, no exceptions. Items inside braces are indented 4 spaces, no exceptions. The `switch ` command example in the documentation is wrong.

Good:
```
struct xyz
{
    int x;
    int y;
    int z;
}

if (blah == blahBlah)
{
    some_function_call(var);
}

void some_function(int var)
{
    some_main_function(var);
}
```
Bad:
```
struct xyz {
int x;
int y;
int z;
}

if (blah == blahBlah) {
some_function_call(var);
}

void some_function(int var) {
some_main_function(var);
}
```

MatheusFarias03:
Here it is: [the PR](https://github.com/apache/age-website/pull/197/files)

PR #1236, opened 2023-09-19
Refactor the IN operator to use '= ANY()' syntax

dehowef:
This change reduces use of internal function calls from the
IN implementation to optimize performance. This also changes IN behavior to correctly return NULL upon NULL elements included in the list that IN checks against.

Added and corrected regression tests.

Co-authored by: Josh Innis <JoshInnis@gmail.com>

### DIFF
--- a/regress/sql/expr.sql
+++ b/regress/sql/expr.sql
@@ -121,14 +121,19 @@ SELECT * FROM cypher('expr',
 $$RETURN [1,3,5,[2,4,6]] IN ['str', 1, 1.0, true, null, [1,3,5,[2,4,6]]]$$) AS r(c boolean);
 SELECT * FROM cypher('expr',
 $$RETURN {bool: true, int: 1} IN ['str', 1, 1.0, true, null, {bool: true, int: 1}, [1,3,5,[2,4,6]]]$$) AS r(c boolean);
+SELECT * FROM cypher('expr',
+$$RETURN 1 IN [1.0, [NULL]]$$) AS r(c boolean);
+SELECT * FROM cypher('expr',
+$$RETURN [NULL] IN [1.0, [NULL]]$$) AS r(c boolean);
 -- should return SQL null, nothing
+SELECT * FROM cypher('expr',
+$$RETURN true IN NULL $$) AS r(c boolean);
 SELECT * FROM cypher('expr',
 $$RETURN null IN ['str', 1, 1.0, true, null]$$) AS r(c boolean);
 SELECT * FROM cypher('expr',
 $$RETURN null IN ['str', 1, 1.0, true]$$) AS r(c boolean);
 SELECT * FROM cypher('expr',
 $$RETURN 'str' IN null $$) AS r(c boolean);
--- should all return false
 SELECT * FROM cypher('expr',
 $$RETURN 0 IN ['str', 1, 1.0, true, null]$$) AS r(c boolean);
 SELECT * FROM cypher('expr',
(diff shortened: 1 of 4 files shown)


src/backend/parser/cypher_expr.c
```diff
@@ -489,25 +489,143 @@ static Node *transform_AEXPR_OP(cypher_parsestate *cpstate, A_Expr *a)
                            a->location);
 }
 
+//static Node *transform_AEXPR_IN(cypher_parsestate *cpstate, A_Expr *a)
```
jrgemignani, line 492: No need to comment out the code, just remove it. Unless you have a particular reason for leaving it in and commented out.

jrgemignani (changes requested):

src/backend/parser/cypher_expr.c
dehowef: I have corrected this change, thank you for pointing that out

regress/sql/expr.sql
```diff
@@ -139,6 +144,19 @@ SELECT * FROM cypher('expr',
 $$RETURN [1,3,5,[2,4,5]] IN ['str', 1, 1.0, true, null, [1,3,5,[2,4,6]]]$$) AS r(c boolean);
 SELECT * FROM cypher('expr',
 $$RETURN {bool: true, int: 2} IN ['str', 1, 1.0, true, null, {bool: true, int: 1}, [1,3,5,[2,4,6]]]$$) AS r(c boolean);
+-- should return false
+SELECT * FROM cypher('expr',
+$$RETURN 'str' IN ['StR', 1, true]$$) AS r(c boolean);
+SELECT * FROM cypher('expr',
+$$RETURN 2 IN ['StR', 1. true]$$) AS r(c boolean);
```
rafsun42, line 151: Did  you mean a `,` after `1` instead of a `.`?

src/backend/parser/cypher_expr.c
```diff
@@ -491,23 +491,120 @@ static Node *transform_AEXPR_OP(cypher_parsestate *cpstate, A_Expr *a)
...
+                        errmsg("object of IN must be a list")));
+    }
 
-    /* get the agtype_access_slice function */
-    func_in_oid = get_ag_func_oid("agtype_in_operator", 2, AGTYPEOID,
-                                  AGTYPEOID);
+    Assert(is_ag_node(a->rexpr, cypher_list));
+
+    // If the operator is <>, combine with AND not OR.
+    if (strcmp(strVal(linitial(a->name)), "<>") == 0)
```
rafsun42, line 523: Should we wrap the blocks with curly braces?

src/backend/parser/cypher_expr.c
```diff
@@ -491,23 +491,120 @@ static Node *transform_AEXPR_OP(cypher_parsestate *cpstate, A_Expr *a)
...
+
+    foreach(l, (List *) rexpr->elems)
+    {
+        Node *rexpr = transform_cypher_expr_recurse(cpstate, lfirst(l));
 
-    result = makeFuncExpr(func_in_oid, AGTYPEOID, args, InvalidOid, InvalidOid,
-                          COERCE_EXPLICIT_CALL);
+        rexprs = lappend(rexprs, rexpr);
+                if (contain_vars_of_level(rexpr, 0))
+                        rvars = lappend(rvars, rexpr);
```
rafsun42, line 540: Here, as well.

src/backend/parser/cypher_expr.c
```diff
@@ -491,23 +491,120 @@ static Node *transform_AEXPR_OP(cypher_parsestate *cpstate, A_Expr *a)
...
+        Node *rexpr = (Node *) lfirst(l);
+        Node *cmp;
+
+        // Ordinary scalar operator
+        cmp = (Node *) make_op(pstate, a->name, copyObject(lexpr), rexpr,
+                               pstate->p_last_srf, a->location);
+
+        cmp = coerce_to_boolean(pstate, cmp, "IN");
+        if (result == NULL)
+            result = cmp;
```
rafsun42, line 600: Wrap with curle braces here too?

src/backend/parser/cypher_expr.c
dehowef: Good catches, I made the changes you suggested~

jrgemignani (approved):

src/backend/parser/cypher_expr.c
rafsun42: Thank you!

rafsun42 (approved):

PR #1237, opened 2023-09-20
PG 16 initial version

muhammadshoaib:
Fixed empty string handling

Previously, the PG 16 outToken emitted `NULL` for an empty string, but now it emits an empty string `""`. Consequently, our cypher read function required modification to properly decode the empty string as a plain token, thereby addressing the comparison issue. 

Compared the branches and added the necessary changes so that the query's `rteperminfos` variable doesn't stay `NULL`.

Fix missing include varatt.h (causing undefined symbol `VARDATA_ANY)` & Fixing some test cases of the failings

Added missing include `varatt.h` to fix the undefined symbol while loading age into postgresql because usage of `VARDATA_ANY` needs to import `varatt.h` in PG16

Modified initialisation of `ResultRelInfo` and removed unnecessary `RTEs`

Compared the branches and added the necessary changes so that the query's `rteperminfos` variable doesn't stay `NULL`.

Modified initialisation of `ResultRelInfo` and removed unnecessary `RTEs`

One of the problems that we were facing was related to the `ResultRelInfo` pointing at the wrong `RTE` via its `ri_RangeTableIndex`. The `create_entity_result_rel_info()` function does not have the capability of setting the `ri_RootResultRelInfo` to the correct `ResultRelInfo` node because it does not have access to the `ModifyTableState` node. The solution for this was to set the third argument in `InitResultRelInfo()` to be zero instead of `list_length(estate->es_range_table)`.

In the `update_entity_tuple()` function, when we call `table_tuple_update()` and assign the returned value to the result variable, the buffer variable receives the value of 0.
Made a workaround so that the original value isn't lost.

This is a work in progress for the new field that was added to the struct Var called `varnullingrels`. According to the documentation, this field is responsible for marking the Vars as nullable, if they are coming from a `JOIN`, either `LEFT JOIN`, `RIGHT JOIN`, or `FULL OUTER JOIN`. The changes were made following an "optional match" clause which is being treated as a `LEFT JOIN` from our extension.

A function `markRelsAsNulledBy` is added because its internal in Postgres and doesn't belong in a header file, therefore it can't be exported. This function is added before the creation of the Vars from the `make_vertex_expr` and `make_edge_expr`, to correctly mark the specific `PNSI` as `nullable`, so later in the planner stage, the Vars will be correctly nulled.

Fix incorrect typecasting in `agtype_to_graphid` function. 

Fix incorrect returns to the fuction `_label_name`, `_ag_build_vertex` and `_ag_build_edge`.

### DIFF
--- a/.github/workflows/go-driver.yml
+++ b/.github/workflows/go-driver.yml
@@ -2,10 +2,10 @@ name: Go Driver Tests
 
 on:
   push:
-    branches: [ "master", "PG15" ]
+    branches: [ "master", "PG16" ]
 
   pull_request:
-    branches: [ "master", "PG15" ]
+    branches: [ "master", "PG16" ]
 
 jobs:
   build:
@@ -26,14 +26,14 @@ jobs:
         if [[ "$GITHUB_EVENT_NAME" == "push" ]]; then
           if [[ "$GITHUB_REF" == "refs/heads/master" ]]; then
             echo "TAG=latest" >> $GITHUB_ENV
-          elif [[ "$GITHUB_REF" == "refs/heads/PG15" ]]; then
-            echo "TAG=PG15_latest" >> $GITHUB_ENV
+          elif [[ "$GITHUB_REF" == "refs/heads/PG16" ]]; then
+            echo "TAG=PG16_latest" >> $GITHUB_ENV
           fi
         elif [[ "$GITHUB_EVENT_NAME" == "pull_request" ]]; then
           if [[ "$GITHUB_BASE_REF" == "master" ]]; then
             echo "TAG=latest" >> $GITHUB_ENV
-          elif [[ "$GITHUB_BASE_REF" == "PG15" ]]; then
-            echo "TAG=PG15_latest" >> $GITHUB_ENV
+          elif [[ "$GITHUB_BASE_REF" == "PG16" ]]; then
+            echo "TAG=PG16_latest" >> $GITHUB_ENV
           fi
         fi
 
--- a/.github/workflows/installcheck.yaml
+++ b/.github/workflows/installcheck.yaml
@@ -2,32 +2,32 @@ name: Build / Regression
 
 on:
   push:
-    branches: [ 'master', 'PG15' ]
+    branches: [ 'master', 'PG16' ]
   pull_request:
-    branches: [ 'master', 'PG15' ]
+    branches: [ 'master', 'PG16' ]
 
 jobs:
   build:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Get latest commit id of PostgreSQL 15
+      - name: Get latest commit id of PostgreSQL 16
         run: |
-          echo "PG_COMMIT_HASH=$(git ls-remote git://git.postgresql.org/git/postgresql.git refs/heads/REL_15_STABLE | awk '{print $1}')" >> $GITHUB_ENV
+          echo "PG_COMMIT_HASH=$(git ls-remote git://git.postgresql.org/git/postgresql.git refs/heads/REL_16_STABLE | awk '{print $1}')" >> $GITHUB_ENV
 
-      - name: Cache PostgreSQL 15
+      - name: Cache PostgreSQL 16
         uses: actions/cache@v3
-        id: pg15cache
+        id: pg16cache
         with:
-          path: ~/pg15
-          key: ${{ runner.os }}-v1-pg15-${{ env.PG_COMMIT_HASH }}
+          path: ~/pg16
+          key: ${{ runner.os }}-v1-pg16-${{ env.PG_COMMIT_HASH }}
 
-      - name: Install PostgreSQL 15
-        if: steps.pg15cache.outputs.cache-hit != 'true'
+      - name: Install PostgreSQL 16
+        if: steps.pg16cache.outputs.cache-hit != 'true'
         run: |
-          git clone --depth 1 --branch REL_15_STABLE git://git.postgresql.org/git/postgresql.git ~/pg15source
-          cd ~/pg15source
-          ./configure --prefix=$HOME/pg15 CFLAGS="-std=gnu99 -ggdb -O0" --enable-cassert
+          git clone --depth 1 --branch REL_16_STABLE git://git.postgresql.org/git/postgresql.git ~/pg16source
+          cd ~/pg16source
+          ./configure --prefix=$HOME/pg16 CFLAGS="-std=gnu99 -ggdb -O0" --enable-cassert
           make install -j$(nproc) > /dev/null
 
       - uses: actions/checkout@v3
@@ -35,12 +35,12 @@ jobs:
       - name: Build
         id: build
         run: |
-          make PG_CONFIG=$HOME/pg15/bin/pg_config install -j$(nproc)
+          make PG_CONFIG=$HOME/pg16/bin/pg_config install -j$(nproc)
 
       - name: Regression tests
         id: regression_tests
         run: |
-          make PG_CONFIG=$HOME/pg15/bin/pg_config installcheck
+          make PG_CONFIG=$HOME/pg16/bin/pg_config installcheck
         continue-on-error: true
 
       - name: Dump regression test errors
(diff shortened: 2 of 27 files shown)


.github/workflows/go-driver.yml
```diff
@@ -2,10 +2,10 @@ name: Go Driver Tests
 
 on:
   push:
-    branches: [ "master", "PG15" ]
+    branches: [ "master", "pg16" ]
```
jrgemignani, line 5: All `pg16 ` entries need to be uppercase `PG16`. Except where that `pg15` is lowercase.

.github/workflows/installcheck.yaml
```diff
@@ -2,9 +2,9 @@ name: Build / Regression
 
 on:
   push:
-    branches: [ 'master', 'PG15' ]
+    branches: [ 'master', 'pg16' ]
```
jrgemignani, line 5: All `pg16 ` entries need to be uppercase `PG16`. Except where that `pg15` is lowercase.

.github/workflows/jdbc-driver.yaml
```diff
@@ -2,10 +2,10 @@ name: JDBC Driver Tests
 
 on:
   push:
-    branches: [ "master", "PG15" ]
+    branches: [ "master", "pg16" ]
```
jrgemignani, line 5: All `pg16 ` entries need to be uppercase `PG16`. Except where that `pg15` is lowercase.

.github/workflows/nodejs-driver.yaml
```diff
@@ -2,10 +2,10 @@ name: Nodejs Driver Tests
 
 on:
   push:
-    branches: [ "master", "PG15" ]
+    branches: [ "master", "pg16" ]
```
jrgemignani, line 5: All `pg16 ` entries need to be uppercase `PG16`. Except where that `pg15` is lowercase.

.github/workflows/python-driver.yaml
```diff
@@ -2,10 +2,10 @@ name: Python Driver Tests
 
 on:
   push:
-    branches: [ "master", "PG15" ]
+    branches: [ "master", "pg16" ]
```
jrgemignani, line 5: All `pg16 ` entries need to be uppercase `PG16`. Except where that `pg15` is lowercase.

jrgemignani (changes requested):

jrgemignani:
@muhammadshoaib only a partial review, more to come.

src/backend/utils/adt/agtype.c
```diff
@@ -1965,7 +1965,7 @@ Datum _agtype_build_path(PG_FUNCTION_ARGS)
      */
     if (nargs >= 1 && nargs <= 3)
     {
-        int i = 0;
+        i = 0;
```
jrgemignani, line 1968: If this `i` is only used within this `if` block, then let's only have it defined here. If this `i` is used outside of this `if` block for other things, then let's change the interior `i` to some other variable name. Let's try to keep the scope of the variable as close to where it is used and let's not duplicate variables unless absolutely necessary.

src/backend/parser/cypher_gram.y
```diff
@@ -1543,12 +1543,10 @@ expr:
              * supported
              */
             else
-            {
```
jrgemignani, line 1546: Why remove the surrounding braces?

src/backend/parser/cypher_gram.y
```diff
@@ -1543,12 +1543,10 @@ expr:
...
                 ereport(ERROR,
                         (errcode(ERRCODE_SYNTAX_ERROR),
                          errmsg("invalid indirection syntax"),
                          ag_scanner_errposition(@1, scanner)));
-            }
```
jrgemignani, line 1551: Why remove the surrounding braces?

src/backend/parser/cypher_expr.c
```diff
@@ -251,14 +251,11 @@ static Node *transform_A_Const(cypher_parsestate *cpstate, A_Const *ac)
         d = boolean_to_agtype(boolVal(&ac->val));
         break;
     default:
-        if (ac->isnull)
-        {
```
jrgemignani, line 255: Why are the braces being removed?

src/backend/parser/cypher_expr.c
```diff
@@ -251,14 +251,11 @@ static Node *transform_A_Const(cypher_parsestate *cpstate, A_Const *ac)
...
+        if (ac->isnull) {
```
jrgemignani, line 254: Braces do *not* go at the end of the line of code, they go on the following line.,

jrgemignani (changes requested):
Please go through the **entire** patch and make sure that -

1. No braces are being removed for just one line `if/else`.
2. Braces are on their **own** line, by themselves.

jirutka:
We are planning to release Alpine v3.19 in November and I’d really love to include AGE in this release. The problem is that is must be built against the latest PostgreSQL, i.e. version 16.

jrgemignani:
> We are planning to release Alpine v3.19 in November and I’d really love to include AGE in this release. The problem is that is must be built against the latest PostgreSQL, i.e. version 16.

Our PG16 branch is currently in the final stages of review. It should be ready in a ~2 weeks. @muhammadshoaib ?

jrgemignani:
@muhammadshoaib Please review the **build** log -

```
src/backend/executor/cypher_merge.c: In function ‘exec_cypher_merge’:
src/backend/executor/cypher_merge.c:475:26: warning: declaration of ‘econtext’ shadows a previous local [-Wshadow=compatible-local]
  475 |             ExprContext *econtext = node->ss.ps.ps_ExprContext;
      |                          ^~~~~~~~
src/backend/executor/cypher_merge.c:320:[18](https://github.com/apache/age/actions/runs/6769752846/job/18396782787?pr=1237#step:6:19): note: shadowed declaration is here
  3[20](https://github.com/apache/age/actions/runs/6769752846/job/18396782787?pr=1237#step:6:21) |     ExprContext *econtext = css->css.ss.ps.ps_ExprContext;
      |                  ^~~~~~~~
```

src/backend/utils/adt/agtype.c
```diff
@@ -2268,9 +2271,12 @@ Datum _agtype_build_edge(PG_FUNCTION_ARGS)
 
     if (fcinfo->args[0].isnull)
     {
+        PG_RETURN_NULL();
+        /*
```
jrgemignani, line 2275: Don't leave in commented out code please :)

src/backend/utils/adt/agtype.c
jrgemignani: Still needs to be fixed.

jrgemignani (changes requested):

src/backend/parser/cypher_clause.c
```diff
@@ -5788,7 +5800,7 @@ transform_create_cypher_new_node(cypher_parsestate *cpstate,
     if (!label_exists(node->label, cpstate->graph_oid))
     {
         List *parent;
-        RangeVar *rv;
+        //RangeVar *rv;
```
jrgemignani, line 5803: Commented out?

src/backend/parser/cypher_clause.c
```diff
@@ -6765,7 +6779,7 @@ transform_merge_cypher_edge(cypher_parsestate *cpstate, List **target_list,
     if (edge->label && !label_exists(edge->label, cpstate->graph_oid))
     {
         List *parent;
-        RangeVar *rv;
+        //RangeVar *rv;
```
jrgemignani, line 6782: Commented out?

src/backend/parser/cypher_clause.c
```diff
@@ -6884,7 +6898,7 @@ transform_merge_cypher_node(cypher_parsestate *cpstate, List **target_list,
     if (node->label && !label_exists(node->label, cpstate->graph_oid))
     {
         List *parent;
-        RangeVar *rv;
+        //RangeVar *rv;
```
jrgemignani, line 6901: Commented out?

jrgemignani (changes requested):

jrgemignani (approved):

jrgemignani:
@jirutka Could you respond to #1652 ?

pras-baghel:
Please share steps and needed binaries for installing age extension on alpine-3.19.1 image.

PR #1238, opened 2023-09-20
Simplified some if statements

MatheusFarias03:
Some if statements were reduced using a ternary operator instead.

### DIFF
--- a/src/backend/parser/cypher_clause.c
+++ b/src/backend/parser/cypher_clause.c
@@ -701,14 +701,7 @@ transform_cypher_union_tree(cypher_parsestate *cpstate, cypher_clause *clause,
     else if (cmp->op == SETOP_UNION)
     {
         Assert(cmp->larg != NULL && cmp->rarg != NULL);
-        if (cmp->order_by || cmp->limit || cmp->skip)
-        {
-            isLeaf = true;
-        }
-        else
-        {
-            isLeaf = false;
-        }
+        isLeaf = (cmp->order_by || cmp->limit || cmp->skip) ? true : false;
     }
     else
     {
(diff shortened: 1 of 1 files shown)


dehowef:
Hi, thank you for these changes, but I am of the opinion that if-else statements are more readable in most of these cases. I have done very limited research into if ternary operators are faster than if-else statements, but from what I have seen on the web, they are almost equivalent in speed when it comes to using C. @jrgemignani @rafsun42 @muhammadshoaib if you'd like to give your input as well, that'd be good

jrgemignani:
@MatheusFarias03 I tend to agree with @dehowef here. Ternary operators are good in some places but tend to make it more difficult to read and debug in others. We do use them in some places and don't ban their usage. But, generally, I find them unnecessary most of the time and I tend to avoid them, unless necessary.

Here is a SO article about them -

[https://stackoverflow.com/questions/160218/to-ternary-or-not-to-ternary](https://stackoverflow.com/questions/160218/to-ternary-or-not-to-ternary)

MatheusFarias03:
Oh ok! I've also created some PRs like this one, should I close them all?

jrgemignani:
@MatheusFarias03 I hate to say yes because I [we] appreciate your work and desire to make AGE better,... but, yeah,... I don't think these add value. That is my opinion, at least. @dehowef ? @rafsun42

rafsun42:
@MatheusFarias03 My opinion- I appreciate your effort, but I would prefer readability if performance is unchanged.

MatheusFarias03:
Okay folks! I'll close them, no problem! :+1:

dehowef:
@jrgemignani @MatheusFarias03 I wrote transform_cypher_union, which largely referenced the transformSetOperationStmt code in Postgres. I followed a lot of their coding style choices. I think that a good point the thread John linked also points out that nested if/else blocks also allow for debuggers to place breakpoints on certain statements, which condensed ternary operators don't allow for. That is another dimension of "readability" that I think is important. Thanks for understanding @MatheusFarias03 , we really do appreciate your enthusiasm and effort.

dehowef:
I think that this preference in code will differ from project to project. Maybe some projects would prefer the condensed statements and reduced number of lines. But, on a project like AGE where we oftentimes are in the debugger looking at code, and oftentimes code someone else wrote, I think it's nice to have somewhat more atomized code blocks to aid with that.

PR #1248, opened 2023-09-24
shorten comparison

cmucheru:
Combined the nested condition with the main for single check of the logic.


dehowef:
Hey this simplification removes the lfirst() function which is necessary for extracting the pointer value from each iteration of the list. It breaks the code. Closed

PR #1256, opened 2023-09-28
Add missing dependency in cypher_expr.c

dehowef:
The function contain_vars_of_level() was missing a header. Added that header to prevent a warning during compilation.


rafsun42 (approved):
Perfect. The warning is gone now.

PR #1261, opened 2023-10-05
Update README.md

vinay-agens:
Updated the Discord Link.


rafsun42:
The link redirects to discord.com.

Also, the link should be updated in two places in the README, not just one.

rafsun42:
@vinay-agens The second link still redirects to home. Try removing the trailing slash.

rafsun42 (approved):

PR #1262, opened 2023-10-05
Extend EXPLAIN and add config param to switch transformation of property filter

rafsun42:


jrgemignani (approved):

MuhammadTahaNaveed:
@rafsun42 Shoudn't this work for filters applied using `WHERE` clause as well? Currently its not....
```
explain_review=# SET age.enable_containment = off;
SET
explain_review=# SELECT * FROM cypher('test_enable_containmentt', $$ EXPLAIN MATCH (x {name:'Bob'}) RETURN x $$) as (a agtype);
                                             QUERY PLAN                                             
----------------------------------------------------------------------------------------------------
 Seq Scan on _ag_label_vertex x  (cost=0.00..28.03 rows=6 width=32)
   Filter: (agtype_access_operator(VARIADIC ARRAY[properties, '"name"'::agtype]) = '"Bob"'::agtype)
(2 rows)

explain_review=# SELECT * FROM cypher('test_enable_containmentt', $$ EXPLAIN MATCH (x) WHERE x.name='Bob' RETURN x $$) as (a agtype);
                                             QUERY PLAN                                             
----------------------------------------------------------------------------------------------------
 Seq Scan on _ag_label_vertex x  (cost=0.00..28.03 rows=6 width=32)
   Filter: (agtype_access_operator(VARIADIC ARRAY[properties, '"name"'::agtype]) = '"Bob"'::agtype)


```
```
explain_review=# SET age.enable_containment = on;
SET
explain_review=# SELECT * FROM cypher('test_enable_containmentt', $$ EXPLAIN MATCH (x {name:'Bob'}) RETURN x $$) as (a agtype);
                             QUERY PLAN                             
--------------------------------------------------------------------
 Seq Scan on _ag_label_vertex x  (cost=0.00..25.00 rows=1 width=32)
   Filter: (properties @> '{"name": "Bob"}'::agtype)
(2 rows)

explain_review=# SELECT * FROM cypher('test_enable_containmentt', $$ EXPLAIN MATCH (x) WHERE x.name='Bob' RETURN x $$) as (a agtype);
                                             QUERY PLAN                                             
----------------------------------------------------------------------------------------------------
 Seq Scan on _ag_label_vertex x  (cost=0.00..28.03 rows=6 width=32)
   Filter: (agtype_access_operator(VARIADIC ARRAY[properties, '"name"'::agtype]) = '"Bob"'::agtype)
(2 rows)

```

MuhammadTahaNaveed:
> @MuhammadTahaNaveed 
> 
> Good question. Currently, WHERE clause is not affected by `enable_containment` param.The original patch, from which the idea was taken from, seem to only convert containment to access operator to support Btree indexing, but not the other direction. So, WHERE clause, which uses only access operator, is not converted to containment when the parameter is on.

Ok... Thank you for the explanation....

MuhammadTahaNaveed (approved):
Lgtm

PR #1266, opened 2023-10-10
Fix sanitizer errors

antamel:
Hello!

In the build with address and undefined behavior sanitizers, my colleague
Andrew Bille <andrewbille@gmail.com> found several errors
in the current master branch.

The build was made with options like that:
```
CC=gcc-11
CXX=g++-11
CPPFLAGS="-ggdb -O0 -g3 -fno-omit-frame-pointer -fsanitize=address -fsanitize=undefined -fsanitize=alignment -fno-sanitize-recover=all -fno-sanitize=nonnull-attribute -fstack-protector"
LDFLAGS='-fsanitize=address -fsanitize=undefined -static-libasan -static-libubsan'
ASAN_OPTIONS=detect_leaks=0:abort_on_error=1:halt_on_error=1:disable_coredump=0:strict_string_checks=1:check_initialization_order=1:strict_init_order=1:detect_odr_violation=0
```

There are several "load of misaligned address" sanitizer errors in the log
during installcheck tests with PG at REL_15_STABLE (6db384f2f9).
See [master-at-a04fd69b.zip](https://github.com/apache/age/files/12856017/master-at-a04fd69b.zip) attached. (Other errors are masked by this, about them below.)
Some of these errors are not related to age, is a bug of the sanitizer itself
(see  https://gcc.gnu.org/bugzilla/show_bug.cgi?id=111382).
The other ones i tried to fix with the [Fix sanitizer aligment errors with memcpy.](https://github.com/apache/age/commit/de3633d21e6b879176b3e57ffc377db9aa3edc3f) commit.

After that "misaligned" errors disappeared and other errors became visible.
Please see postmaster.log in the [after-first-fix-at-de3633d2.zip](https://github.com/apache/age/files/12856019/after-first-fix-at-de3633d2.zip).
The second  commit [Fix sanitizer "invalid values for type '_Bool' " errors with palloc0.](https://github.com/apache/age/commit/3afbad612251b3e95db1ea0b8daac9344754c2c3) fixes
runtime errors: "load of value xxx, which is not a valid value for type '_Bool'"
while the third [Fix sanitizer heap-buffer-overflow errors during csv files parsing.](https://github.com/apache/age/commit/2840ce1fdbd2027a9a83ffa06334c703fc7e6e24) one
fixes "heap-buffer-overflow on address.." one during csv file parsing.

On applying the first three patches, another error heap-buffer-overflow
became visible. See log in [before-last-fix-at-2840ce1f.zip](https://github.com/apache/age/files/12856022/before-last-fix-at-2840ce1f.zip).
The last commit [Save the metadata representing by the extensible node with the null t…](https://github.com/apache/age/commit/71f445a61d604f4ca8a2bc5f364f4c0077e09ac5) 
fixes it add all tests began to pass without errors with
REL_15_STABLE and REL_14_STABLE branches.

Would be happy for any comments and concerns.

With kind regards,

--
Anton A. Melnikov
Postgres Professional: http://www.postgrespro.com
The Russian Postgres Company

### DIFF
--- a/src/backend/utils/adt/agtype_ext.c
+++ b/src/backend/utils/adt/agtype_ext.c
@@ -59,7 +59,7 @@ bool ag_serialize_extended_type(StringInfo buffer, agtentry *agtentry,
         /* copy in the int_value data */
         numlen = sizeof(int64);
         offset = reserve_from_buffer(buffer, numlen);
-        *((int64 *)(buffer->data + offset)) = scalar_val->val.int_value;
+        memcpy( buffer->data + offset, &scalar_val->val.int_value, numlen);
 
         *agtentry = AGTENTRY_IS_AGTYPE | (padlen + numlen + AGT_HEADER_SIZE);
         break;
@@ -70,7 +70,7 @@ bool ag_serialize_extended_type(StringInfo buffer, agtentry *agtentry,
         /* copy in the float_value data */
         numlen = sizeof(scalar_val->val.float_value);
         offset = reserve_from_buffer(buffer, numlen);
-        *((float8 *)(buffer->data + offset)) = scalar_val->val.float_value;
+        memcpy(buffer->data + offset, &scalar_val->val.int_value, numlen);
 
         *agtentry = AGTENTRY_IS_AGTYPE | (padlen + numlen + AGT_HEADER_SIZE);
         break;
@@ -156,12 +156,12 @@ void ag_deserialize_extended_type(char *base_addr, uint32 offset,
     {
     case AGT_HEADER_INTEGER:
         result->type = AGTV_INTEGER;
-        result->val.int_value = *((int64 *)(base + AGT_HEADER_SIZE));
+        memcpy(&result->val.int_value, base + AGT_HEADER_SIZE, sizeof(int64));
         break;
 
     case AGT_HEADER_FLOAT:
         result->type = AGTV_FLOAT;
-        result->val.float_value = *((float8 *)(base + AGT_HEADER_SIZE));
+        memcpy(&result->val.float_value, base + AGT_HEADER_SIZE, sizeof(float8));
         break;
 
     case AGT_HEADER_VERTEX:
--- a/src/backend/utils/adt/agtype_ext.c
+++ b/src/backend/utils/adt/agtype_ext.c
@@ -195,7 +195,7 @@ static void ag_deserialize_composite(char *base, enum agtype_value_type type,
     //offset container by the extended type header
     char *container_base = base + AGT_HEADER_SIZE;
 
-    r = palloc(sizeof(agtype_value));
+    r = palloc0(sizeof(agtype_value));
 
     it = agtype_iterator_init((agtype_container *)container_base);
     while ((tok = agtype_iterator_next(&it, r, true)) != WAGT_DONE)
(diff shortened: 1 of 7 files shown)


src/backend/utils/adt/agtype_ext.c
```diff
@@ -59,7 +59,7 @@ bool ag_serialize_extended_type(StringInfo buffer, agtentry *agtentry,
         /* copy in the int_value data */
         numlen = sizeof(int64);
         offset = reserve_from_buffer(buffer, numlen);
-        *((int64 *)(buffer->data + offset)) = scalar_val->val.int_value;
+        memcpy( buffer->data + offset, &scalar_val->val.int_value, numlen);
```
jrgemignani, line 62: I don't see why this is necessary, maybe it is just me, could you please elaborate?

src/backend/utils/adt/agtype_ext.c
```diff
@@ -70,7 +70,7 @@ bool ag_serialize_extended_type(StringInfo buffer, agtentry *agtentry,
         /* copy in the float_value data */
         numlen = sizeof(scalar_val->val.float_value);
         offset = reserve_from_buffer(buffer, numlen);
-        *((float8 *)(buffer->data + offset)) = scalar_val->val.float_value;
+        memcpy(buffer->data + offset, &scalar_val->val.int_value, numlen);
```
jrgemignani, line 73: I don't see why this is necessary, maybe it is just me, could you please elaborate?

src/backend/utils/adt/agtype_ext.c
```diff
@@ -156,12 +156,12 @@ void ag_deserialize_extended_type(char *base_addr, uint32 offset,
     {
     case AGT_HEADER_INTEGER:
         result->type = AGTV_INTEGER;
-        result->val.int_value = *((int64 *)(base + AGT_HEADER_SIZE));
+        memcpy(&result->val.int_value, base + AGT_HEADER_SIZE, sizeof(int64));
```
jrgemignani, line 159: I don't see why this is necessary, maybe it is just me, could you please elaborate?

src/backend/utils/adt/agtype_ext.c
```diff
@@ -156,12 +156,12 @@ void ag_deserialize_extended_type(char *base_addr, uint32 offset,
...
         break;
 
     case AGT_HEADER_FLOAT:
         result->type = AGTV_FLOAT;
-        result->val.float_value = *((float8 *)(base + AGT_HEADER_SIZE));
+        memcpy(&result->val.float_value, base + AGT_HEADER_SIZE, sizeof(float8));
```
jrgemignani, line 164: I don't see why this is necessary, maybe it is just me, could you please elaborate?

jrgemignani (changes requested):
A lot of these changes appear unnecessary and need justification before being considered and added.

rafsun42:
+1 for the palloc0 changes as PostgreSQL recommends it as well:

https://www.postgresql.org/docs/current/xfunc-c.html
> Always zero the bytes of your structures using memset (or allocate them with palloc0 in the first place). Even if you assign to each field of your structure, there might be alignment padding (holes in the structure) that contain garbage values. Without this, it's difficult to support hash indexes or hash joins, as you must pick out only the significant bits of your data structure to compute a hash. The planner also sometimes relies on comparing constants via bitwise equality, so you can get undesirable planning results if logically-equivalent values aren't bitwise equal.

rafsun42:
@jrgemignani Sounds good. I am fine with doing either of palloc0 or memcpy.

PR #1277, opened 2023-10-14
Fix python driver's test

KTaskn:
Fixed a problem with python driver testing where arguments were not working properly.

### DIFF
--- a/drivers/python/test_age_py.py
+++ b/drivers/python/test_age_py.py
@@ -19,18 +19,16 @@ import decimal
 import age
 import argparse
 
-DSN = "host=localhost port=5432 dbname=postgres user=dehowefeng password=agens"
-TEST_HOST = "localhost"
-TEST_PORT = 5432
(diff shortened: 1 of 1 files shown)


dehowef:
It works fine for me. Can you show me how it didn't work for you

dehowef:
Maybe you can remove "dehowefeng" in the DSN string though and replace that with something generic

KTaskn:
@dehowef 

It is not consistent with the readme, it should be able to be changed dynamically from the arguments as in the readme.
Also, it is strange that your handle is in the code.

https://github.com/apache/age/blob/5fbe064a913f03dd13058cc21b7b7fb0cdf910cb/drivers/python/README.md?plain=1#L48C1-L57C4

aked21:
@dehowef, can you comment if this can be merged?

dehowef:
Closed for breaking the regression tests.

PR #1288, opened 2023-10-24
Fix Issue 945 - incorrect count(*) return values

jrgemignani:
Fixed issue 945 where count(*) would have incorrect return values.

NOTE -

We need to re-evaluate if we want to use output_node or not. If output_node is set to false, then it basically short circuits match for instances where a variable isn't specified -

    MATCH () RETURN 0;
    MATCH () MATCH () RETURN 0;

While, on the surface, this appears to be a good way to improve execution time of commands that won't do anything, it also causes chained commands to not work correctly. This is because a match without a variable will still feed its tuples to the next stage(s) even though they won't necessarily be sent to the output. For example, with count(*) -

    MATCH () RETURN count(*);
    MATCH () MATCH RETURN count(*);

With output_node set to false, it won't send the tuples. We will likely need to remove all of the output_node logic. However, this needs to be reviewed. For now, we just set it to true and update the output of the regression tests.

Updated regression tests to accomodate the change. Added new regression tests to cover overlooked cases.


jrgemignani:
Please carefully review the regression tests as the change impacted the order of some results.

rafsun42 (approved):

regress/expected/cypher_match.out
```diff
@@ -2310,13 +2310,31 @@ SELECT * FROM cypher('cypher_match', $$ MATCH () MATCH (_{name: "Dave"}) RETURN
...
+ {"id": 281474976710667, "label": "", "properties": {"name": "Dave"}}::vertex
+ {"id": 281474976710667, "label": "", "properties": {"name": "Dave"}}::vertex
+ {"id": 281474976710667, "label": "", "properties": {"name": "Dave"}}::vertex
+ {"id": 281474976710667, "label": "", "properties": {"name": "Dave"}}::vertex
+ {"id": 281474976710667, "label": "", "properties": {"name": "Dave"}}::vertex
+ {"id": 281474976710667, "label": "", "properties": {"name": "Dave"}}::vertex
+ {"id": 281474976710667, "label": "", "properties": {"name": "Dave"}}::vertex
+ {"id": 281474976710667, "label": "", "properties": {"name": "Dave"}}::vertex
+ {"id": 281474976710667, "label": "", "properties": {"name": "Dave"}}::vertex
+(10 rows)
```
MuhammadTahaNaveed, line 2338: This one needs to be looked into... previously `1 row` but now `10 rows`

regress/expected/cypher_match.out
```diff
@@ -2328,7 +2346,16 @@ SELECT * FROM cypher('cypher_match', $$ MATCH () MATCH (my_age_default_{name: "D
...
```
MuhammadTahaNaveed, line 2359: This one needs to be looked into... previously `1 row` but now `10 rows`

regress/expected/cypher_match.out
```diff
@@ -2310,13 +2310,31 @@ SELECT * FROM cypher('cypher_match', $$ MATCH () MATCH (_{name: "Dave"}) RETURN
...
+ 0
+ 0
+ 0
+ 0
+ 0
+ 0
+ 0
+ 0
+ 0
+(10 rows)
```
MuhammadTahaNaveed, line 2323: This one needs to be looked into... previously `1 row` but now `10 rows`

jrgemignani: Yes, please try on Neo4j to see what it outputs :)

regress/expected/cypher_match.out
jrgemignani: Yes, please try on Neo4j to see what it outputs :)

regress/expected/cypher_match.out
jrgemignani: Yes, please try on Neo4j to see what it outputs :)

jrgemignani:
@rafsun42 @MuhammadTahaNaveed @dehowef Please note and try the queries that output more than expected on Neo4j.

regress/expected/cypher_match.out
MuhammadTahaNaveed: > Yes, please try on Neo4j to see what it outputs :)

Yes.. already doing that

regress/expected/cypher_match.out
MuhammadTahaNaveed: @jrgemignani just checked and actually the results are now aligned with neo4j.....
image

MuhammadTahaNaveed (approved):
Looks good to me.

PR #1311, opened 2023-10-26
Modify COUNT() to output agtype

dehowef:
Modified the make_function_expr logic to wrap the PG COUNT() function with a cast to agtype. This enables COUNT() to be used as a subquery in CASE.

Also added logic for casting for future PG function additions.

This modification passes all regression tests. Also added regression tests for COUNT in CASE statements.

### DIFF
--- a/src/backend/parser/cypher_gram.y
+++ b/src/backend/parser/cypher_gram.y
@@ -225,6 +225,9 @@ static Node *make_typecast_expr(Node *expr, char *typecast, int location);
 
 // functions
 static Node *make_function_expr(List *func_name, List *exprs, int location);
+static Node *make_star_function_expr(List *func_name, List *exprs, int location);
+static Node *make_distinct_function_expr(List *func_name, List *exprs, int location);
+static FuncCall *wrap_pg_funccall_to_agtype(Node* fnode, char *type, int location);
 
 // setops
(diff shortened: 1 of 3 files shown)


src/backend/parser/cypher_gram.y
```diff
@@ -2274,25 +2274,180 @@ static Node *make_function_expr(List *func_name, List *exprs, int location)
...
+             * transform layer and allows us to know which functions are ours.
+             */
+            funcname = func_name;
+
+            /* build the function call */
+            fnode = makeFuncCall(funcname, exprs, COERCE_SQL_SYNTAX, location);
+        }
+    }
+    /* all other functions are passed as is */
+    else
```
jrgemignani, line 2301: Missing {}

jrgemignani (changes requested):
One minor issue with else missing {}

Also, would it be possible to combine `make_function_expr`, `make_star_function_expr`, and `make_distinct_function_expr `into one? with flags passed? There appears to be a lot of duplicate code between them.

dehowef:
> One minor issue with else missing {}
> 
> Also, would it be possible to combine `make_function_expr`, `make_star_function_expr`, and `make_distinct_function_expr `into one? with flags passed? There appears to be a lot of duplicate code between them.

Added the parentheses

jrgemignani (approved):

rafsun42 (approved):

Issue #1303, opened 2023-10-26
Server crashes when trying to execute SELECT * FROM agtype(null);


saygoodbyye:
**Describe the bug**
Server crashes when executing the following SQL-script

**How are you accessing AGE (Command line, driver, etc.)?**
Accessing AGE through command line.

**What data setup do we need to do?**
Apache AGE (PG15 branch) with PostgreSQL (REL_15_STABLE).

**What is the necessary configuration info needed?**
My PostgreSQL configuration:
```
./configure CFLAGS=" -Og" --enable-tap-tests --enable-debug --enable-cassert
```

**What is the command that caused the error?**
```pgsql
create extension age;
SET search_path TO ag_catalog;
SELECT * FROM agtype(null);
```
backtrace:
```
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007f6038a42866 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007f6038a268b7 in __GI_abort () at ./stdlib/abort.c:79
#5  0x000055c450cd4e41 in ExceptionalCondition (conditionName=conditionName@entry=0x7f602f7ce463 "HeapTupleIsValid(proctup)", 
    errorType=errorType@entry=0x7f602f7cc1d7 "FailedAssertion", fileName=fileName@entry=0x7f602f7ce428 "src/backend/utils/ag_func.c", 
    lineNumber=lineNumber@entry=49) at assert.c:69
#6  0x00007f602f7c6dba in is_oid_ag_func (func_oid=<optimized out>, func_name=func_name@entry=0x7f602f7ccd9c "cypher") at src/backend/utils/ag_func.c:49
#7  0x00007f602f7949ba in is_func_cypher (funcexpr=<optimized out>) at src/backend/parser/cypher_analyze.c:329
#8  0x00007f602f7949f4 in is_rte_cypher (rte=<optimized out>) at src/backend/parser/cypher_analyze.c:305
[...]

PR #1317, opened 2023-10-27
Fix issue 1303: Server crashes on executing SELECT * FROM agtype(null);

Zainab-Saad:
This issue is a fix for #1303 

    - Add additional checks for Const expression nodes for which the
      walker code would crash.

    - The server crashes for few other expr types too including Const,
      OpExpr, Var, BoolExpr and CoerceViaIO for which checks are added

    - Add regression tests

### DIFF
--- a/regress/expected/expr.out
+++ b/regress/expected/expr.out
@@ -7305,6 +7305,152 @@ SELECT results, pg_typeof(user) FROM cypher('issue_1124', $$ MATCH (u) RETURN u 
 (1 row)
 
 --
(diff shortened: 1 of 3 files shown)


jrgemignani (approved):
Good job Zainab!!

dehowef (approved):
Looks good

jrgemignani:
@Zainab-Saad Now you can move it to PG14, 13, 12, & 11 :)

Keep up the good work!

PR #1345, opened 2023-11-02
Implement EXISTS subquery for CASE

dehowef:
Modified logic of EXISTS to wrap itself in a agtype cast. This allows EXISTS to work with in CASE statements, among other implementations that may require an agtype.

Also renamed the wrapper function to be more generic to denote its usage in other implementations if need be.

Changed some variable names in regression tests for clarity.

Added regression tests for EXISTS.

### DIFF
--- a/src/backend/parser/cypher_gram.y
+++ b/src/backend/parser/cypher_gram.y
@@ -227,7 +227,7 @@ static Node *make_typecast_expr(Node *expr, char *typecast, int location);
 static Node *make_function_expr(List *func_name, List *exprs, int location);
 static Node *make_star_function_expr(List *func_name, List *exprs, int location);
 static Node *make_distinct_function_expr(List *func_name, List *exprs, int location);
-static FuncCall *wrap_pg_funccall_to_agtype(Node* fnode, char *type, int location);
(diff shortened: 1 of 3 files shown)


src/backend/parser/cypher_gram.y
```diff
@@ -1742,12 +1742,17 @@ expr_func_subexpr:
...
     | EXISTS '(' property_value ')'
         {
-            $$ = make_function_expr(list_make1(makeString("exists")),
-                                    list_make1($3), @2);
+            FuncCall *n;
+            n = makeFuncCall(list_make1(makeString("exists")),
+                                    list_make1($3), COERCE_SQL_SYNTAX, @2);
+
+            $$ = (Node *)node_to_agtype((Node *)n, "boolean", @2);
+
```
jrgemignani, line 1754: extra line?

jrgemignani (changes requested):
minor cosmetic

src/backend/parser/cypher_gram.y
dehowef: addressed

jrgemignani (approved):

PR #1346, opened 2023-11-02
Fix Issue 1329 - agtype_to_int4 crash (#1339)

jrgemignani:
Fixed issue 1329 where `agtype_to_int`<8,4,2> and `agtype_to_int4_array` crashed due to not properly checking input.

As these functions take "any" input, the input has to be properly checked before casting it to a specific type. The input section assumed it was agtype, which caused crashes for non-agtypes.

The functions `agtype_to_int`<8,4,2> will convert non-agtypes into agtype. However, there were no regression tests for this.

The functions `agtype_to_int`<8,4,2> will convert non-agtypes to agtype ints but, did not for their string equivs. Meaning, passing a ('true') or ('3.14') would fail but, passing a (true) or (3.14) would not. This has been corrected for all 3 functions.

TODO -
The function `agtype_to_int4_array` only takes agtype, currently, and we should consider allowing it to take "any" types.

Added regression tests.
Added missing regression tests.

Conflicts:
	regress/expected/agtype.out
	regress/expected/expr.out
	src/backend/utils/adt/agtype.c

### DIFF
--- a/regress/expected/agtype.out
+++ b/regress/expected/agtype.out
@@ -2216,6 +2216,91 @@ SELECT agtype_to_int8(agtype_in('false'));
               0
 (1 row)
 
+-- should return SQL NULL
+SELECT agtype_to_int8(agtype_in('null'));
+ agtype_to_int8 
+----------------
+               
+(1 row)
+
+SELECT agtype_to_int8(NULL);
+ agtype_to_int8 
+----------------
+               
+(1 row)
+
+-- non agtype input
+SELECT agtype_to_int8(1);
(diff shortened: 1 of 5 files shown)


rafsun42:
What does Conflicts mean in the commit message? Are these merge conflict?

jrgemignani:
> What does Conflicts mean in the commit message? Are these merge conflict?

@rafsun42 There were merge conflicts. I wanted to leave it in so that it was known.

rafsun42 (approved):

PR #1355, opened 2023-11-04
Hotfix for DockerHub on PG11

jrgemignani:
PR 1353 applied a change that added an additional header file. This caused the build on DockerHub to fail.

This removes the included file and should correct the error. This is only for PG11.

`#include "commands/label_commands.h"`


dehowef:
Nice

PR #1385, opened 2023-11-08
Fix DockerHub warning messages for PG14_latest

jrgemignani:
Fixed DockerHub warning messages for the build PG14_latest. One was due to an Assert statement that used a variable that was not used for anything else. I changed it to an if with error log output. The other was due to an uninitialized variable.


dehowef:
Looks good

PR #1387, opened 2023-11-09
Fix DockerHub warning messages for PG12_latest

jrgemignani:
Fixed DockerHub warning message for the build PG12_latest. This was due to an Assert statement that used a variable that was not used for anything else. I changed it to an if with error log output instead.


dehowef:
Looks good

PR #1392, opened 2023-11-13
Remove unnecessary #include in src/backend/utils

jrgemignani:
Removed unnecessary #includes in src/backend/utils files.


dehowef:
Looks good

Issue #1399, opened 2023-11-15
Filter NOT EXISTS leads to an error if a relation label does not exists as database table


nupis-DanielS:
**Describe the bug**
We run in an error if we query for a non existing relationship. If the filter contains the label of the relationship it will return "Variable `foo` does not exist". When the relationship label is not used everything works as expected. It looks like that the error occurs only when the database table for the label is not present..

**How are you accessing AGE (Command line, driver, etc.)?**
- JDBC
- pg (Node)

**What data setup do we need to do?**
```pgsql
/
```

**What is the necessary configuration info needed?**
- /

**What is the command that caused the error?**
```pgsql
SELECT * FROM cypher('property-graph', $$ 
  MATCH (foo)
  WHERE NOT exists((foo)-[:BAR]->())
  RETURN count(foo) > 0
$$) as (c agtype)
```
```
Variable `foo` does not exist
```

**Expected behavior**
false

**Environment (please complete the following information):**
- Version: 1.4.0
- PostgreSQL: 14.6 (Ubuntu 14.6-1.pgdg22.04+1)

**Additional context**
Without non existing relation label it works as expected (result -> false).

jimenasansav:
Hi @nupis-DanielS! Thank you for letting us know about this issue. Our technical team is already going through it and will provide a solution soon. In the meantime, please feel free to reach out if you have any other questions.

jrgemignani:
@nupis-DanielS I will look into this.

jrgemignani:
@nupis-DanielS PR #1400 should resolve this once reviewed and merged.

jrgemignani:
@nupis-DanielS The fix is now in the master branch. If you could verify that it satisfactorily resolves your issue and close it, that would be great.

nupis-DanielS:
> @nupis-DanielS The fix is now in the master branch. If you could verify that it satisfactorily resolves your issue and close it, that would be great.

Hi @jrgemignani . We would love to test this but we use the "PG14" Branch. I think we will create a separate PG14 branch and cherry pick your commit. We close this issue if everything works as expected.

Thank you

MuhammadTahaNaveed:
> Hi @jrgemignani . We would love to test this but we use the "PG14" Branch. I think we will create a separate PG14 branch and cherry pick your commit. We close this issue if everything works as expected.
> 
> Thank you

The fix is now on PG14 branch as well. It would be great if you could check it out.

nupis-DanielS:
[...]

PR #1400, opened 2023-11-16
Fix issue #1399 EXISTS doesn't handle non-existent labels

jrgemignani:
Fixed issue #1399 where EXISTS doesn't handle non-existent labels. However, the issue actually is due to the extra processing (read: extra distance between the AST nodes) that prevented the transform logic from being able to resolve variables created in previous clauses further than one parent up.

Added logic to allow the `find_variable` function to search up the parent parse nodes.

Added regression tests to cover these cases.

### DIFF
--- a/regress/expected/cypher_match.out
+++ b/regress/expected/cypher_match.out
@@ -2830,6 +2830,167 @@ SELECT * FROM cypher('cypher_match', $$ MATCH p1=(n {name:'Dave'})-[]->() MATCH 
 (1 row)
 
 --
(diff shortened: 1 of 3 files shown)


MuhammadTahaNaveed:
Issue number in commit description is wrong. I should be 1399 instead of 1339.

MuhammadTahaNaveed (approved):
LGTM 👍🏻 .

jrgemignani:
> @jrgemignani Issue number in commit description is wrong. It should be 1399 instead of 1339.

Oops

dehowef (approved):
Looks good to me~

Issue #1398, opened 2023-11-14
Server crashes when trying to execute SELECT * FROM cypher(...) AS (result agtype);


saygoodbyye:
**Describe the bug**
Server crashes when executing the following SQL-script

**How are you accessing AGE (Command line, driver, etc.)?**
Accessing AGE through command line.

**What data setup do we need to do?**
Apache AGE (master branch) with PostgreSQL (REL_15_STABLE).

**What is the necessary configuration info needed?**
My PostgreSQL configuration:
```
./configure CFLAGS=" -Og" --enable-tap-tests --enable-debug --enable-cassert
```

**What is the command that caused the error?**
```pgsql
CREATE EXTENSION age;
SET search_path TO ag_catalog;
SELECT create_graph('cypher_delete');
SELECT * FROM cypher('cypher_delete', $$CREATE (n:v)-[:e]->(:v)$$) AS (a agtype);
SELECT * FROM cypher('cypher_delete', $$CREATE (n:v)-[:e]->(:v)$$) AS (a agtype);
SELECT * FROM cypher('cypher_delete', $$MATCH(n)-[e]->(m) SET e.i = 1 DETACH DELETE n RETURN e$$) AS (a agtype);
SELECT * FROM cypher('cypher_delete', $$ MATCH (n) RETURN vertex_stats(n) $$) AS (result agtype);
```
backtrace:
```
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=<optimized out>, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007f3656242866 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007f36562268b7 in __GI_abort () at ./stdlib/abort.c:79
#5  0x0000558ed8bd9249 in ExceptionalCondition (conditionName=conditionName@entry=0x7f364cf02173 "found", errorType=errorType@entry=0x7f364cf011d7 "FailedAssertion", 
    fileName=fileName@entry=0x7f364cf09fc0 "src/backend/utils/adt/age_global_graph.c", lineNumber=lineNumber@entry=358) at assert.c:69
[...]

PR #1412, opened 2023-11-21
Fix issue #1398 - SET followed by DELETE does not delete

rafsun42:
In the DELETE executor, tuples were being locked\deleted using the command ID
returned by GetCurrentCommandId(). However, the original tuples were retrieved
using a command ID stored in the estate.

To fix it- before the retrieval of a tuple, estate command IDs are set using
GetCurrentCommandId() so it matches if the tuple is to be deleted later.

Additional changes:
-------------------
Fixed an incorrect cypher_delete test. The query
`MATCH (n1)-[e]->() DELETE n1, e RETURN n1` should be able to delete n1 and e
without requiring DETACH DELETE.

TODO:
-----
It may be a good idea to audit the executors for any inconsistent use of
command IDs.


jrgemignani (approved):
Good job Rafsun!

PR #1441, opened 2023-12-08
Fix issue 1219 - MERGE not seeing previous clause var

jrgemignani:
Fixed issue 1219 where MERGE did not see the previous clause's variable.

This description is a bit misleading as the transform phase did see the variable and was able to use it. However, the planner phase removed the variable by replacing it with a NULL Const. This caused MERGE to see a NULL Const for the previous tuple, generating incorrect results. However, this only occurred for very specific cases.

Fx:    MATCH (x) MERGE (y {id: id(x)})                      -- worked
        MATCH (x) MERGE (y {id: id(x)}) RETURN y     -- didn't
        MATCH (x) MERGE (y {id: id(x)}) RETURN x, y  -- worked

The change impacted no current regression tests and involved wrapping all explicitly defined variables' target entries with a volatile wrapper.

Added new regression tests.


rafsun42 (approved):
The issue is resolved. Perfect!

dehowef:
Looks great.

dehowef (approved):

MuhammadTahaNaveed (approved):
Looks good to me 👍

PR #1466, opened 2023-12-29
Update age_global_graph.out

danielwambo:
Added Begin and commit wrapping all the process to enable rollback when needed


jrgemignani:
@danielwambo These output files are only modified in tandem with their sql files.

PR #1483, opened 2024-01-03
Update regression test to prevent bogus failure

rafsun42:
Changes:
--------
 - The ORDER BY clause is added to a query that was causing row-ordering difference.

 - The debug_parallel_query variable is set to 'regress' around a query that may
   choose parallel worker and therefore, may output a notice from a parallel worker.

   Setting this option will suppress notices from parallel workers and result in a
   consistent query output regardless of parallel worker being selected or not. Note
   that, this option also forces a query to choose parallel worker if it is safe.


rafsun42:
This PR is applied via #1678 .

PR #1484, opened 2024-01-04
Add checks for array functions to recognize and decode VPC (#1064)

jrgemignani:
Author: Muhammad Taha Naveed <m.taha.naveed27@gmail.com>

* Add checks for array functions to recognize and decode VPC

- Added for array functions size, head, last, isEmpty, reverse, agtype_in_operator and agtype_access_slice.

- Fixed a bug where object input to reverse would terminate the server instead of erroring out.

- Added regression tests.

* Fix issue with in operator tranformation

- We need to build a function call here if the rexpr is already tranformed. It can be already tranformed cypher_list as columnref.


muhammadshoaib (approved):
Approved

rafsun42 (approved):

PR #1515, opened 2024-01-22
Fix issue#1513 - Invalid variable reuse in CREATE and MERGE clause

Zainab-Saad:
Allow the reuse of a previously declared vertex variable only if the succeeding CREATE/MERGE vertex has associated edges
i.e 
CREATE (n) CREATE (n)                   -- invalid
CREATE (n) CREATE (n)-[:edge]->() -- valid

Fixed another invalid variable reuse where in a CREATE path, edge variable could be duplicated which resulted in ambiguous reference error

Added regression tests


jrgemignani (approved):
Looks good! Great job Zainab!

MuhammadTahaNaveed (approved):
Looks good to me 👍🏻

PR #1506, opened 2024-01-12
Add template for upgrading between versions of Apache AGE

jrgemignani:
Added a template to be used when upgrading between versions of Apache AGE.

This template is to be added to for all additions, deletions, and modifications to the sql files that define operators, types, and functions.

Modified .gitignore to allow these changes.

    modified:   .gitignore
    new file:   age--1.5.0--y.y.y.sql

PR #1516, opened 2024-01-22
Add optional parameter '=' in property constraints

MuhammadTahaNaveed:
- The '=' operator checks if the original property value(as a whole) is equal to the given value.

  e.g 

      MATCH (n {school:{addr:{city:'Toronto'}}}) tranforms into

      either(in case age.enable_containment is off)
      properties.school.addr.city = 'Toronto'

      or(in case age.enable_containment is on)
      properties @> {school:{addr:{city:'Toronto'}}}

      But MATCH (n ={school:{addr:{city:'Toronto'}}}) will tranform into

      either(in case age.enable_containment is off)
      properties.school = {addr:{city:'Toronto'}}

      or(in case age.enable_containment is on)
      properties @>> {school:{addr:{city:'Toronto'}}}

- Added @>> and <<@ operators. Unlike @> and <@, these operators does not recurse into sub-objects.

- Added regression tests.

### DIFF
--- a/sql/agtype_gin.sql
+++ b/sql/agtype_gin.sql
@@ -70,6 +70,8 @@ DEFAULT FOR TYPE agtype USING gin AS
   OPERATOR 9 ?(agtype, agtype),
   OPERATOR 10 ?|(agtype, agtype),
   OPERATOR 11 ?&(agtype, agtype),
+  OPERATOR 12 @>>(agtype, agtype),
+  OPERATOR 13 <<@(agtype, agtype),
   FUNCTION 1 ag_catalog.gin_compare_agtype(text,text),
   FUNCTION 2 ag_catalog.gin_extract_agtype(agtype, internal),
   FUNCTION 3 ag_catalog.gin_extract_agtype_query(agtype, internal, int2,
--- a/sql/agtype_operators.sql
+++ b/sql/agtype_operators.sql
@@ -53,3 +53,37 @@ CREATE OPERATOR <@ (
   RESTRICT = contsel,
   JOIN = contjoinsel
 );
+
+CREATE FUNCTION ag_catalog.agtype_contains_top_level(agtype, agtype)
+    RETURNS boolean
+    LANGUAGE c
+    IMMUTABLE
+RETURNS NULL ON NULL INPUT
+PARALLEL SAFE
+AS 'MODULE_PATHNAME';
+
+CREATE OPERATOR @>> (
+  LEFTARG = agtype,
+  RIGHTARG = agtype,
+  FUNCTION = ag_catalog.agtype_contains_top_level,
+  COMMUTATOR = '<<@',
+  RESTRICT = contsel,
+  JOIN = contjoinsel
+);
+
+CREATE FUNCTION ag_catalog.agtype_contained_by_top_level(agtype, agtype)
+    RETURNS boolean
+    LANGUAGE c
+    IMMUTABLE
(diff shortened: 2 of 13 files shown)


sql/agtype_operators.sql
```diff
@@ -53,3 +53,37 @@ CREATE OPERATOR <@ (
   RESTRICT = contsel,
   JOIN = contjoinsel
 );
+
```
jrgemignani, line 56: Please update this part once PR #1506 has been merged. These will need to be added to the upgrade file.

jrgemignani (changes requested):
Looks good less one minor thing that needs to be added. This is partially my fault as my PR got forgotten and I forgot to follow up on it.

rafsun42:
Is `properties @>> {school:{addr:{city:'Toronto'}}}` different from `properties = {school:{addr:{city:'Toronto'}}}`?

MuhammadTahaNaveed:
> Is properties @>> {school:{addr:{city:'Toronto'}}} different from properties = {school:{addr:{city:'Toronto'}}}?

Yes. So for example we consider a node has `properties = {name:'xyz', object:{key:{nestedkey:'nested value'}}, list:[1,2,3]}`

So if we query `MATCH (n ={list:[1,2,3]})`

- `properties @>> {list:[1,2,3]}` will check if the original properties contain a key `list`, and then if the value for that key is equal to the one in original properties.

- While `properties = {list:[1,2,3]}` would check that if the whole properties object i.e. `{name:'xyz', object:{key:{nestedkey:'nested value'}}, list:[1,2,3]}` is equal to `{list:[1,2,3]}`

I hope it explains.

sql/agtype_gin.sql
```diff
@@ -70,6 +70,8 @@ DEFAULT FOR TYPE agtype USING gin AS
   OPERATOR 9 ?(agtype, agtype),
   OPERATOR 10 ?|(agtype, agtype),
   OPERATOR 11 ?&(agtype, agtype),
+  OPERATOR 12 @>>(agtype, agtype),
+  OPERATOR 13 <<@(agtype, agtype),
```
MuhammadTahaNaveed, line 74: @jrgemignani any idea how I should be adding this change to the template file?

jrgemignani: The template file is basically the sql commands that are necessary to modify the previous (current) installed version to the next (new) version. So, if you are adding something, then just the sqls to add it. Basically, the sql for the age*.sql file. If you were modifying something existing, changing it, then the sql to modify the previous version to the new version. If you are removing something, then the sql to remove it.

Hopefully, that helps.

jrgemignani: Hi @MuhammadTahaNaveed Where you able to set it up?

MuhammadTahaNaveed: > Hi @MuhammadTahaNaveed Where you able to set it up?

I couldn't find any option to add an operator in operator class using ALTER command https://www.postgresql.org/docs/current/sql-alteropclass.html, so I dropped and recreated the operator class in that template file.

I have pushed these changes, please review at your convenience.

jrgemignani: You should just be able to add the commands that were added to the split age--x.x.x.sql file?

jrgemignani: @MuhammadTahaNaveed Oh, I see what you mean now

jrgemignani (approved):

rafsun42 (approved):

PR #1519, opened 2024-01-23
Update age_load to load scalar property values with appropriate type

rafsun42:
Previously, property values from csv files were always loaded as strings. This
patch converts a value to an appropriate scalar type (i.e. string, bool,
numeric, null) while loading. It uses the agtype_value_from_cstring()
function for conversion.

Additional change(s):
-------------------
 - Fix: for csv rows in edge files, create_agtype_from_list_i()'s start_index
   is corrected to 4

### DIFF
--- a/regress/expected/age_load.out
+++ b/regress/expected/age_load.out
@@ -233,3 +233,87 @@ NOTICE:  graph "agload_test_graph" has been dropped
  
 (1 row)
 
+--
+-- Test property type conversion
+--
+SELECT create_graph('agload_conversion');
+NOTICE:  graph "agload_conversion" has been created
+ create_graph 
+--------------
+ 
+(1 row)
+
+SELECT create_vlabel('agload_conversion','Person1');
+NOTICE:  VLabel "Person1" has been created
+ create_vlabel 
+---------------
+ 
+(1 row)
+
+SELECT load_labels_from_file('agload_conversion', 'Person1', 'age_load/conversion_vertices.csv');
+ load_labels_from_file 
+-----------------------
+ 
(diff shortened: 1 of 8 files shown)


jrgemignani (approved):

regress/expected/age_load.out
```diff
@@ -233,3 +233,87 @@ NOTICE:  graph "agload_test_graph" has been dropped
...
+(6 rows)
+
+SELECT create_vlabel('agload_conversion','Person2');
+NOTICE:  VLabel "Person2" has been created
+ create_vlabel 
+---------------
+ 
+(1 row)
+
+SELECT load_labels_from_file('agload_conversion', 'Person2', 'age_load/conversion_vertices.csv');
```
MuhammadTahaNaveed, line 277: Isn't this function name confusing for loading the vertices ? I know this function is not added in this PR but still, just pointing out.

MuhammadTahaNaveed (approved):

regress/expected/age_load.out
rafsun42: Right. Do you have a suggestion?

MuhammadTahaNaveed: Maybe load_vertices_from_file like we have load_edges_from_file for edges.....

rafsun42: Sounds good. 

Maybe it can be separate PR? This will also require updating the age-website repo.

MuhammadTahaNaveed: > Maybe it can be separate PR?

Yeah i agree 👍🏻.

PR #1525, opened 2024-01-29
Update age_load to make property value conversion optional

rafsun42:
This is an addition to a previous patch that introduced loading
property values from CSV files as agtype. This behavior is
made optional in this patch with a boolean function parameter
`load_as_agtype`. When this parameter is false, values are
loaded as string.


rafsun42:
@jrgemignani @muhammadshoaib This patch makes the `age_load` test finish in ~same time as before.

jrgemignani (approved):

PR #1527, opened 2024-01-31
Update the Go driver documentation and Linux installer

jrgemignani:
Updated the Go driver documentation and the Linux installer.

The driver installation will no longer install Java, Golang, or ANTLR for the user. This is left to the users system administrator. Instead, it will attempt to validate the installation of those components, then build and install the driver.

Also updated -
    modified:   drivers/golang/go.mod
    modified:   drivers/golang/go.sum

PR #1559, opened 2024-02-05
Update the Go driver documentation, Linux installer, and CI (#1527)

jrgemignani:
Updated the Go driver documentation, Linux installer, and Github CI file. The documentation needed to clarify what was needed. The installer needed to be fixed. The CI file needed to have the updated Go versions.

The driver installation for Linux will no longer install Java, Golang, or ANTLR for the user. This is left to the users system administrator. Instead, it will validate the current installation of those components, then build and install the driver.

Also updated -

### DIFF
--- a/.github/workflows/go-driver.yml
+++ b/.github/workflows/go-driver.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [ '1.18', '1.19' ]
+        go-version: [ '1.20', '1.21' ]
 
     defaults:
       run:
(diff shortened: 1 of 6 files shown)


.github/workflows/go-driver.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go-version: [ '1.18', '1.19' ]
+        go-version: [ '1.20', '1.21' ]
 
     defaults:
       run:
```
dehowef, line 18: The PG13 commit omits the command       run: go test . -v, while the rest mostly have them. Is this intentional? @jrgemignani Otherwise, looks good, but I will hold off on merging for now.

MuhammadTahaNaveed: @dehowef `run: go test . -v` is at the end of this file. Other PRs(on other branches) just removed an extra line at the end of this file, but this one does not have any.

PR #1561, opened 2024-02-05
Update the Go driver documentation, Linux installer, and CI (#1527)

jrgemignani:
Updated the Go driver documentation, Linux installer, and Github CI file. The documentation needed to clarify what was needed. The installer needed to be fixed. The CI file needed to have the updated Go versions.

The driver installation for Linux will no longer install Java, Golang, or ANTLR for the user. This is left to the users system administrator. Instead, it will validate the current installation of those components, then build and install the driver.

Also updated -

### DIFF
--- a/.github/workflows/go-driver.yml
+++ b/.github/workflows/go-driver.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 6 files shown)


MuhammadTahaNaveed (approved):

dehowef (approved):
Looks splendid~

PR #1564, opened 2024-02-06
Update age_load to load scalar property values with appropriate type (#1519) (#1525)

rafsun42:
- Update age_load to load scalar property values with appropriate type (#1519) (#1525)
- Borrow json_validate() from PG16

### DIFF
--- a/sql/age_main.sql
+++ b/sql/age_main.sql
@@ -122,17 +122,23 @@ CREATE FUNCTION ag_catalog.drop_label(graph_name name, label_name name,
     LANGUAGE c
     AS 'MODULE_PATHNAME';
 
+--
+-- If `load_as_agtype` is true, property values are loaded as agtype; otherwise
+-- loaded as string.
+--
 CREATE FUNCTION ag_catalog.load_labels_from_file(graph_name name,
                                                  label_name name,
                                                  file_path text,
-                                                 id_field_exists bool default true)
+                                                 id_field_exists bool default true,
+                                                 load_as_agtype bool default false)
     RETURNS void
     LANGUAGE c
     AS 'MODULE_PATHNAME';
 
 CREATE FUNCTION ag_catalog.load_edges_from_file(graph_name name,
                                                 label_name name,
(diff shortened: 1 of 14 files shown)


sql/age_main.sql
```diff
@@ -122,17 +122,23 @@ CREATE FUNCTION ag_catalog.drop_label(graph_name name, label_name name,
     LANGUAGE c
     AS 'MODULE_PATHNAME';
 
+--
+-- If `load_as_agtype` is true, property values are loaded as agtype; otherwise
+-- loaded as string.
+--
 CREATE FUNCTION ag_catalog.load_labels_from_file(graph_name name,
```
jrgemignani, line 129: This needs to be added to the `age--1.5.0--y.y.y.sql` file

sql/age_main.sql
```diff
@@ -122,17 +122,23 @@ CREATE FUNCTION ag_catalog.drop_label(graph_name name, label_name name,
...
                                                  label_name name,
                                                  file_path text,
-                                                 id_field_exists bool default true)
+                                                 id_field_exists bool default true,
+                                                 load_as_agtype bool default false)
     RETURNS void
     LANGUAGE c
     AS 'MODULE_PATHNAME';
 
 CREATE FUNCTION ag_catalog.load_edges_from_file(graph_name name,
```
jrgemignani, line 138: This needs to be added to the age--1.5.0--y.y.y.sql file

jrgemignani (changes requested):

sql/age_main.sql
rafsun42: Added to the upgrade script. I didn't have to use the `ALTER EXTENSION` commnad to remove dependency first. Because, Postgres does it automatically if DROP command is called from an upgrade script.

sql/age_main.sql
rafsun42: Same as above.

jrgemignani (approved):

Issue #1455, opened 2023-12-22
The Go driver becomes "undefined: parser", and the application cannot be built.


maehara-n:
**Describe the bug**
When trying to build an application that uses the Go driver, an error occurs. "undefined: parser" is output by the Go driver.

**What data setup do we need to do?**
**What is the necessary configuration info needed?**
Specify the following version of the Go driver in go.mod.
```text
github.com/apache/age/drivers/golang v0.0.0-20231212232227-0ea118689d78
```

**What is the command that caused the error?**
Build the application that is using the Go driver.
```
#0 5.670 # github.com/apache/age/drivers/golang/age
#0 5.670 /go/pkg/mod/github.com/apache/age/drivers/golang@v0.0.0-20231212232227-0ea118689d78/age/builder.go:42:22: undefined: parser.AgeParser
#0 5.670 /go/pkg/mod/github.com/apache/age/drivers/golang@v0.0.0-20231212232227-0ea118689d78/age/builder.go:43:21: undefined: parser.AgeVisitor
#0 5.670 /go/pkg/mod/github.com/apache/age/drivers/golang@v0.0.0-20231212232227-0ea118689d78/age/builder.go:114:9: undefined: parser.AgeVisitor
#0 5.670 /go/pkg/mod/github.com/apache/age/drivers/golang@v0.0.0-20231212232227-0ea118689d78/age/builder.go:129:52: undefined: parser.AgeoutContext
#0 5.670 /go/pkg/mod/github.com/apache/age/drivers/golang@v0.0.0-20231212232227-0ea118689d78/age/builder.go:135:52: undefined: parser.VertexContext
#0 5.670 /go/pkg/mod/github.com/apache/age/drivers/golang@v0.0.0-20231212232227-0ea118689d78/age/builder.go:151:50: undefined: parser.EdgeContext
#0 5.670 /go/pkg/mod/github.com/apache/age/drivers/golang@v0.0.0-20231212232227-0ea118689d78/age/builder.go:164:50: undefined: parser.PathContext
#0 5.670 /go/pkg/mod/github.com/apache/age/drivers/golang@v0.0.0-20231212232227-0ea118689d78/age/builder.go:186:51: undefined: parser.ValueContext
#0 5.670 /go/pkg/mod/github.com/apache/age/drivers/golang@v0.0.0-20231212232227-0ea118689d78/age/builder.go:201:56: undefined: parser.PropertiesContext
[...]

PR #1582, opened 2024-02-09
Update Go installation and add in parser files

jrgemignani:
Updated Go installation and added in built parser files.

While the parser files can be built by the user, there may be instances where the user doesn't have the ability to do so. For the latter case, they have been added.

    new file:   drivers/golang/parser/age_base_listener.go
    new file:   drivers/golang/parser/age_base_visitor.go
    new file:   drivers/golang/parser/age_lexer.go
    new file:   drivers/golang/parser/age_listener.go
    new file:   drivers/golang/parser/age_parser.go
    new file:   drivers/golang/parser/age_visitor.go


jrgemignani:
Please check out issue #1455 before any approvals to this PR and to give any comments.

rafsun42:
> Please check out issue #1455 before any approvals to this PR and to give any comments.

The change looks good. If it closes the issue, I will approve.

rafsun42 (approved):

PR #1507, opened 2024-01-12
Fix unsorted output of some queries in the cypher_match test

saygoodbyye:
Hello! In the process of discussing issue #1439, I found a way to quickly reproduce test failure (allocating 2 cores to the virtual machine). This also resulted in periodic failure of the "cypher_match" test. 
regression.out:
```
============== running regression test queries        ==============
test scan                         ... ok          269 ms
test graphid                      ... ok            8 ms
test agtype                       ... ok           67 ms
test catalog                      ... ok           66 ms
test cypher                       ... ok           21 ms
test expr                         ... ok          419 ms
test cypher_create                ... ok           88 ms
test cypher_match                 ... FAILED      518 ms
test cypher_unwind                ... ok           44 ms
test cypher_set                   ... ok           96 ms
test cypher_remove                ... ok           69 ms
test cypher_delete                ... ok          171 ms
test cypher_with                  ... ok           65 ms
test cypher_vle                   ... ok          811 ms
test cypher_union                 ... ok           21 ms
test cypher_call                  ... ok           37 ms
test cypher_merge                 ... ok          189 ms
test age_global_graph             ... ok          103 ms
test age_load                     ... ok         1243 ms
test index                        ... ok           92 ms
test analyze                      ... ok           19 ms
test graph_generation             ... ok           54 ms
test name_validation              ... ok          101 ms
test jsonb_operators              ... ok           64 ms
test drop                         ... ok          208 ms
============== shutting down postmaster               ==============

=======================
 1 of 25 tests failed. 
=======================
```
The essence of failures is that the execution of some queries may result in different output. 
regression.diffs:
```
diff -U3 /home/egor/work/age/regress/expected/cypher_match.out /home/egor/work/age/regress/results/cypher_match.out
--- /home/egor/work/age/regress/expected/cypher_match.out       2024-01-12 08:58:58.614061424 +0300
+++ /home/egor/work/age/regress/results/cypher_match.out        2024-01-12 09:16:11.639188103 +0300
@@ -2001,12 +2001,12 @@
        MATCH (a)-[r]-(b) RETURN r $$) as (r agtype);
                                                                                 r
 -----------------------------------------------------------------------------------------------------------------------------------------------------------------
+ {"id": 1407374883553283, "label": "e1", "end_id": 281474976710661, "start_id": 281474976710660, "properties": {}}::edge
+ {"id": 1407374883553283, "label": "e1", "end_id": 281474976710661, "start_id": 281474976710660, "properties": {}}::edge
[...]

PR #1600, opened 2024-02-14
Fix unsorted output of some queries in the cypher_match test (#1507)

jrgemignani:
Fix unsorted output of some queries in the cypher_match test


dehowef (approved):
Looks good.

PR #1615, opened 2024-02-15
Fix PG13 DockerHub build warnings

jrgemignani:
Fixed PG13 DockerHub build warnings by converting an Assert to an error message.

This PR is the PG14 PR as they both have the same warnings.

No regression tests needed adjusting.


dehowef:
Builds correctly, looks good

PR #1617, opened 2024-02-15
Fix PG11 DockerHub build warnings

jrgemignani:
Fixed PG11 DockerHub build warnings by converting an Assert to an error message.

This PR is the PG14 PR as they both have similar warnings.

No regression tests needed adjusting.

Resolved -

Conflicts:
	src/backend/parser/cypher_expr.c


dehowef (approved):
Excellent

PR #1623, opened 2024-02-17
Update Go installation and add in parser files (#1582)

jrgemignani:
Updated Go installation and added in built parser files.

While the parser files can be built by the user, there may be instances where the user doesn't have the ability to do so. For the latter case, they have been added.

    new file:   drivers/golang/parser/age_base_listener.go
    new file:   drivers/golang/parser/age_base_visitor.go
    new file:   drivers/golang/parser/age_lexer.go
    new file:   drivers/golang/parser/age_listener.go
    new file:   drivers/golang/parser/age_parser.go
    new file:   drivers/golang/parser/age_visitor.go

This addresses issue #1455


dehowef (approved):
Looks Good

PR #1636, opened 2024-03-04
Fix Issue 1630: MERGE using array not working in some cases

jrgemignani:
This PR fixes the issue where some previous clause user variables were not getting passed to the next clause.

Basically, there is a function, remove_unused_subquery_outputs, that tries to remove unnecessary variables by replacing them, in place, with a NULL Const. As these variables are needed, we need to wrap them with the volatile wrapper to stop that function from removing them.

Added regression tests.

### DIFF
--- a/regress/expected/cypher_merge.out
+++ b/regress/expected/cypher_merge.out
@@ -1263,8 +1263,137 @@ $$) as (a agtype);
 ---
 (0 rows)
 
+---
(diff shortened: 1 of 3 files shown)


MuhammadTahaNaveed (approved):
lgtm!

dehowef (approved):
Great work.

PR #1637, opened 2024-03-05
Fix Issue 1634: Setting all properties with map object causes error

jrgemignani:
Fixed the issue where setting the properties would not work, if it where an object passed as the properties.

Added regression tests.


MuhammadTahaNaveed (approved):
lgtm!

dehowef (approved):
Looks good.

PR #1665, opened 2024-03-13
Implement COUNT Subquery with PG specifications

dehowef:
Implements a version of COUNT subquery that evaluates the underlying subquery according to Postgres subquery specifications.

Modifies the grammar to accept all valid input for both COUNT and EXISTS subqueries.

Add regression tests related to COUNT

Modify the subquery regression tests to accomodate changes introduced in this update.


jrgemignani (approved):

dehowef:
Outdated

Issue #1439, opened 2023-12-06
Periodic regression test "expr" failures


saygoodbyye:
**Describe the bug**
Regression test "expr" fails in rare cases

**How are you accessing AGE (Command line, driver, etc.)?**
Accessing AGE through command line.

**What data setup do we need to do?**
Apache AGE (master branch) with PostgreSQL (REL_15_STABLE).

**What is the necessary configuration info needed?**
./configure CFLAGS=" -Og" --enable-tap-tests --enable-debug --enable-cassert --prefix=/tmp/pgsql

**What is the command that caused the error?**
I removed all tests except "expr" from the Makefile and put "make installcheck" in a loop to reproduce the failure.
```
diff --git a/Makefile b/Makefile
index ab98665..2f08858 100644
--- a/Makefile
+++ b/Makefile
@@ -76,30 +76,8 @@ EXTENSION = age
 DATA = age--1.4.0.sql
 
 # sorted in dependency order
-REGRESS = scan \
-          graphid \
-          agtype \
-          catalog \
-          cypher \
-          expr \
-          cypher_create \
-          cypher_match \
-          cypher_unwind \
-          cypher_set \
-          cypher_remove \
-          cypher_delete \
-          cypher_with \
-          cypher_vle \
-          cypher_union \
-          cypher_call \
-          cypher_merge \
-          age_global_graph \
-          age_load \
-          index \
-          analyze \
-          graph_generation \
-          name_validation \
-          drop
+
+REGRESS = expr
 
 srcdir=`pwd`
```
Start tests:
```
for i in `seq 1000000`;do echo "iteration $i";NO_TEMP_INSTALL=1 make -s installcheck || break;done
```
The test may fail at a random iteration. I received failures at iteration ~600, at iteration ~3500, and during the last run the test failed at iteration 199 with following "regression.diffs":
```
iteration 199
+++ regress install-check in  +++
============== creating temporary instance            ==============
============== initializing database system           ==============
============== starting postmaster                    ==============
running on port 61958 with PID 3302580
============== creating database "contrib_regression" ==============
CREATE DATABASE
ALTER DATABASE
ALTER DATABASE
ALTER DATABASE
ALTER DATABASE
ALTER DATABASE
ALTER DATABASE
============== installing age                         ==============
CREATE EXTENSION
============== running regression test queries        ==============
test expr                         ... FAILED      643 ms
============== shutting down postmaster               ==============

======================
 1 of 1 tests failed. 
======================

```
regression.diffs:
[...]

PR #1677, opened 2024-03-15
Fix bogus failure of test expr

saygoodbyye:
All discussions were conducted in #1439 
Changes:
1) debug_parallel_query was set to 'regress' to avoid "CONTEXT: parallel worker" message appear 
2) fixed unsorted output in the specified query


jrgemignani:
@saygoodbyye No further merging of this or related PRs can occur until the issue of regression runtimes is addressed in #1678

PR #1711, opened 2024-03-22
Add PGXN release workflow

theory:
It's triggered by a semver tag. I don't see any tags on this repo, but do notice that the [version on PGXN](https://pgxn.org/dist/apacheage/) appears to be a couple versions back. Use this workflow if you'd like to automate its release; change the `on` triggers to trigger a release by some other means.

Be sure to add workflow secrets to this project named `PGXN_USERNAME` and `PGXN_PASSWORD` to allow it to authenticate to publish the extension.

Other stuff:

* Add `.gitattributes` so that the contents of `.github` and `.git` are excluded from the exported archive
* Fix the extension name in the `META.json` file to the actual extension name (same as the control file name)
* Updated the version in the `META.json` and shortened the abstract and description a bit (otherwise they take up a lot of space on the [PGXN page](https://pgxn.org/dist/apacheage/).
* Read the versions from the `META.json` file and set the extension name from `MODULE_big`.
* Add a `dist` make target to produce a zip file to manually upload to PGXN.

### DIFF
--- a/META.json
+++ b/META.json
@@ -1,24 +1,24 @@
 {
    "name": "ApacheAGE",
-   "abstract": "Apache AGE is a PostgreSQL Extension that provides graph database functionality",
-   "description": "Apache AGE is a PostgreSQL Extension that provides graph database functionality. AGE is an acronym for A Graph Extension, and is inspired by Bitnine's fork of PostgreSQL 10, AgensGraph, which is a multi-model database. The goal of the project is to create single storage that can handle both relational and graph model data so that users can use standard ANSI SQL along with openCypher, the Graph query language. A graph consists of a set of vertices (also called nodes) and edges, where each individual vertex and edge possesses a map of properties. A vertex is the basic object of a graph, that can exist independently of everything else in the graph. An edge creates a directed connection between two vertices. A graph database is simply composed of vertices and edges. This type of database is useful when the meaning is in the relationships between the data. Relational databases can easily handle direct relationships, but indirect relationships are more difficult to deal with in relational databases. A graph database stores relationship information as a first-class entity. Apache AGE gives you the best of both worlds, simultaneously.",
-   "version": "1.3.0",
+   "abstract": "Graph database functionality for PostgreSQL",
+   "description": "Apache AGE is a PostgreSQL Extension that provides graph database functionality. The goal of the project is to create single storage that can handle both relational and graph model data so that users can use standard ANSI SQL along with openCypher, the Graph query language.",
+   "version": "1.5.0",
    "maintainer": [
       "users@age.apache.org"
    ],
    "license": "apache_2_0",
    "provides": {
-      "ApacheAGE": {
+      "age": {
          "abstract": "Apache AGE is a PostgreSQL Extension that provides graph database functionality",
-         "file": "age--1.3.0.sql",
+         "file": "age--1.5.0.sql",
          "docfile": "README.md",
-         "version": "1.3.0"
+         "version": "1.5.0"
       }
    },
    "prereqs": {
       "runtime": {
          "requires": {
-            "PostgreSQL": "14.0.0"
+            "PostgreSQL": "11.0.0"
          }
       }
    },
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+.gitignore export-ignore
+.gitattributes export-ignore
+.github export-ignore
+dockerignore export-ignore
--- a/.github/workflows/pgxn-release.yml
+++ b/.github/workflows/pgxn-release.yml
@@ -1,0 +1,20 @@
+name: 🚀 Release on PGXN
+on:
+  push:
+    # Release on semantic version tag.
+    tags: ['v[0-9]+.[0-9]+.[0-9]+']
+jobs:
+  release:
+    name: Release on PGXN
+    runs-on: ubuntu-latest
+    container: pgxn/pgxn-tools
+    steps:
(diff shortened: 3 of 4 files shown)


aked21:
Hi @rafsun42 ... Can you look into this PR?

rafsun42:
@aked21 Sure, I am reviewing it.

aked21:
Hi @rafsun42, can this get merged?

rafsun42:
@theory Hello. Is this PR tested?

jrgemignani:
I am a bit weary of modifications to the `Makefile`. Why are those modifications necessary?

theory:
Yes, I tested it with the exception of actually releasing it. See [this build](/theory/age/actions/runs/8395514381/job/22994946387).

The changes to the Makefile are no strictly speaking required; I thought it useful to demonstrate how to use it to set the version in one place. pgrx projects often read the version from a `Cargo.toml` file, but many PGXS-based projects use the `META.json` file. The `dist` target is not required, but handy to manually check that that the PGXN zip file is generated properly.

We could probably remove it all if you'd like; the [`pgxn-bundle`](https://github.com/pgxn/docker-pgxn-tools/blob/main/bin/pgxn-bundle) command used in the release workflow does pretty much the same thing, depending on the `META.json` file to read the version.

jrgemignani:
@theory My preference, for now, is to not touch the `Makefile`. I do get what you're trying to do, though, and it is appreciated.

theory:
Okay, removed the `Makefile` changes in 969842b.

META.json
```diff
@@ -1,24 +1,24 @@
...
          "docfile": "README.md",
-         "version": "1.3.0"
+         "version": "1.5.0"
       }
    },
    "prereqs": {
       "runtime": {
          "requires": {
-            "PostgreSQL": "14.0.0"
+            "PostgreSQL": "11.0.0"
```
jrgemignani, line 21: Is this the minimum PG? If so, while we do support 11, we've stopped most new additions to 11 & 12. 13 would be better.

META.json
```diff
@@ -1,24 +1,24 @@
...
+   "abstract": "Graph database functionality for PostgreSQL",
+   "description": "Apache AGE is a PostgreSQL Extension that provides graph database functionality. The goal of the project is to create single storage that can handle both relational and graph model data so that users can use standard ANSI SQL along with openCypher, the Graph query language.",
+   "version": "1.5.0",
    "maintainer": [
       "users@age.apache.org"
    ],
    "license": "apache_2_0",
    "provides": {
-      "ApacheAGE": {
+      "age": {
```
jrgemignani, line 11: Why **age** as opposed to Apache AGE?

jrgemignani (changes requested):
Some questions

META.json
jrgemignani: Additionally, we support up to PG16.

theory: Yes, this is the minimum supported version. I took it from the README, which says:

> for now AGE supports Postgres 11, 12, 13, 14, 15 & 16.

Do you want to drop support for 11 and 12?

META.json
theory: Because that's the name of the extension, as indicated by the file name `age.control`.

META.json
jrgemignani: Would this cause it to build against 11? The main issue with 11 & 12 is that we have mostly dropped support of them beyond 1.5.0.

jrgemignani: Btw, a lot of this has to do with 11 and 12 being much different from 13 and above.

theory: It just means that it supports PostgreSQL 11 and up. If someone downloaded it and looked, they'd know they could build it against Postgres 11. At some point a smoke test setup might depend on this config to decide what versions of Postgres to test builds against.

jrgemignani: @rafsun42 @MuhammadTahaNaveed @dehowef @Zainab-Saad 
thoughts?

rafsun42: @theory An age release for PG16 (which is master branch currently) will not compile on Postgres 11, or any other Postgres version (at least it is not meant to). For context, we have different development branches for different PG versions. Given that, I believe the version in the meta file would depend on the branch this PR is being applied on. For example, if the PR is being applied on PG15, then the version can point to the latest minor version of PG15.

jrgemignani:
@theory 

> Be sure to add workflow secrets to this project named PGXN_USERNAME and PGXN_PASSWORD to allow it to authenticate to publish the extension.

I'm not sure what or where to add this information.

theory:
> I'm not sure what or where to add this information.

Presumably you have access to the PGXN username and password, since Apache Age is [on PGXN already](https://pgxn.org/dist/apacheage/). What you need to do is go to https://github.com/apache/age/settings/secrets/actions and add two repository secrets, one named `PGXN_USERNAME` and the other `PGXN_PASSWORD`. These will then be injected into the GitHub workflow and used to authenticate to PGXN and publish the release.

META.json
theory: Also, be aware that these will be considered separate distributions by PGXN. Wheres today you have `https://pgxn.org/dist/apacheage/`, future releases will be under `https://pgxn.org/dist/apacheage16/`, `https://pgxn.org/dist/apacheage15/`, etc.

jrgemignani:
@theory Unfortunately, I don't have access to that. We will need to find out who does before moving forward and merging this.

jrgemignani:
@theory We're still waiting on access. @eyab can we get a status?

PR #1717, opened 2024-04-01
More clarification is needed for the purpose of AGE.

elmeilgyahmed:
More clarification is needed on why utilizing graph analytics (AGE) is preferable to relational databases when dealing with interconnected data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@
 
 </br>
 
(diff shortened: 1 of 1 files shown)


M-Bostone:
actually it all depends on the specific requirements needed by your usecase, though graph analytics has an added advantage when it comes to interconnected data but the relational database is more suited when it comes to well defined data schemas.

jrgemignani:
@eyab @aked21 Is this PR needed? It seems superfluous.

dehowef:
I think the grammar changes are nice, but the title and desc of the PR can be updated to reflect the changes made.

PR #1729, opened 2024-04-05
Fix Issue 1691 - MERGE incorrectly creates multiple vertices (#1718)

jrgemignani:
Fixed issue 1691 where MERGE would incorrectly create multiple vertices. This only occurred when MERGE was being driven by a previous clause.

NOTE: To be more correct, the issue is with creating duplicate
      paths.

The reason this happened was due to the visibility of tuples that were created during the MERGE instance. It is not possible to add them in to be rescanned.

Because of this limitation, it required adding the ability to MERGE to know what path an instance had already created.

Added regression tests.

Modified the following files to add missing functionality in PG13 and below branches. Specifically adding the function datum_image_hash.

    modified:   src/backend/executor/cypher_utils.c
    modified:   src/include/executor/cypher_utils.h

Resolved Conflicts: (due to differences between PG13 & PG12)
	src/backend/executor/cypher_merge.c

Modified datum_image_hash to use hash_any.

Changed datum_image_eq to datumIsEqual as the former doesn't exist in PG11.

### DIFF
--- a/regress/expected/cypher_merge.out
+++ b/regress/expected/cypher_merge.out
@@ -1263,8 +1263,8 @@ $$) as (a agtype);
 ---
 (0 rows)
 
----
---- Issue 1630 MERGE using array not working in some cases
+--
+-- Issue 1630 - MERGE using array not working in some cases
 --
 SELECT * FROM create_graph('issue_1630');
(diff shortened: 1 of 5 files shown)


MuhammadTahaNaveed (approved):
looks good!

PR #1742, opened 2024-04-11
Add helpful messages to the VLE subsystem

jrgemignani:
Added helpful messaging to the VLE subsystem.

These changes are due to issue 365 where there was an issue with the VLE that turned out to not be with the VLE. Basically, there is a way, apparently, to have a malformed edge or possibly one not fully removed. When the VLE subsystem does its load on the graph, it would error out on these edges as this was considered to be a potential data corruption issue.

The changes added are to report and ignore these edges. Reporting them, so as to allow the user to take corrective action by removing them. Ignoring them, as they are likely not relevant to the query. Basically, leaving it up to the user to decide.

No impact to current regression tests.

### DIFF
--- a/src/backend/utils/adt/age_global_graph.c
+++ b/src/backend/utils/adt/age_global_graph.c
@@ -101,7 +101,7 @@ static bool insert_edge(GRAPH_global_context *ggctx, graphid edge_id,
                         graphid end_vertex_id, Oid edge_label_table_oid);
 static bool insert_vertex_edge(GRAPH_global_context *ggctx,
                                graphid start_vertex_id, graphid end_vertex_id,
-                               graphid edge_id);
+                               graphid edge_id, char *edge_label_name);
 static bool insert_vertex_entry(GRAPH_global_context *ggctx, graphid vertex_id,
                                 Oid vertex_label_table_oid,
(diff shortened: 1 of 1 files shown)


src/backend/utils/adt/age_global_graph.c
```diff
@@ -568,11 +610,13 @@ static void load_edge_hashtable(GRAPH_global_context *ggctx)
 
             /* insert the edge into the start and end vertices edge lists */
             inserted = insert_vertex_edge(ggctx, edge_vertex_start_id,
-                                          edge_vertex_end_id, edge_id);
-            /* this insert must not fail */
+                                          edge_vertex_end_id, edge_id,
+                                          edge_label_name);
             if (!inserted)
```
rafsun42, line 615: Overall looks good. Just a quick questions. Does `!insert` always imply a `WARNING` instead of an `ERROR`?

jrgemignani: The only time `insert_vertex_edge` will fail to insert is if there is an issue with the edge's start and end vertices. This means a malformed or dangling edge.

rafsun42 (approved):

PR #1747, opened 2024-04-12
Add helpful messages to the VLE subsystem (#1742)

jrgemignani:
Added helpful messaging to the VLE subsystem.

These changes are due to issue 365 where there was an issue with the VLE that turned out to not be with the VLE. Basically, there is a way, apparently, to have a malformed edge or possibly one not fully removed. When the VLE subsystem does its load on the graph, it would error out on these edges as this was considered to be a potential data corruption issue.

The changes added are to report and ignore these edges. Reporting them, so as to allow the user to take corrective action by removing them. Ignoring them, as they are likely not relevant to the query. Basically, leaving it up to the user to decide.

No impact to current regression tests.


mannas006:
It appears that significant modifications have been made to the VLE subsystem to address possible problems with incomplete or distorted edges in the graph that could lead to data corruption. It's a good idea to notify users of these problems and provide them with the choice to make corrections. Maintaining system performance and stability can also be aided by ignoring unnecessary edges. That these modifications haven't had an impact on your ongoing regression testing is wonderful to hear. Please let me know if you require any additional help or if there are any updates!

PR #1749, opened 2024-04-15
[PROPOSAL] I'd like to propose an adding a graph database to the project

seo-kw:
Hello, I'd like to propose adding a graph database to this project.

- It's more like AGE project would be integrated with postgresql13 itself as native AGE support. as a result, it’s a kind of DBMS version of AGE.
- Adding the first version of pg_graph to the branch with an apache license.
- Also it has a commit history like the "ageviewer_go" branch. it is just above one commit of AGE's first one.
- I'm not sure about whether the pg_graph name itself would be a problem or not, so, please assume that this is a kind of temporal name.
- Also this PR is a kind of proposal, so I just made this PR as a draft and allowed edits by maintainers.

- Initially, I'd like to make a new branch for this one, but PR is not a kind of that one. So I have no choice. please do not think about these conflicts are actual conflicts. **_It should not merge into the master branch._**

Any suggestion, or something, I would be thankful.

Thanks, and regard.


jrgemignani:
@seo-kw Thank you for this. We will need to create a separate branch that you can make this PR for.

PR #1752, opened 2024-04-16
Add the command graph_stats and improve VLE messaging for load (#1750)

jrgemignani:
Added the command graph_stats. This command will force a reload of the specified graph. This is done to generate current statistics and dump any warning messages about the graph generated during a the load.

Cleaned up some messaging from the VLE load process for duplicate vertices and edges.

Added regression tests.


dehowef:
Looks good to me.

dehowef (approved):

PR #1678, opened 2024-03-15
Fix bogus failure of test expr

saygoodbyye:
All discussions were conducted in #1439 
Changes:

1) force_parallel_mode was set to 'regress' to avoid "CONTEXT: parallel worker" message appear
2) fixed unsorted output in the specified query

PR #1772, opened 2024-04-19
Fix slow running regression test expr

rafsun42:
Fixes slow runtime, introduced in PR #1678, of the regression test expr. This patch sets the config param `force_parallel_mode TO regress` only to a selective query rather than the entire expr.sql file.

### DIFF
--- a/regress/expected/expr.out
+++ b/regress/expected/expr.out
@@ -19,7 +19,6 @@
 SET extra_float_digits = 0;
 LOAD 'age';
 SET search_path TO ag_catalog;
-SET force_parallel_mode TO regress;
 SELECT * FROM create_graph('expr');
 NOTICE:  graph "expr" has been created
  create_graph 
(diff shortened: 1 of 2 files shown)


rafsun42:
@saygoodbyye This PR fixes the slow runtime of expr that John mentioned in #1678. Could you verify if it still solves your original issue?

jrgemignani:
@rafsun42 I can confirm that this PR addresses the slow `expr` regression test.

From -

`test expr                         ... ok         5277 ms`

To -

`test expr                         ... ok         1022 ms`

saygoodbyye:
@rafsun42, Thank you! There really was a mistake on my part. I didn’t pay attention to the execution time of the tests. This pull request fixes the problem

dehowef (approved):

dehowef:
Great work! I'll go ahead and merge this. @saygoodbyye Thanks for your contribution, this stuff happens~

jrgemignani:
@rafsun42 Thank you for fixing this!

PR #1782, opened 2024-04-23
Update META.json

AneeqaCodes:
I removed the duplicate abstract field from the provides.ApacheAGE section, as it's already present at the root level.

### DIFF
--- a/META.json
+++ b/META.json
@@ -9,7 +9,6 @@
    "license": "apache_2_0",
    "provides": {
       "ApacheAGE": {
-         "abstract": "Apache AGE is a PostgreSQL Extension that provides graph database functionality",
          "file": "age--1.3.0.sql",
          "docfile": "README.md",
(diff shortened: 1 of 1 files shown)


jrgemignani:
@AneeqaCodes Not sure if this conflicts with #1711, which is pending Apache.
@eyab

AneeqaCodes:
I don't think so that it will create any conflict.

AneeqaCodes:
Redundancy should be eliminated.

jrgemignani:
@AneeqaCodes As this is just fixing a redundancy, which isn't causing any issues, I would like to wait for the other PR to be processed first. I don't want to generate any potential merge conflicts.

AneeqaCodes:
@jrgemignani I think you're right and I will try to find out more crucial issues which need to fixed asap.

PR #1793, opened 2024-04-25
python driver psycopg3

mhmaguire:
Updated python driver code to psycopg 3 for my own purposes and figured you all might like to do the same sometime.


aked21:
Hi @mhmaguire, can you specify what have been changed and/or added for your needs?  
Thank you for contributing!
AK

dehowef:
Wow! Thank you so much this is great.

dehowef:
@mhmaguire I pulled down your changes to take a look, and the test files are complaining with the line:

AttributeError: 'TestAgeBasic' object has no attribute 'args'

Granted I just took a cursory glance, but I wanted to ask, do the tests work on your machine? Thanks for your time.

dehowef:
@mhmaguire scratch my previous comment-- User error on my part! I'll continue reviewing this when I find a moment

dehowef:
> @mhmaguire scratch my previous comment-- User error on my part! I'll continue reviewing this when I find a moment

Lol okay actually sorry to walk back my comment again, but when I run the unittests via "python -m unittest -v test_age_.py" and python -m unittest -v test_networkx.py as stated in the documentation, they throw an attribute error as follows:

Traceback (most recent call last):
  File "/home/user/age/drivers/python/test_age_py.py", line 36, in setUp
    host=self.args.host,
         ^^^^^^^^^
AttributeError: 'TestAgeBasic' object has no attribute 'args'

The unit tests still run fine if I call them directly, but I think it'd be a relatively easy fix to get this error to disappear. Would you mind amending your commit so that they work with these unit test commands? Otherwise it looks great and I appreciate what you've done here. I think after that change, I'll give it another once over and probably good to go. @mhmaguire

mhmaguire:
@dehowef I set a default prefilled namespace so tests won't fail when bypassing argparse.

`python test_age_py.py` would have a similar effect...

dehowef:
> @dehowef I set a default prefilled namespace so tests won't fail when bypassing argparse.
> 
> `python test_age_py.py` would have a similar effect...

@mhmaguire Yeah, that way of testing was written by the original author of the python driver, been thinking of updating/removing that section of the documentation. I just looked over everything, and it looks great! We usually want to bring all updates to the main branch down to previous versions, from PG16 down til PG13. If you want to create the PRs for those branches as well, feel free to, otherwise I might just go ahead and do that within the next few days!! Thank you so much for your contribution :slightly_smiling_face:

PR #1799, opened 2024-04-26
Fix issue 1767: CREATE TABLE AS SELECT * FROM cypher, errors

jrgemignani:
Fixed issue 1767: create table as select * from cypher()... throws an error -

ERROR:  unhandled cypher(cstring) function call

CREATE TABLE AS is a utility command and therefore is not seen by the query tree walker. This is by design, per the query_tree_walker code comments. This is why the cypher command wasn't transformed causing the above error.

The fix was to add a check for utility commands in the Query section of the convert_cypher_walker. If found, the embedded query from CreateTableAsStmts nodes is extracted and processed.

NOTE: This only enables the utility command CreateTableAsStmts nodes.
      Others will need to be added on a case by case basis.

Added regression tests.


dehowef (approved):
Short, simple, sweet. Looks good.

PR #1801, opened 2024-04-26
[PROPOSAL] Integrated age extension into the postgresql database

seo-kw:
Proposed pg_graph project will provide AGE extension features in DBMS core engine which is based on PG13 for now.

Thanks.


seo-kw:
@jrgemignani  Sorry for the late reply, and here is the answer about your three questions.

- Any origin story? Where is it from? Who created it?

    It was created been ago for graph processing in RDBMS. you can still visit the original one.
    It comes from Bitnine's Agensgraph. also, created and copyright reserved by Bitnine.

- Is it being donated? If so, could you briefly tell us why?

    Yes. It will be donated.
    AGE is an extension, and originally it was derived from Agensgraph's extension project.
    so developing overlapped or something couldn't be provided with the extension itself, this could be a one of a choice.
    and also sharing the vision of graph processing.

- Are there any previous licenses or owners which could interfere with any donation?

    It has been integrated with Postgresql so, there's a postgres license. you can check in most of the codes.
    and there was bitnine licensed for codes. but it's removed because it'll be donated to the AGE project without copyright issues.


Sorry for the delayed reply, and thanks again.

MironAtHome:
Is it compatible with Age? I mean if I develop my graph database using age, would a build of this Postgres integrated graph-db be able to work with same data?

jrgemignani:
@MironAtHome Unless it stores data the same as AGE, then the answer is likely no.

PR #1803, opened 2024-04-26
Implement Returnless Unions in Subqueries

dehowef:
Added logic to support returnless unions in subqueries.

Added regression tests to support this addition. Modified regression tests to capture more cases.


regress/sql/cypher_subquery.sql
```diff
@@ -73,13 +73,24 @@ SELECT * FROM cypher('subquery', $$ MATCH (a:person)
 												 }
 									RETURN (a) $$) AS (result agtype);
 
---union, no returns, not yet implemented, should error out
+--union, no returns
 SELECT * FROM cypher('subquery', $$ MATCH (a:person)
 									WHERE EXISTS {
```
rafsun42, line 78: Just curious (because I don't know EXISTS subquery much), how does WHERE EXISTS work if the inner subquery does not return?

dehowef: In this case you commented on, no returns means that there is no RETURN clause specified. The query will still have a result, even if there is no return clause.

I am not sure what you mean by does not return, as all queries should return a result.

rafsun42 (approved):

PR #1808, opened 2024-04-26
Minor VLE and agtype_eq/ne performance updates

jrgemignani:
Integrated datum_image_hash and datum_image_eq to the logic for the VLE edge property datum match.

Additionally, both were added to agtype_eq and agtype_ne.

The hope is that, for large datums, these routines will improve performance.

No impact to regression tests.


dehowef:
Great!

PR #1810, opened 2024-04-27
Update .dockerignore

AneeqaCodes:
Added a slash (/) after .idea and .deps to indicate that they are directories.


jrgemignani:
@MuhammadTahaNaveed Are these safe additions for `.dockerignore`?

PR #1814, opened 2024-04-29
Update .gitignore

AneeqaCodes:
Added a slash (/) after .idea and .deps to indicate that they are directories.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,12 @@
+@@ -1,11 +1,9 @@
 *.o
 *.so
+.gitignore
 build.sh
 .idea
 .deps
+.idea/
+.deps/
 .DS_Store
 *.tokens
 *.interp


.gitignore
```diff
@@ -1,8 +1,12 @@
+@@ -1,11 +1,9 @@
 *.o
 *.so
+.gitignore
```
rafsun42, line 4: I don't think we should put `.gitignore` there.

AneeqaCodes: I know it seems like redundant but I added .gitignore to maintain consistency and prevent accidental commits of this file, ensuring clarity in our version control process. And actually it also serves as a reminder for all contributors that we should ignore this file itself to prevent accidental commits.

emotionbug: @rafsun42 +1

AneeqaCodes: I don't know what's wrong with my changes, it gives clarity and better understanding of the file.

jrgemignani:
Closed as unnecessary.

PR #1740, opened 2024-04-08
Remove duplicate check for label validity

jberzy:
While looking through the code, I noticed a redundant check to see if the supplied label name was valid. This PR removed that redundant check.

PR #1848, opened 2024-05-08
Remove duplicate check (#1740) (#1833)

MironAtHome:
This commit removes [a benign] redundant check for label validity

Based on earlier checkins by jrgemignani and jberzy <john.berzy@gmail.com>


jrgemignani:
Release branches are not modified once released.

Issue #1850, opened 2024-05-08
Server crashes when executing SELECT * FROM cypher('list_comprehension' ...)


saygoodbyye:
Describe the bug
Server crashes when executing the following SQL-script

How are you accessing AGE (Command line, driver, etc.)?
Accessing AGE through command line.

What data setup do we need to do?
Apache AGE (PG16 branch) with PostgreSQL (REL_16_STABLE).

What is the necessary configuration info needed?
My PostgreSQL configuration:
```
./configure CFLAGS=" -Og" --enable-tap-tests --enable-debug --enable-cassert
```
What is the command that caused the error?
```sql
CREATE EXTENSION age;
SET search_path TO ag_catalog;
SELECT create_graph('list_comprehension');
SELECT * FROM cypher('list_comprehension', $$ WITH [1, 2, 3] AS u UNWIND collect(u) AS v RETURN v $$) AS (u agtype);
SELECT * FROM cypher('list_comprehension', $$ MATCH(u {list: [0, 2, 4, 6, 8, 10, 12]}) WITh u, collect(u.list) AS v SET u += {b: [u IN range(0, 5)]} SET u.c = [u IN v[0]] RETURN u $$) AS (u agtype);
```
backtrace:
```c
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=140110869255104) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=140110869255104) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=140110869255104, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007f6e1c824476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007f6e1c80a7f3 in __GI_abort () at ./stdlib/abort.c:79
#5  0x000056309a7a4c24 in ExceptionalCondition (conditionName=conditionName@entry=0x56309a8fb478 "attnum >= 0 && attnum < outerslot->tts_nvalid", fileName=fileName@entry=0x56309a8fab27 "execExprInterp.c", lineNumber=lineNumber@entry=582) at assert.c:66
#6  0x000056309a46d4c1 in ExecInterpExpr (state=0x56309c452ce0, econtext=0x56309c451d70, isnull=0x7fffc5c16027) at execExprInterp.c:582
[...]

PR #1873, opened 2024-05-13
Update cypher_set.sql (Addressing issue #1850).

AneeqaCodes:
Try to Resolve server crashes related to list comprehension in Cypher queries, addressing issue #1850.


AneeqaCodes:
And issue # 1850 occurred due to query because aggregations are not properly done . Revised query 
SELECT * FROM cypher('list_comprehension', $$ MATCH(u {list: [0, 2, 4, 6, 8, 10, 12]})
    WITH u, collect(u.list) AS v 
    SET u += {b: [u IN range(0, 5)]} 
    SET u.c = [u IN v[0]] 
    RETURN u 
$$) AS (u agtype);

emotionbug:
- wrong target branch.
- this isn't fix the problem. your fix ignores problem.

PR #1885, opened 2024-05-17
Fix issue 1884 - Ambiguous column reference error in SET clause

Zainab-Saad:
The error occured due to SET transform not checking if the target entry for the RHS item is already present in the query target list and added the item to tl which resulted in variable name conflict.

Instead transformed the item and added to tl only if target entry for the var name could not be found

Added regression tests


src/backend/parser/cypher_clause.c
```diff
@@ -1836,27 +1837,59 @@ cypher_update_information *transform_cypher_set_item_list(
...
-
-        if (has_a_cypher_list_comprehension_node(set_item->expr))
+        /*
+         * if the RHS is a ColumnRef and the TargetEntry for the variable
+         * already exists, no need to transform the RHS item,
+         * just add volatile wrapper to the target entry of the existing item
+         */
+        if (IsA(set_item->expr, ColumnRef))
         {
-            query->hasAggs = true;
```
jrgemignani, line 1847: Was this intentionally removed?

src/backend/parser/cypher_clause.c
```diff
@@ -1836,27 +1837,59 @@ cypher_update_information *transform_cypher_set_item_list(
...
+                // RHS variable already exists
+                rhs_var_exists = true;
+                item->prop_position = prop_resno;
+                target_item = findTarget(query->targetList, prop_name);
+                add_volatile_wrapper_to_target_entry(query->targetList,
+                                                     item->prop_position);
+            }
         }
 
-        if (!query->hasAggs && nodeTag(target_item->expr) == T_Aggref)
```
jrgemignani, line 1850: Same question

jrgemignani:
Resolved with PR #2306

PR #1892, opened 2024-05-23
Order some regression tests for stability on big-endian

df7cb:
On Debian's s390x architecture, some regression tests were failing because the result set was reordered. Fix by attaching ORDER BY in problematic cases.


df7cb:
Fwiw, this change is the same as submitted in #1503, and it is still good to go. In the earlier PR I had made the mistake of wording like it was WIP - it does leave some problems open, but the first half is ready to be committed.

jrgemignani (approved):

jrgemignani:
@df7cb Could you please create PRs for PG16, PG15, PG14, & PG13 so that we can keep all the branches in sync?

PR #1924, opened 2024-06-19
Fix issue 1910: Server crashes when using exists(vle path)

jrgemignani:
Fixed issue 1910: Server crashes when using exists(path) with explicit length paths.

The crash was caused because transform_ColumnRef did not pass down the levels_up value into scanNSItemForColumn.

Upon fixing that issue, it was found that the transform_cypher_node function did not correctly process the vle_function_end_var end variable. In some cases, in the WHERE clause, it needs to be created.

Added regression tests.


eyab (approved):
Looks good to me.

Issue #839, opened 2023-04-24
Allowing non-superuser to access Apache AGE through python driver


CC-Hsu:
Hi, Team,

Previously we try to use Apache AGE Python driver to access an AGE database.

But we find that we can not run the driver successfully without a DB user with superuser privilege due to `LOAD age;` command.

We know that in the prerequisite information one need to test the setup with `LOAD age;` command in the `psql`.
https://github.com/apache/age/tree/master/drivers/python#check-age-loaded-on-your-postgresql

And we also see the driver also execute `LOAD age;` first (I copy the corresponding code snippet below).
https://github.com/apache/age/blob/master/drivers/python/age/age.py#L32
```python
def setUpAge(conn:ext.connection, graphName:str):
    with conn.cursor() as cursor:
        cursor.execute("LOAD 'age';")
```

Although the [`LOAD` command](https://www.postgresql.org/docs/current/sql-load.html) is superuser-restricted, it may be replaced by setting [`local_preload_libraries`](https://postgresqlco.nf/doc/en/param/local_preload_libraries/) GUC, so that `LOAD` command can be skipped and allow non-superuser DB roles to use Apache AGE objects.

Is it possible to improve the python driver to also support non-superuser usage in this way?

Best Regards.

aru-d-at:
Hi Hsu,

Thank you for pointing out this limitation. It seems like non-superuser connections are not possible with the current drivers. This may cause an issue in deploying production-ready applications in the future.

For now, I will look into your suggestion first.

Regards

jrgemignani:
@CC-Hsu 

> Although the [LOAD command](https://www.postgresql.org/docs/current/sql-load.html) is superuser-restricted, it may be replaced by setting [local_preload_libraries](https://postgresqlco.nf/doc/en/param/local_preload_libraries/) GUC, so that LOAD command can be skipped and allow non-superuser DB roles to use Apache AGE objects.

From the link for load - 

> Non-superusers can only apply LOAD to library files located in $libdir/plugins/ — the specified filename must begin with exactly that string. (It is the database administrator's responsibility to ensure that only “safe” libraries are installed there.)

We do not install AGE there by default for security reasons as this is the job of the DBA for that database.

aru-d-at:
@jrgemignani Noted. Thank you!

Do you think it would be within the best interest to have another function to connect for non-superuser? Of course they would require sufficient privilege to use the schema `ag_catalog`.

[...]

PR #1926, opened 2024-06-19
Add support for non-superuser loading of the age plugin

daanwa:
In relation to issue [839](https://github.com/apache/age/issues/839)

### DIFF
--- a/regress/expected/cypher_vle.out
+++ b/regress/expected/cypher_vle.out
@@ -1044,6 +1044,71 @@ NOTICE:  graph "issue_1043" has been dropped
  
 (1 row)
 
+-- issue 1910
+SELECT create_graph('issue_1910');
+NOTICE:  graph "issue_1910" has been created
+ create_graph 
+--------------
+ 
+(1 row)
+
+SELECT * FROM cypher('issue_1910', $$ MATCH (n) WHERE EXISTS((n)-[*1]-({name: 'Willem Defoe'}))
+                                      RETURN n.full_name $$) AS (full_name agtype);
+ full_name 
+-----------
+(0 rows)
+
+SELECT * FROM cypher('issue_1910', $$ CREATE ({name: 'Jane Doe'})-[:KNOWS]->({name: 'John Doe'}) $$) AS (result agtype);
+ result 
+--------
+(0 rows)
+
+SELECT * FROM cypher('issue_1910', $$ CREATE ({name: 'Donald Defoe'})-[:KNOWS]->({name: 'Willem Defoe'}) $$) AS (result agtype);
+ result 
+--------
+(0 rows)
+
+SELECT * FROM cypher('issue_1910', $$ MATCH (u {name: 'John Doe'})
+                                      MERGE (u)-[:KNOWS]->({name: 'Willem Defoe'}) $$) AS (result agtype);
+ result 
+--------
+(0 rows)
+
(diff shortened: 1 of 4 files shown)


daanwa:
In my initial PR I did not add the superuser parameter to the init file from which most of the time you initiate the connect.

drivers/python/age/__init__.py
```diff
@@ -24,12 +24,13 @@ def version():
     return VERSION.VERSION
 
 
-def connect(dsn=None, graph=None, connection_factory=None, cursor_factory=ClientCursor, **kwargs):
+def connect(dsn=None, graph=None, connection_factory=None, cursor_factory=ClientCursor, superuser:bool=True, **kwargs):
```
rafsun42, line 27: To be consistent with rest of the parameters, I would avoid specifiying the type.

drivers/python/age/age.py
```diff
@@ -39,9 +39,13 @@ def load(self, data: bytes | bytearray | memoryview) -> Any | None:
...
 
 
-def setUpAge(conn:psycopg.connection, graphName:str):
+def setUpAge(conn:psycopg.connection, graphName:str, superuser:bool=True):
     with conn.cursor() as cursor:
-        cursor.execute("LOAD 'age';")
+        if superuser:
+            cursor.execute("LOAD 'age';")
+        else:
+            cursor.execute("LOAD '$libdir/plugins/age';")
```
rafsun42, line 47: This assumes that the shared object is present in the `$libdir/plugins` directory. That may not be correct if DBA does not install age there.

rafsun42:
@daanwa The python driver test did not pass.

drivers/python/age/age.py
daanwa: I'm under the impression that while $libdir may change on the system, the "plugins" directory is somewhat static and should only contain plugins that may be used by non-super users. So if a DBA installs this in a different directory, it won't work either way. However I'm not a DBA and not fully aware of all postgres specifics. Perhaps we could leave it like this for now and see if/when this requires further configuration options.

drivers/python/age/__init__.py
daanwa: Makes sense.

daanwa:
The build fail seems to be due to some kind of timeout on files.pythonhosted.org. Not sure if there is anything I can change there.

drivers/python/age/age.py
rafsun42: @daanwa In that case, I suggest we rename the parameter to something more descriptive, such as `in_plugins_dir` or `load_from_plugins`. Additionally, we should explain what this parameter does in the documentation.

daanwa: Yes, agree, that's a better way to go about it. I'll use the `load_from_plugins` name.

rafsun42:
@daanwa Looks good. Could you please squash all the commits into one commit and add a body to the commit message briefly describing all the changes being added?

daanwa:
I Squashed the commits for my changes into two because I unfortunatly synced the fork between several of my changes. Hopefully this will do.

rafsun42:
@daanwa I see there is an unintended commit. Unfortunately, we cannot accept that with this PR. My suggestion is to rebase your changes to latest master. Ideally, your changes should be added to a non-master branch. It makes it easy to rebase in case master is updated before the PR is merged.

PR #1945, opened 2024-07-03
Add the `load_from_plugins` parameter in the Python driver to load AGE from the $libdir/plugins directory

daanwa:
Add the `load_from_plugins` parameter in the Python driver to load AGE from the $libdir/plugins directory

This is useful when a non-superuser login is used for the connection.

Additionally, fixed a small bug where a memoryview was not transformed to bytes before decoding into a UTF-8 string.


jrgemignani:
@daanwa There is an error with this PR as the Python driver check will not pass. Could you correct this?

rafsun42:
@jrgemignani @daanwa Closing this PR because PG11 is not supported anymore.

PR #1989, opened 2024-07-26
Fix issue 1986: Failure creating label name close to MAX_LABEL_NAME_LEN

jrgemignani:
Fixed the issue with creating a label name close to `MAX_LABEL_NAME_LEN`. Well, sort of.

The issue here is that a label name is the name of a relation in PostgreSQL. While the relation name is a `char *` it is basically converted to a PG type `Name` when it is used to create a relation. This conversion is a silent truncation, there are no warnings or error messages. The `Name` type is defined as follows -

```
    /*
     * Representation of a Name: effectively just a C string, but null-padded to
     * exactly NAMEDATALEN bytes.  The use of a struct is historical. 
     */ 
    typedef struct nameData
    {
        char data[NAMEDATALEN]; 
    } NameData;
    typedef NameData *Name;
```

This effectively gives us a max label name length of `NAMEDATALEN - 1`.

```
    /*
     * Maximum length for identifiers (e.g. table names, column names,
     * function names).  Names actually are limited to one fewer byte than this, 
     * because the length must include a trailing zero byte.
     * 
     * This should be at least as much as NAMEDATALEN of the database the 
     * applications run against. 
     */ 
    #define NAMEDATALEN 64
```

Since there isn't a way to get around this, the code was modified to reflect the usage of `NAMEDATALEN` for the length of label names. This required modifying the input parameters to be cstrings, for the functions `create_vlabel` and `create_elabel` which allows us to tell when `NAMEDATALEN` has been exceeded.

Additionally, the function `is_valid_label` was renamed to `is_valid_label_name` for consistency.

Regression tests were updated and added.

### DIFF
--- a/src/backend/utils/name_validation.c
+++ b/src/backend/utils/name_validation.c
@@ -55,12 +55,17 @@ int is_valid_graph_name(const char *graph_name)
  * @param label_type label type defined in label_commands.h
  * @return int
  */
-int is_valid_label(char *label_name, char label_type)
+int is_valid_label_name(char *label_name, char label_type)
 {
     int len = strlen(label_name);
 
     if (len < MIN_LABEL_NAME_LEN || len > MAX_LABEL_NAME_LEN)
     {
+        ereport(WARNING,
+                (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+                 errmsg("label name length not in range (%d <= length <= %d) length = %d",
+                        MIN_LABEL_NAME_LEN, MAX_LABEL_NAME_LEN, len)));
+
         return 0;
     }
 
--- a/age--1.5.0--y.y.y.sql
+++ b/age--1.5.0--y.y.y.sql
@@ -110,3 +110,20 @@ CREATE FUNCTION ag_catalog.graph_exists(graph_name name)
     RETURNS agtype
     LANGUAGE c
     AS 'MODULE_PATHNAME', 'age_graph_exists';
+
+CREATE FUNCTION ag_catalog.age_is_valid_label_name(agtype)
+    RETURNS boolean
+    LANGUAGE c
+    IMMUTABLE
+PARALLEL SAFE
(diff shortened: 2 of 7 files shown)


MironAtHome:
Review, just cursory, looks great. One moment to run actually run regression.

MironAtHome:
@jrgemignani 
I have still to cover regression script changes
Please see comments with suggestions.

MironAtHome (approved):

src/backend/utils/name_validation.c
```diff
@@ -55,12 +55,16 @@ int is_valid_graph_name(const char *graph_name)
  * @param label_type label type defined in label_commands.h
  * @return int
  */
-int is_valid_label(char *label_name, char label_type)
+int is_valid_label_name(char *label_name, char label_type)
```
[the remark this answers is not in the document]
jrgemignani: It would be nice to remove, but is there for a possible change in the future.

src/backend/utils/name_validation.c
```diff
@@ -55,12 +55,16 @@ int is_valid_graph_name(const char *graph_name)
...
 {
     int len = strlen(label_name);
 
     if (len < MIN_LABEL_NAME_LEN || len > MAX_LABEL_NAME_LEN)
     {
+        ereport(WARNING,
+                (errcode(ERRCODE_INVALID_PARAMETER_VALUE),
+                 errmsg("label name length not in range")));
```
MuhammadTahaNaveed, line 66: Would be nice to mention the range in warning as well...

jrgemignani:
@MuhammadTahaNaveed Added the range printing.

src/backend/utils/name_validation.c
jrgemignani: Done.

MuhammadTahaNaveed (approved):

PR #2005, opened 2024-08-06
Fix Issue 1988: How to update a property which is a keyword

jrgemignani:
Fixed issue 1988: Reserved Keyword Handling: How to update a property which is a keyword. Basically, how to use a keyword, aka reserved word, in a query.

Added the appropriate scanner and parser rules to allow keywords to be back ticked and used in queries as their name. For example, like in the following commands -

    MATCH (n:Person { id: 123 }) SET n.`match` = 'matched' RETURN n;
    MATCH (n:Person { id: 123 }) SET n.`create` = 'created' RETURN n;

The changes did not break any regression tests.
Added additional regression tests.

### DIFF
--- a/src/backend/parser/cypher_parser.c
+++ b/src/backend/parser/cypher_parser.c
@@ -50,7 +50,9 @@ int cypher_yylex(YYSTYPE *lvalp, YYLTYPE *llocp, ag_scanner_t scanner)
         ACCESS_PATH,
         ANY_EXISTS,
         ALL_EXISTS,
-        CONCAT
+        CONCAT,
+        CHAR,
+        BQIDENT
     };
 
     ag_token token;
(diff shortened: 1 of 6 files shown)


src/backend/parser/cypher_parser.c
```diff
@@ -93,6 +95,18 @@ int cypher_yylex(YYSTYPE *lvalp, YYLTYPE *llocp, ag_scanner_t scanner)
         lvalp->string = ident;
         break;
     }
+    case AG_TOKEN_BQIDENT:
+    {
+        char *ident;
+
+        /* these are identitiers, just back ticked */
```
MuhammadTahaNaveed, line 102: `identifiers` misspelled

MuhammadTahaNaveed (approved):
looks good to me! just a minor spelling issue

Issue #2027, opened 2024-08-08
get_graph_oid function declaration return type uint32 definition Oid


MironAtHome:
**Describe the bug**
Code bug.
Different declaration and definition of the function return parameter type

**How are you accessing AGE (Command line, driver, etc.)?**
N/A

**What data setup do we need to do?**
N/A

**What is the necessary configuration info needed?**
N/A

**What is the command that caused the error?**
N/A

**Expected behavior**
N/A

**Environment (please complete the following information):**
branches checked, master and PG16

**Additional context**
N/A

jrgemignani:
@MironAtHome If you are going to change the function return type, you need to verify that all locations that call it will not have an issue with that change.

MironAtHome:
I did.
The good part is that it's only declaration that had uint32.
Function definition + all function calls ( I could find ) had return value declared as Oid.
I don't want to bias someone looking at the code as code reviewer, so read next few lines with a grain of salt.
When I looked over function calls, that is get_graph_oid function, all of the calls return to variable named graph_oid, all, save for one, where it's named "graph" and its type is declared as Oid, as well.
This is true for branches master, 12, 13, 14, 15 and 16, branch 11, for some reason, had type declared, as Oid in function declaration ( header file ).
In case I missed one or another ( and I checked a few times ), please add comment, I am happy to fix.

jrgemignani:
@MironAtHome Just wanted to make sure.

PR #2037, opened 2024-08-09
Mironathome issue 2027 fix

MironAtHome:
Fix issue #2027
get_graph_oid function declaration return type uint32 definition Oid
Make both return Oid

### DIFF
--- a/src/include/catalog/ag_graph.h
+++ b/src/include/catalog/ag_graph.h
@@ -37,7 +37,7 @@ void insert_graph(const Name graph_name, const Oid nsp_id);
 void delete_graph(const Name graph_name);
 void update_graph_name(const Name graph_name, const Name new_name);
 
-uint32 get_graph_oid(const char *graph_name);
+Oid get_graph_oid(const char *graph_name);
 char *get_graph_namespace_name(const char *graph_name);
 
 List *get_graphnames(void);


jrgemignani:
@MironAtHome Have you verified that everywhere it is used uses Oid and not uint32? Additionally, if they do use uint32 why? (it could be a mistake from the migration away from Oids, btw)

MironAtHome:
> @MironAtHome Have you verified that everywhere it is used uses Oid and not uint32? Additionally, if they do use uint32 why? (it could be a mistake from the migration away from Oids, btw)

I did more than one time.
Everywhere function call is performed, it is returning to variable declared as Oid.
This is intuitive for the most cases, since variable, to which the function returns, is declared as graph_oid for all cases, save for one, where it is named just "graph" and the type of the variable declared is Oid.

MironAtHome:
> (it could be a mistake from the migration away from Oids, btw)
Any additional guidance would be appreciated.
If the actual plan is to migrate away from Oid type, I will create one more PR with unsigned int of 32 bit size (uint32).
In this, provided PR, the goal was to ensure that function declaration matches function definition and usage.
Plus get_graph_oid function name seems to carry reference to type oid.
If return type and usage in this planned PR to change away frm Oid, I would propose renaming function to get_graph_id, with new moniker "id" suggesting type independent treatment of graph numerical identifier used by graph database catalogue internally.

PR #2056, opened 2024-08-19
Remove all Discord mentions

shinyaaa:
Apache AGE's Discord has been shutdown and is no longer in use.
Therefore, I have removed the mention of Discord.


uhayat:
Change looks good to me

PR #2066, opened 2024-08-21
Fix memory leaks in functions part 1

jrgemignani:
Fixing memory leaks in functions due to contexts not clearing out the memory soon enough.

Background: Destroying a context will free up its associated memory. However, before that happens, a process can still accumulate a lot of memory working on large data sets if it expects the context free to do it for it. For example, when PG does quicksort the context isn't destroyed until after the qs has finished. This can cause memory to be exhausted.

Put in more aggressive freeing of memory. However, there are a lot of areas that need this applied. So, this is part 1 of at least 3 or 4 parts.

Worked mainly on `agtype.c` However, dealt with linked functions in other files.

    modified:   src/backend/utils/adt/agtype_util.c
    modified:   src/include/utils/agtype.h

### DIFF
--- a/src/backend/utils/adt/agtype.c
+++ b/src/backend/utils/adt/agtype.c
@@ -239,8 +239,11 @@ PG_FUNCTION_INFO_V1(graphid_recv);
 Datum graphid_recv(PG_FUNCTION_ARGS)
 {
     StringInfo buf = (StringInfo) PG_GETARG_POINTER(0);
+    int64 result = pq_getmsgint64(buf);
 
-    PG_RETURN_INT64(pq_getmsgint64(buf));
+    PG_FREE_IF_COPY(buf, 0);
+
+    PG_RETURN_INT64(result);
 }
 
 /*
@@ -278,6 +281,7 @@ Datum agtype_recv(PG_FUNCTION_ARGS)
     int version = pq_getmsgint(buf, 1);
     char *str = NULL;
     int nbytes = 0;
+    Datum result;
 
     if (version == 1)
     {
@@ -288,7 +292,12 @@ Datum agtype_recv(PG_FUNCTION_ARGS)
         elog(ERROR, "unsupported agtype version number %d", version);
     }
 
-    return agtype_from_cstring(str, nbytes);
+    result = agtype_from_cstring(str, nbytes);
+
+    PG_FREE_IF_COPY(buf, 0);
+    pfree(str);
+
+    return result;
 }
 
 /*
@@ -314,6 +323,8 @@ Datum agtype_send(PG_FUNCTION_ARGS)
     pfree(agtype_text->data);
     pfree(agtype_text);
 
+    PG_FREE_IF_COPY(agt, 0);
+
     PG_RETURN_BYTEA_P(pq_endtypsend(&buf));
 }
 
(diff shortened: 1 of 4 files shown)


src/backend/utils/adt/agtype.c
```diff
@@ -2704,10 +2774,14 @@ Datum agtype_to_int8(PG_FUNCTION_ARGS)
         if (temp->type != AGTV_ARRAY ||
             !temp->val.array.raw_scalar)
         {
+            int type = temp->type;
+
+            pfree_agtype_value(temp);
```
MuhammadTahaNaveed, line 2779: Shouldn't agtv be freed here as well?

src/backend/utils/adt/agtype.c
```diff
@@ -3324,8 +3416,11 @@ static agtype_value *execute_array_access_operator(agtype *array,
                 (errmsg("array index must resolve to an integer value")));
```
MuhammadTahaNaveed, line 3416: array_index_value needs to be freed here as well?

src/backend/utils/adt/agtype.c
jrgemignani: `ereport(ERROR` frees memory, so cleanup isn't necessary here. Additionally, I have removed all of the cleanups preceding `ereport(ERROR`

src/backend/utils/adt/agtype.c
jrgemignani: `ereport(ERROR` frees memory, so cleanup isn't necessary here. Additionally, I have removed all of the cleanups preceding `ereport(ERROR`

src/backend/utils/adt/agtype.c
```diff
@@ -3090,18 +3152,27 @@ Datum agtype_to_text(PG_FUNCTION_ARGS)
...
     /* Return null if arg_agt is null. This covers SQL and Agtype NULLS */
     if (arg_agt == NULL)
+    {
         PG_RETURN_NULL();
+    }
 
     /* check that we have a scalar value */
     if (!AGT_ROOT_IS_SCALAR(arg_agt))
+    {
+        PG_FREE_IF_COPY(arg_agt, 0);
```
MuhammadTahaNaveed, line 3162: We don't need this here, right?

src/backend/utils/adt/agtype.c
```diff
@@ -3968,8 +4093,14 @@ Datum agtype_access_operator(PG_FUNCTION_ARGS)
...
     }
 
+    pfree(args);
+    pfree(types);
+    pfree(nulls);
+
     /* serialize and return the result */
-    return AGTYPE_P_GET_DATUM(agtype_value_to_agtype(container_value));
+    result = agtype_value_to_agtype(container_value);
+
```
MuhammadTahaNaveed, line 4102: `container_value` should be freed here?

src/backend/utils/adt/agtype.c
jrgemignani: Good catch, I thought that I removed all those. It is unnecessary.

src/backend/utils/adt/agtype.c
jrgemignani: Unfortunately, due to how objects and arrays are iterated through, it can't be freed.

MuhammadTahaNaveed (approved):
Looks good to me!

PR #2117, opened 2024-09-23
Add index on id columns

MuhammadTahaNaveed:
- Whenever a label will be created, indices on id columns will be created by default. In case of vertex, a unique index on id column will be created, which will also serve as a unique constraint. In case of edge, a non-unique index on start_id and end_id columns will be created.

- This change is expected to improve the performance of queries that involve joins. From some performance tests, it was observed that the performance of queries improved alot.

- Loader was updated to insert tuples in indices as well. This has caused to slow the loader down a bit, but it was necessary.

- A bug related to command ids in cypher_delete executor was also fixed.


jrgemignani:
@MuhammadTahaNaveed Will you be able to get a chance to rebase this to the latest **master** branch? Or, would it be okay for someone else to take this on?

Issue #2111, opened 2024-09-15
PostgreSQL 17 not supported


JamesInform:
Hi all

I built age for PostgreSQL 17. Basically compiling worked without errors.

But when I try to create the extension I get the following error:

<img width="887" alt="image" src="https://github.com/user-attachments/assets/daf96f0f-cbf9-4660-b752-78d72b296a7a">

When will a PostgreSQL 17 compatible version of age be available?

Cheers,
James

jrgemignani:
@JamesInform Right now, PG 17 isn't in our roadmap. This could change in the future, but due to current resources, it is considered lower priority.

uhayat:
@JamesInform I created a PR ([2118](https://github.com/apache/age/pull/2118)) to resolve this issue. Currently there is no separate branch for branch for PG17,  so I added PG17 version number check.

@jrgemignani do we have a plan when should be create a separate branch, or who could create separate branch, or should this PR merged to master as is?

df7cb:
It's not ready for PG17 yet. Running the regression tests against PG17:
```
2024-10-28 09:43:48.644 UTC client backend[346252] pg_regress/expr ERROR:  labels() argument must be a vertex
2024-10-28 09:43:48.644 UTC client backend[346252] pg_regress/expr STATEMENT:  SELECT * from cypher('list', $$RETURN labels("string")$$) as (Labels agtype);
TRAP: failed Assert("HeapTupleIsValid(proctup)"), File: "src/backend/utils/ag_func.c", Line: 48, PID: 346252
postgres: myon contrib_regression [local] SELECT(ExceptionalCondition+0x75) [0x555e07a0fe45]
/usr/lib/postgresql/17/lib/age.so(+0x5be85) [0x7f0b17033e85]
/usr/lib/postgresql/17/lib/age.so(+0x2a04b) [0x7f0b1700204b]
postgres: myon contrib_regression [local] SELECT(range_table_entry_walker_impl+0x26) [0x555e077458b6]
postgres: myon contrib_regression [local] SELECT(range_table_walker_impl+0x6a) [0x555e07745a2a]
/usr/lib/postgresql/17/lib/age.so(+0x2ac0f) [0x7f0b17002c0f]
postgres: myon contrib_regression [local] SELECT(parse_analyze_fixedparams+0xac) [0x555e075f4a2c]
postgres: myon contrib_regression [local] SELECT(pg_analyze_and_rewrite_fixedparams+0x2a) [0x555e078b2b4a]
[...]

PR #2118, opened 2024-09-25
Fix error with PG17 palloc0fast compilation

uhayat:
In PG17 palloc0fast is removed and palloc0 is instead

### DIFF
--- a/src/backend/nodes/ag_nodes.c
+++ b/src/backend/nodes/ag_nodes.c
@@ -156,7 +156,7 @@ ExtensibleNode *_new_ag_node(Size size, ag_node_tag tag)
 {
     ExtensibleNode *n;
 
-    n = (ExtensibleNode *)palloc0fast(size);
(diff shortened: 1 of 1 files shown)


shinyaaa:
I think that change is correct.
However, in this project, the condition by PG_VERSION_NUM may be unnecessary, as the branches are separated by PG version.
(I don't know how to do this, as the PG17 branch has not yet been created.)

jrgemignani:
@uhayat I am okay with merging a change **where palloc0fast is replaced by palloc0**. But, we don't use version flags as stated by @shinyaaa above, so I won't approve that.

uhayat:
@jrgemignani understood. I updated the commit and removed the version check.

uhayat:
Anybody available for review ?

jrgemignani:
@uhayat See https://github.com/apache/age/issues/2111#issuecomment-2441084857

PR #2154, opened 2025-02-16
Update README for Rocky Linux support

uhayat:
Redhat discontinued CentOS and EOL was on June 2024.
Rocky Linux is replacement of CentOS and bug compatible with RHEL. This PR update the README along with pre-installation packages update for Rocky.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,9 +104,9 @@ Refer to our latest [Apache AGE documentation](https://age.apache.org/age-manual
 
 Install the following essential libraries according to each OS. Building AGE from the source depends on the following Linux libraries (Ubuntu package names shown below):
 
-- **CentOS**
+- **Rocky Linux**
 ```bash
-yum install gcc glibc glib-common readline readline-devel zlib zlib-devel flex bison
+yum install gcc glibc glibc-common readline readline-devel zlib zlib-devel flex bison perl
 ```
 - **Fedora**
 ```bash


jrgemignani:
A lot of people may continue to use CentOS, supported or not.

Shouldn't we keep CentOS in the README and just state that it is discontinued and future versions of AGE may not work, in addition to adding the replacement?

sorrell:
I tend to agree with John. I think a clearer explanation would be change the headings from `<distro>` to `<package manager> (example distros)`. Another example is Debian uses apt but only Ubuntu is listed. So maybe we should make the change to be:
`yum (Centos/Rocky Linux)` 
`apt (Debian/Ubuntu)`

uhayat:
@jrgemignani 
My thinking was to remove it from front face to encourage users to use stable platform (Rocky) because neither PostgreSQL nor the platform itself provide any update on CentOS 7 for quite sometime, and its not recommend for production use anymore.
Of course we can maintain a separate list with all support platform in documentation. ReadMe list neither cover our CI nor release testing platforms, but theoretically we do support all other distributions and platform which PostgreSQL support.

@sorrell dependancies are `distro` dependant, not `package manager` dependant. 
e.g 'dnf` is same for RHEL 7/8/9, CentOS and Rocky Linux . But dependency list for all of them are different.

And I do agree we should combine `distros`, when/if dependencies are same. I am open to ideas.

uhayat:
`pr_keep_alive = 1`

PR #2156, opened 2025-02-23
Update CI, README and repo settings for PG17

MuhammadTahaNaveed:
- Updated all the github workflows
- Updated the README
- Updated repo settings
- Updated the Dockerfiles

### Update:
Currently, all workflows are targeting the `PG17_prepare` branch, which will be changed to `PG17` once the branch is renamed.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -49,6 +49,10 @@ github:
       required_pull_request_reviews:
         required_approving_review_count: 2
 
+    PG17:
+      required_pull_request_reviews:
+        required_approving_review_count: 2
+
     PG16:
       required_pull_request_reviews:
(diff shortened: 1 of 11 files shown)


uhayat:
@MuhammadTahaNaveed all workflows are target PG17 branch, IMO they will be executed when branch will be renamed to `PG17` from `PG17_prepare`. Is this intentional ?

jrgemignani:
I think @MuhammadTahaNaveed was hoping to make this an easy PR for setting up PG17. But, we can't always assume that preparing a new version is going to be simple. It is probably best to pretend this version may take some work to get it ready and prepare it the long way.

jrgemignani (approved):
Looks good

Issue #1611, opened 2024-02-14
List comprehension as property constraint of MERGE clause errors out


MuhammadTahaNaveed:
**Describe the bug**
List comprehension as property constraint of MERGE clause errors out.

**How are you accessing AGE (Command line, driver, etc.)?**
- command line

**What is the command that caused the error?**
```pgsql
SELECT * from cypher('graph_name', $$
  MERGE ({list:[u IN [1,2,3]]})
$$) as (a agtype);
```
```
ERROR:  Aggref found in non-Agg plan node
```

**Expected behavior**
Should execute without any error.

**Environment (please complete the following information):**
- Version: latest

diangamichael:
To apply a workaround to the query and avoid using list comprehensions in the MERGE clause, you can modify the query to precompute the list before using it in the MERGE clause. Here's how you can rewrite the query:

WITH input_list AS (
  SELECT [1,2,3] AS list
)
SELECT * FROM cypher('graph_name', $$
  MERGE ({list: u.list})
  ON CREATE SET ...
  ON MATCH SET ...
$$) AS (a agtype);
In this modified query:

We first create a Common Table Expression (CTE) named input_list to compute the list [1,2,3].
Then, we reference this computed list u.list within the MERGE clause.
You need to replace ON CREATE SET ... and ON MATCH SET ... with the appropriate actions you want to perform when the MERGE clause creates or matches a node.
By precomputing the list outside the MERGE clause, we avoid using list comprehensions directly within it, potentially bypassing the error you encountered.





To apply a workaround to the query and avoid using list comprehensions in the MERGE clause, you can modify the query to precompute the list before using it in the MERGE clause. Here's how you can rewrite the query:

sql
Copy code
WITH input_list AS (
  SELECT [1,2,3] AS list
)
SELECT * FROM cypher('graph_name', $$
  MERGE ({list: u.list})
  ON CREATE SET ...
  ON MATCH SET ...
$$) AS (a agtype);
In this modified query:

We first create a Common Table Expression (CTE) named input_list to compute the list [1,2,3].
Then, we reference this computed list u.list within the MERGE clause.
You need to replace ON CREATE SET ... and ON MATCH SET ... with the appropriate actions you want to perform when the MERGE clause creates or matches a node.
By precomputing the list outside the MERGE clause, we avoid using list comprehensions directly within it, potentially bypassing the error you encountered.

sir-sa:
SELECT * FROM cypher('graph_name', $$
[...]

PR #2163, opened 2025-04-16
Fix CI build errors caused by missing dependencies

MuhammadTahaNaveed:
Adds extra step in CI workflow to install missing dependencies.

PR #2169, opened 2025-04-27
Reimplement list comprehension

MuhammadTahaNaveed:
- Reimplement list comprehension to use ARRAY sublink.
- Some test results were not correct. All the test results that are
  modified are correct and are verified with neo4j.
- Also resolves the issue of using list comprehension in MERGE clause (#1611 )
  and issue #1850.


MuhammadTahaNaveed:
CI failure is due to #2163

KhoobBabe:
I have seen a monkey write better code

PR #2172, opened 2025-05-11
Add support for operators in cypher query

MuhammadTahaNaveed:
- Added support for PG operators in cypher. Some hardcoded operators were removed, since they are now covered by the general operator handling.
- Added full typecast syntax that allows for type modifiers.
- These changes also improve interoperability with other extensions, as reflected in the regression tests.
- Fixed some operator signatures in .sql


src/backend/catalog/ag_graph.c
```diff
@@ -173,3 +173,14 @@ char *get_graph_namespace_name(const char *graph_name)
 {
     return get_namespace_name(get_graph_namespace(graph_name));
 }
+
+bool graph_namespace_exists(Oid graph_oid)
+{
+    graph_cache_data *cache_data;
+
+    cache_data = search_graph_namespace_cache(graph_oid);
+    if (cache_data)
```
jrgemignani, line 182: needs {} for if/else

Issue #2174, opened 2025-05-18
Bug: load_edges_from_file Fails to Properly Import Single Property from CSV


Munmud:
# Bug: load_edges_from_file Fails to Import Single Property from CSV
When loading edges from a CSV file using load_edges_from_file, any single property included in the file is not added to the edge’s properties in Apache AGE.

### 🔧 Command Used
```sql
SELECT load_edges_from_file(
  'test_graph',
  'has_city',
  '/home/munmud/Desktop/pg_new_installation/src/dataset/edges_sample.csv'
);
```

###  CSV files data
```csv
start_id,start_vertex_type,end_id,end_vertex_type,new_field
153,City,3,Country,new_data
154,City,3,Country,new_data
```

### Result
edges                                                                                                 
-----------------------------------------------------------------------------------------------------------------------------------
{"id": 1407374883553281, "label": "has_city", "end_id": 1125899906842627, "start_id": 844424930132121, "properties": {}}::edge
(1 row)

### ✅ Expected Behavior
The edge should include the single property from the CSV in the properties object, like so:

```
"properties": {"new_field": "new_data"}
```

PR #2175, opened 2025-05-18
Fix BUG (Isssue #2174) CSV import for edge with one property

Munmud:
fix isssue #2174 (load_edges_from_file Fails to Properly Import Single Property from CSV)


MuhammadTahaNaveed (approved):

MuhammadTahaNaveed:
@Munmud Thanks for the fix 👍🏻

PR #2188, opened 2025-06-11
Reimplement list comprehension (#2169)

MuhammadTahaNaveed:
- Reimplement list comprehension to use ARRAY sublink.
- Some test results were not correct. All the test results that are
  modified are correct and are verified with neo4j.
- Also resolves the issue of using list comprehension in MERGE clause (#1611)
  and issue #1850.


jrgemignani:
@MuhammadTahaNaveed Missing commit message and description?

image

PR #2199, opened 2025-08-08
Add additional user variables during node and edge transform

jrgemignani:
This PR adds additional variables to a node and edge during the transform node/edge phase, provided that a variable was specified for the node or edge.

These additional variables are added to allow direct access to the columns of the specific node or edge. These variables can be used to improve the performance of some queries.

Variables added have the following suffixes to the node or edge variable name -

    _idc            ID Column
    _propertiesc    PROPERTIES Column
    _start_idc      EDGE START_ID Column
    _end_idc        EDGE END_ID Column

For example -

    MATCH (u) RETURN u, u_idc, u_propertiesc

Regression tests added.

    modified:   regress/expected/cypher_match.out
    modified:   regress/sql/cypher_match.sql
    modified:   src/backend/parser/cypher_item.c
    modified:   src/include/parser/cypher_parse_node.h


jrgemignani:
@MuhammadTahaNaveed Please test and suggest additional regression tests, if you have the time.

PR #2258, opened 2025-12-03
Fix possible memory and file descriptors leaks

ZigzagAK:
Please, review this patch.

### DIFF
--- a/src/backend/utils/adt/agtype.c
+++ b/src/backend/utils/adt/agtype.c
@@ -184,8 +184,11 @@ static agtype_value *tostring_helper(Datum arg, Oid type, char *msghdr);
 
 void *repalloc_check(void *ptr, size_t len)
 {
-   if (ptr)
+   if (ptr != NULL)
+   {
        return repalloc(ptr, len);
+   }
+
    return palloc(len);
 }
 


src/backend/utils/adt/agtype.c
```diff
@@ -181,6 +181,14 @@ static agtype_value *agtype_build_map_as_agtype_value(FunctionCallInfo fcinfo);
 agtype_value *agtype_composite_to_agtype_value_binary(agtype *a);
 static agtype_value *tostring_helper(Datum arg, Oid type, char *msghdr);
 
+
+void *repalloc_check(void *ptr, size_t len)
+{
+   if (ptr)
```
jrgemignani: Our current standard for all new code (for the past few years) is to always use {} for all control blocks (if, for, while, etc.)

src/backend/utils/adt/agtype.c
```diff
@@ -5708,8 +5716,6 @@ Datum age_startnode(PG_FUNCTION_ARGS)
 
     result = get_vertex(graph_name, label_name, start_id);
 
-    free(label_name);
```
jrgemignani, line 5711: Why remove this?

src/backend/utils/adt/agtype.c
```diff
@@ -5773,8 +5779,6 @@ Datum age_endnode(PG_FUNCTION_ARGS)
 
     result = get_vertex(graph_name, label_name, end_id);
 
-    free(label_name);
```
jrgemignani, line 5776: Why remove this?

jrgemignani (changes requested):
See comments. Looks good otherwise.

src/backend/utils/adt/agtype.c
ZigzagAK: Because `result = strdup(result);` was replaced with `result = pstrdup(result);` in `get_label_name` by this PR.
We can't use standard malloc/realloc/free/strdup/... functions safely because in every moment when we use postgresql functions our codeflow may be breaked by longjmp via ereport(ERROR). We must use memory context functions to allocate memory in the request. All allocated memory will be freed at the end of current request.
We can use `pfree`, but it is not required for short-lived requests.

src/backend/utils/adt/agtype.c
ZigzagAK: See previous comment.

src/backend/utils/adt/agtype.c
ZigzagAK: fixed

jrgemignani (changes requested):
@ZigzagAK  There was an update to the master branch that also had changes to age load. Could you rebase this PR off of the latest master, please :) Otherwise, everything looks good.

ZigzagAK:
> @ZigzagAK There was an update to the master branch that also had changes to age load. Could you rebase this PR off of the latest master, please :) Otherwise, everything looks good.

Already rebased.

MuhammadTahaNaveed (approved):
looks good to me

PR #2300, opened 2026-01-09
Fix and improve index.sql regression test coverage

jrgemignani:
NOTE: This PR was created with AI tools and a human.

- Remove unused copy command (leftover from deleted agload_test_graph test)
- Replace broken Section 4 that referenced non-existent graph with comprehensive WHERE clause tests covering string, int, bool, and float properties with AND/OR/NOT operators
- Add EXPLAIN tests to verify index usage:
  - Section 3: Validate GIN indices (load_city_gin_idx, load_country_gin_idx) show Bitmap Index Scan for property matching
  - Section 4: Validate all expression indices (city_country_code_idx, city_id_idx, city_west_coast_idx, country_life_exp_idx) show Index Scan for WHERE clause filtering

All indices now have EXPLAIN verification confirming they are used as expected.

modified:   regress/expected/index.out
modified:   regress/sql/index.sql

### DIFF
--- a/regress/expected/index.out
+++ b/regress/expected/index.out
@@ -16,7 +16,6 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-\! cp -r regress/age_load/data regress/instance/data/age_load
 LOAD 'age';
 SET search_path TO ag_catalog;
 SET enable_mergejoin = ON;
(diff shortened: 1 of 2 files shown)


regress/expected/index.out
```diff
@@ -441,26 +466,259 @@ DROP INDEX cypher_index.load_country_gin_idx;
...
+    WHERE c.life_expectancy < 76.0
+    RETURN c.name
+$$) as (name agtype);
+   name   
+----------
+ "Mexico"
+(1 row)
+
+-- Test WHERE in combination with pattern matching
+SELECT * FROM cypher('cypher_index', $$
```
MuhammadTahaNaveed, line 704: Explain for pattern with where should be added as well imo. Also an explain for pattern (country:Country)<-[:has_city]-(city:City) with filters on both country and city.

jrgemignani: > Explain for pattern with where should be added as well imo.

Explain with WHERE is done above, many times, already.

MuhammadTahaNaveed: > > Explain for pattern with where should be added as well imo.
> 
> Explain with WHERE is done above, many times, already.

I was referring to `pattern with where`. There isn't one showing index scan being used for this case. But it seems the primary reason of this PR is to resolve a bad test, not improving the coverage. So going to resolve it.

PR #2313, opened 2026-01-21
Fix null pointer handling in array iteration

ProjectMutilation:
Previously, when iterating through an agtype container, the code would access `elem->val` even when `elem` was null.
This adds a null check to set the result type to AGTV_NULL when the element is null, preventing a potential segmentation fault.

Fixes: 4274f10 ("Added the toStringList() function (#1084)") 
Found by PostgresPro.


MuhammadTahaNaveed:
@ProjectMutilation Can you please rebase this PR onto the current master? That should resolve the failing checks.

ProjectMutilation:
> @ProjectMutilation Can you please rebase this PR onto the current master? That should resolve the failing checks.


rebased

Issue #2092, opened 2024-09-08
"where is not null" causes segmentation fault


dberardo-com:
"where is not null" statement in this query causes postgres to crash with the following error: "server process (PID 327) was terminated by signal 11: Segmentation fault"

```
SELECT * FROM cypher('access_control_graph', $$
    MATCH direct_group= (u:Operator {name: 'Admin'})-[:BELONGS_TO*]->(nop:Group {_type: 'NOP'})-[:PARENT_OF*]->(g2:Group)
    OPTIONAL MATCH plants=(nop)-[:RESPONSIBLE_FOR*]->(inp:Group {_type: 'PLANT'})  // subgroups
    OPTIONAL MATCH direct_assets=(nop)-[:HAS_PERMISSION*]->(r:Resource {_type: 'ASSET'}) 
    OPTIONAL MATCH indirect_assets=(inp)-[:HAS_PERMISSION*]->(r1:Resource {_type: 'ASSET'})
    where inp IS NOT NULL
    
    RETURN direct_group, 
        plants
        ,direct_assets
      //  ,indirect_assets
$$) AS (direct_group agtype, 
plants agtype 
,direct_assets agtype
--,indirect_assets agtype
);
```

MuhammadTahaNaveed:
@dberardo-com Can you please share the following;

- **AGE version**
- **Postgres version**
- **Data setup we need to reproduce this bug**

dberardo-com:
i have been using the latest official docker image of age

MuhammadTahaNaveed:
@dberardo-com I was able to reproduce the bug. We will look into it and update you as soon as we have a fix for it.

gregfelice:
I've submitted a fix for this in #2337. The segfault was caused by a NULL pointer dereference in `build_local_vle_context()` when the cached VLE context's vertex list was exhausted. The fix also addresses the related `match_vle_terminal_edge() arguments cannot be NULL` error that occurs with the same query pattern without the `WHERE` clause — both are symptoms of VLE matching functions not handling NULL arguments from OPTIONAL MATCH (LEFT JOIN) contexts.

PR #2337, opened 2026-02-26
Fix VLE NULL handling for chained OPTIONAL MATCH

gregfelice:
## Problem

Chained `OPTIONAL MATCH` clauses with VLE (variable-length edges) crash or error when a preceding `OPTIONAL MATCH` produces NULL results.

**Segfault (with `WHERE IS NOT NULL`):**
```cypher
MATCH (u)-[:BELONGS_TO*]->(nop)-[:PARENT_OF*]->(g2)
OPTIONAL MATCH (nop)-[:RESPONSIBLE_FOR*]->(inp)
OPTIONAL MATCH (inp)-[:HAS_PERMISSION*]->(r1)
WHERE inp IS NOT NULL
RETURN u, inp, r1
```
Server terminates with signal 11 (SIGSEGV).

**Error (without `WHERE`):**
Same query without the `WHERE` clause produces:
`ERROR: match_vle_terminal_edge() arguments cannot be NULL`

Both are incorrect. `OPTIONAL MATCH` implements a LEFT JOIN — when the inner side produces no match, the result should be NULL-extended rows, not crashes or errors.

Reported in #2092 with full reproduction data.

## Root Cause

Two related problems in `src/backend/utils/adt/age_vle.c`:

**1. NULL pointer dereference in `build_local_vle_context()` (line 564)**

On the cached VLE context path, when the start vertex argument is NULL (from OPTIONAL MATCH), the code calls `get_graphid(vlelctx->next_vertex)` without checking whether `next_vertex` is NULL. When the cached context's vertex list is exhausted from a prior SRF call, `next_vertex` is NULL, and `get_graphid(NULL)` dereferences `NULL->id`.

GDB backtrace confirms:
```
#0 get_graphid (node=0x0) at src/backend/utils/adt/age_graphid_ds.c:56
#1 build_local_vle_context at src/backend/utils/adt/age_vle.c:564
#2 age_vle at src/backend/utils/adt/age_vle.c:1734
```

The `WHERE IS NOT NULL` expression causes PostgreSQL's optimizer to restructure the execution plan, which triggers the cached context path with an exhausted vertex list — a code path that doesn't occur without the `WHERE` clause.

**2. VLE matching functions error on NULL arguments instead of returning FALSE**

Three functions used as join quals — `age_match_vle_terminal_edge`, `age_match_two_vle_edges`, and `age_match_vle_edge_to_id_qual` — call `ereport(ERROR)` when any argument is NULL. In a LEFT JOIN context, NULL arguments are expected (the inner side produced no match). The correct response is `PG_RETURN_BOOL(false)`, which tells PostgreSQL "no match" and triggers NULL-extended row emission.

## Fix

- **`build_local_vle_context`**: Check `next_vertex` for NULL before dereferencing; return NULL when the vertex list is exhausted.
- **`age_vle`**: Handle NULL return from `build_local_vle_context` with `SRF_RETURN_DONE`.
- **`age_match_vle_terminal_edge`**: Return FALSE on SQL NULL and agtype NULL arguments instead of `ereport(ERROR)`.
- **`age_match_two_vle_edges`**: Return FALSE on NULL arguments.
- **`age_match_vle_edge_to_id_qual`**: Return FALSE on NULL arguments.

## Regression Tests

Added tests to `regress/sql/cypher_vle.sql` covering:
- Chained OPTIONAL MATCH with VLE + `WHERE IS NOT NULL` (was: segfault)
- Chained OPTIONAL MATCH with VLE without `WHERE` (was: error)
- Happy path verification (data exists through the full chain)

All 31 regression tests pass.

## AI Disclosure

### DIFF
--- a/regress/expected/cypher_vle.out
+++ b/regress/expected/cypher_vle.out
@@ -1109,6 +1109,92 @@ NOTICE:  graph "issue_1910" has been dropped
  
 (1 row)
 
+-- issue 2092: VLE with chained OPTIONAL MATCH and NULL handling
+-- Previously, chained OPTIONAL MATCH with VLE would either segfault
+-- (with WHERE IS NOT NULL) or error with "arguments cannot be NULL"
+-- (without WHERE) instead of producing correct NULL-extended rows.
+SELECT create_graph('issue_2092');
+NOTICE:  graph "issue_2092" has been created
+ create_graph 
+--------------
+ 
+(1 row)
+
+-- Set up a small graph where some OPTIONAL MATCH paths exist and some don't
+SELECT * FROM cypher('issue_2092', $$
+    CREATE (a:Person {name: 'Alice'})
+    CREATE (b:Person {name: 'Bob'})
+    CREATE (c:City {name: 'NYC'})
+    CREATE (d:City {name: 'LA'})
+    CREATE (e:Place {name: 'Central Park'})
+    CREATE (a)-[:LIVES_IN]->(c)
+    CREATE (c)-[:HAS_PLACE]->(e)
+    CREATE (b)-[:LIVES_IN]->(d)
+$$) AS (result agtype);
+ result 
+--------
+(0 rows)
+
+-- Alice lives in NYC which has Central Park.
+-- Bob lives in LA which has no places.
+-- VLE + chained OPTIONAL MATCH + WHERE IS NOT NULL: should return rows
+-- without crashing (was: segfault)
+SELECT * FROM cypher('issue_2092', $$
+    MATCH (p:Person)-[:LIVES_IN*]->(c:City)
+    OPTIONAL MATCH (c)-[:HAS_PLACE*]->(place)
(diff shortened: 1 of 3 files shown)


jrgemignani (approved):

jrgemignani:
@gregfelice Ty for the PR! 

Two points -

- If you can look at the modification Copilot suggested above and apply it, if correct.
- If you used an AI, please be sure to note it in the commit log message and PR description. We're not opposed to their usage, just prefer it being noted.

gregfelice:
@jrgemignani Thanks for the review!

Pushed a follow-up commit addressing both points:

1. **Copilot suggestion applied** — fixed the error message in `age_match_vle_terminal_edge()` (was incorrectly referencing `age_match_terminal_edge()`), and added `ORDER BY p.name` to all three regression test queries to avoid nondeterministic row ordering.

2. **AI disclosure** — added to the PR description and commit log. Claude (Anthropic) was used to assist in developing this fix.

All 31 regression tests pass.

PR #2342, opened 2026-02-26
Support doubled-quote escaping in Cypher string literals

### LINKED ISSUE #2222: Error passing strings containing apostrophes as parameters
**Describe the bug**
Passing strings containing apostrophes/single quotes (') as parameters throws an error. There is a work around to get this working, however its unsustainable.

- python driver (psycopg2)

```python

import psycopg2 
import age
import json

GRAPH_NAME = "test_graph"
conn = psycopg2.connect(host="localhost", port="5432", dbname="postgres", user="postgres", password="password")

age.setUpAge(conn, GRAPH_NAME)

with conn.cursor() as cursor:
    try :
        cursor.execute(
            """
            SELECT * from cypher(
                %s, 
                $$ CREATE (c:testnode {node_id: %s, node_text: %s}) RETURN c $$
            ) as (c agtype); 
            """, 
            (GRAPH_NAME , "test_id", "This isn't working!") )
    except Exception as ex:
        print(type(ex), ex)
        # if exception occurs, you must rollback all transaction. 
        conn.rollback()
    else:
        conn.commit()
        result = cursor.fetchall()
        print(result)
```

```python
 cursor.execute(
            """
            SELECT * from cypher(
                %s, 
                $$ CREATE (c:testnode {node_id: %s, node_text: %s}) RETURN c $$
            ) as (c agtype); 
            """, 
            (GRAPH_NAME , "test_id", "This isn't working!") )
```
```
ERROR:  <class 'psycopg2.errors.SyntaxError'> syntax error at or near "'t working!'"
LINE 4: ...ATE (c:testnode {node_id: 'test_id', node_text: 'This isn''t working...
```

**Expected behavior**
The expected output should be:
```python
[({label:testnode, id:1688849860263941, properties:{node_id: test_id, node_text: This isn't working!, }}::VERTEX,)]
```

- Age Version: 1.5.0
- Postgresql Version: 1.6.0

**Additional context**
Enclosing the string in single quotes rather than double quotes and escaping the apostrophe within the string by adding another apostrophe seems to work, although it removes the apostrophe entirely.

```python
with conn.cursor() as cursor:
    try :
        cursor.execute(
            """
            SELECT * from cypher(
                %s, 
                $$ CREATE (c:testnode {node_id: %s, node_text: %s}) RETURN c $$
            ) as (c agtype); 
            """, 
            GRAPH_NAME , "test_id", 'This isn''t working!') )
    except Exception as ex:
        print(type(ex), ex)
        # if exception occurs, you must rollback all transaction. 
        conn.rollback()
    else:
        conn.commit()
        result = cursor.fetchall()
        print(result)
```
```python
OUTPUT: [({label:testnode, id:1688849860263941, properties:{node_id: test_id, node_text: This isnt working!, }}::VERTEX,)]
```
Notice how the output doesn't contain the apostrophe in the word: `isn't`.
This doesn't appear sustainable in an application because while its possible to replace every occurrence of an apostrophe with two apostrophes, I'm not sure its possible to explicitly declare a string with single quotes rather than double quotes.

gregfelice:
## Issue

Fixes #2222

## Problem

SQL drivers (psycopg2, JDBC, Node.js `pg`, etc.) escape single quotes by doubling them when substituting parameters into queries. For example, psycopg2 turns `"isn't"` into `'isn''t'`. When these substitutions land inside Cypher's `$$` block, the Cypher scanner rejects them with a syntax error:

```python
cursor.execute("""
    SELECT * FROM cypher(%s, $$ CREATE (n {text: %s}) RETURN n $$) AS (r agtype);
""", ("graph", "isn't"))
```

```
ERROR: syntax error at or near "'t'"
```

The scanner only recognizes backslash escaping (`\'`), not SQL-style doubled-quote escaping (`''`).

## Fix

Add `''` and `""` as escape sequences in the Cypher scanner's `sqstr` and `dqstr` states respectively. This follows the same pattern already used for backtick-quoted identifiers, where ` `` ` is handled by the `esbquote` rule.

Flex's longest-match rule ensures the two-character `''` match takes priority over the one-character closing-quote match. The change is fully backwards-compatible since `''` inside a single-quoted string was previously a syntax error.

**Before:**
```sql
SELECT * FROM cypher('g', $$ RETURN 'isn''t' $$) AS (r agtype);
-- ERROR: syntax error at or near "'t'"
```

**After:**
```sql
SELECT * FROM cypher('g', $$ RETURN 'isn''t' $$) AS (r agtype);
--  r
-- ---------
--  "isn't"
```

## Changes

- `src/backend/parser/ag_scanner.l`: Added `esdquote` and `essquote` pattern definitions and corresponding rules in the `dqstr`/`sqstr` scanner states
- `regress/sql/scan.sql`: 5 regression tests covering single-quote, double-quote, leading/trailing, and multiple escaped quotes
- `regress/expected/scan.out`: Updated expected output

All 31 regression tests pass.

## Note

While this fix makes SQL driver parameter substitution work, the recommended approach for passing dynamic values to Cypher is through AGE's parameter mechanism using prepared statements, which avoids string escaping entirely:

```sql
PREPARE q(agtype) AS
SELECT * FROM cypher('g', $$ CREATE (n {text: $text}) RETURN n $$, $1) AS (r agtype);
EXECUTE q('{"text": "isn''t"}'::agtype);
```

## AI Disclosure

### DIFF
--- a/regress/expected/scan.out
+++ b/regress/expected/scan.out
@@ -294,6 +294,47 @@ $$) AS t(a agtype, b agtype, c agtype);
  " \" \" ' ' " | " ' ' \" \" " | " / / \\ \b \f \n \r \t "
 (1 row)
 
+-- doubled-quote escape sequences for SQL driver compatibility (issue #2222)
+SELECT * FROM cypher('scan', $$
+RETURN 'isn''t'
+$$) AS t(a agtype);
+    a    
+---------
+ "isn't"
+(1 row)
+
+SELECT * FROM cypher('scan', $$
+RETURN '''hello'
+$$) AS t(a agtype);
+    a     
+----------
+ "'hello"
+(1 row)
+
+SELECT * FROM cypher('scan', $$
+RETURN 'hello'''
+$$) AS t(a agtype);
+    a     
+----------
+ "hello'"
+(1 row)
+
+SELECT * FROM cypher('scan', $$
+RETURN 'it''s a ''test'''
+$$) AS t(a agtype);
+        a        
+-----------------
+ "it's a 'test'"
+(1 row)
+
+SELECT * FROM cypher('scan', $$
+RETURN "she said ""hello"""
+$$) AS t(a agtype);
+          a           
+----------------------
+ "she said \"hello\""
+(1 row)
+
 -- invalid escape sequence
 SELECT * FROM cypher('scan', $$
 RETURN "\a"
--- a/regress/sql/scan.sql
+++ b/regress/sql/scan.sql
@@ -202,6 +202,27 @@ SELECT * FROM cypher('scan', $$
 RETURN " \" \" ' \' ", ' \' \' " \" ', " / \/ \\ \b \f \n \r \t "
 $$) AS t(a agtype, b agtype, c agtype);
 
+-- doubled-quote escape sequences for SQL driver compatibility (issue #2222)
+SELECT * FROM cypher('scan', $$
+RETURN 'isn''t'
+$$) AS t(a agtype);
+
+SELECT * FROM cypher('scan', $$
+RETURN '''hello'
+$$) AS t(a agtype);
+
+SELECT * FROM cypher('scan', $$
+RETURN 'hello'''
+$$) AS t(a agtype);
+
+SELECT * FROM cypher('scan', $$
+RETURN 'it''s a ''test'''
+$$) AS t(a agtype);
+
+SELECT * FROM cypher('scan', $$
+RETURN "she said ""hello"""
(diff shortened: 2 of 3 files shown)


gregfelice:
Friendly ping — this PR fixes doubled-quote escaping in Cypher string literals (issue #2222). The fix is straightforward: the Cypher lexer's string-processing loop now correctly collapses `''` → `'` and `""` → `"` per the openCypher spec. Regression tests included. Would appreciate a review when someone has bandwidth. Thanks\!

jrgemignani (approved):

PR #2365, opened 2026-04-02
Fix nondeterministic age_global_graph regression test

jrgemignani:
The age_global_graph test had two issues that could cause intermittent failures:

1. Nondeterministic warning output: The graph_stats() call on a graph with deliberately deleted vertices produces WARNING messages for dangling edges. These warnings are emitted by iterating edge label tables (knows, stalks), and the iteration order is not guaranteed. Since PostgreSQL WARNING messages cannot be caught or counted from SQL (only ERROR and above are catchable via PL/pgSQL exception handling), we suppress them with SET client_min_messages = error. The suppressed warnings are documented verbatim in comments. The graph_stats() result row still validates correct dangling-edge handling.

2. Nondeterministic row ordering: Multiple MATCH...RETURN queries returned multi-row results without ORDER BY, relying on scan order. Added ORDER BY id(u), id(v), id(e), id(n), or id(a) as appropriate to all MATCH...RETURN queries for future-proofing, even those currently returning a single row.

Files changed:
  regress/sql/age_global_graph.sql
  regress/expected/age_global_graph.out

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/regress/expected/age_global_graph.out
+++ b/regress/expected/age_global_graph.out
@@ -44,19 +44,19 @@ SELECT * FROM cypher('ag_graph_3', $$ CREATE (v:vertex3) RETURN v  $$) AS (v agt
 (1 row)
 
 -- load contexts using the vertex_stats command
-SELECT * FROM cypher('ag_graph_3', $$ MATCH (u) RETURN vertex_stats(u) $$) AS (result agtype);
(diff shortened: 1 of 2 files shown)


tcolo:
@jrgemignani I reviewed the patch it looks fine to me - claude code didn't find anything and also compile and make install works on macosx without problems.

MuhammadTahaNaveed (approved):

PR #2380, opened 2026-04-15
Fix OPTIONAL MATCH dropping null-preserving rows with subquery WHERE (#2378)

### LINKED ISSUE #2378: `OPTIONAL MATCH` may incorrectly drop null-preserving outer rows when its `WHERE` clause contains a correlated subquery predicate.
**Describe the bug**
`OPTIONAL MATCH` may incorrectly drop null-preserving outer rows when its `WHERE` clause contains a correlated subquery predicate.

In the repro below, no matched `friend` satisfies the `EXISTS { ... }` predicate. Under Cypher optional semantics, each outer `p` row should still be preserved with `friend = null`.

Instead, Apache AGE returns no rows at all.

```pgsql
  CREATE (a:Person {name: 'Alice', age: 30}),
         (b:Person {name: 'Bob', age: 25}),
         (c:Person {name: 'Charlie', age: 35}),
         (a)-[:KNOWS]->(b),
         (a)-[:KNOWS]->(c)
$$) AS (v agtype);
```

- PostgreSQL version: `18.1`

```pgsql
  MATCH (p:Person)
  OPTIONAL MATCH (p)-[:KNOWS]->(friend:Person)
  WHERE EXISTS { (friend)-[:KNOWS]->(:Person) }
  RETURN p.name AS name, friend.name AS friend
  ORDER BY name, friend
$$) AS (name agtype, friend agtype);
```

Returned result on AGE:
```text
(0 rows)
```

**Expected behavior**
Since neither `Bob` nor `Charlie` knows another person in this setup, the optional pattern should fail to bind `friend`, but the outer `p` rows should remain:

```text
Alice, null
Bob, null
Charlie, null
```

- Version: Apache AGE `1.7.0`
- PostgreSQL: `18.1`
- Host OS: Windows 10
- Architecture: x86_64
- Deployment: Docker

**Additional context**
Two control cases suggest this is specifically tied to correlated subquery predicates on `OPTIONAL MATCH`, not to `OPTIONAL MATCH` in general.

Control case 1, without the correlated subquery predicate, behaves as expected:

```pgsql
  MATCH (p:Person)
  OPTIONAL MATCH (p)-[:KNOWS]->(friend:Person)
  RETURN p.name AS name, COUNT(DISTINCT friend.name) AS friendCount
  ORDER BY name
$$) AS (name agtype, friendCount agtype);
```

Expected and observed result:
```text
Alice, 2
Bob, 0
Charlie, 0
```

Control case 2, with `WHERE false`, also preserves null rows correctly:

```pgsql
  MATCH (p:Person)
  OPTIONAL MATCH (p)-[:KNOWS]->(friend:Person)
  WHERE false
  RETURN p.name AS name, friend.name AS friend
  ORDER BY name
$$) AS (name agtype, friend agtype);
```

Expected and observed result:
```text
Alice, null
Bob, null
Charlie, null
```

A second variant in the same family also reproduces the issue with `COUNT { ... }`:

```pgsql
  MATCH (p:Person)
  OPTIONAL MATCH (p)-[:FRIEND]->(f)
  WHERE COUNT { MATCH (p)-[:FRIEND*..2]->(x) RETURN x } > 1
  RETURN p.name AS person, f.name AS friend
  ORDER BY person, friend
$$) AS (person agtype, friend agtype);
```

gregfelice:
## Summary

Fixes issue #2378 — `OPTIONAL MATCH` incorrectly drops null-preserving outer rows when its `WHERE` clause contains a correlated sub-pattern predicate (`EXISTS { ... }`, `COUNT { ... }`).

## Reproducer

```cypher
CREATE (a:Person {name:'Alice'}), (b:Person {name:'Bob'}), (c:Person {name:'Charlie'}),
       (a)-[:KNOWS]->(b), (a)-[:KNOWS]->(c);

MATCH (p:Person)
OPTIONAL MATCH (p)-[:KNOWS]->(friend:Person)
WHERE EXISTS { (friend)-[:KNOWS]->(:Person) }
RETURN p.name, friend.name;
```

**Before the fix:** `(0 rows)`
**After the fix:** `Alice | NULL`, `Bob | NULL`, `Charlie | NULL` — one row per person, null-preserved, matching Cypher OPTIONAL MATCH semantics.

## Root cause

In `transform_cypher_match`, when the `WHERE` clause contains a list comprehension or a sub-pattern (`has_list_comp_or_subquery` returns true), the code takes a rewrite path that:

1. Detaches the `WHERE` from the `cypher_match` node (`match_self->where = NULL`).
2. Transforms the match clause as a subquery via `transform_cypher_clause_as_subquery`.
3. Attaches the detached `WHERE` as an outer filter on that subquery: `query->jointree = makeFromExpr(p_joinlist, where_qual)`.

This rewrite is correct for non-optional `MATCH`, because the subquery's output is the full set of matches and the outer `WHERE` filters that set.

But for `OPTIONAL MATCH`, the inner subquery produces a `LATERAL LEFT JOIN` that already emits null-preserving rows for outer tuples with no right-hand match. The outer `WHERE` filter then runs against those nulled rows and drops them when the predicate evaluates to `NULL` or `false` on the nulled side. The result is zero rows for every outer tuple whose optional matches all fail the predicate — a direct violation of Cypher `OPTIONAL MATCH` semantics, which require the outer row to survive with `NULL` in the optional columns.

## Fix

Two coordinated changes in `src/backend/parser/cypher_clause.c`:

1. **`transform_cypher_match`** — only take the `has_list_comp_or_subquery` rewrite path for non-optional MATCH. For OPTIONAL MATCH, fall through to the normal `transform_cypher_match_pattern` path so the WHERE remains attached to the `cypher_match` node.

2. **`transform_cypher_optional_match_clause`** — detach `match_self->where` before recursively transforming the right-hand side of the `LATERAL LEFT JOIN`, so the inner transform does not double-apply or misresolve the predicate in a fresh namespace. After both sides are transformed and the right-side namespace item is in `pstate->p_namespace`, re-attach the transformed predicate as the `LEFT JOIN`'s `ON` condition (`j->quals`). PostgreSQL's `LEFT JOIN ... ON <pred>` correctly preserves left rows with null right columns when the `ON` fails — which is exactly the semantics Cypher `OPTIONAL MATCH ... WHERE` requires.

The fix restores the invariant that for every outer row, `OPTIONAL MATCH ... WHERE pred` either emits the surviving right-hand matches (if any) or exactly one row with nulls in the optional columns.

## Regression tests

Added to `regress/sql/cypher_match.sql` under a new `issue_2378` graph:

- Correlated `EXISTS` referencing the optional variable (`friend`) — the primary reproducer.
- Correlated `EXISTS` referencing the outer variable (`p`) — the second bug family.
- Non-correlated `EXISTS` — was already working; kept as a regression guard.
- Plain scalar predicate on the optional variable (`friend.name = 'Bob'`) — was already working; guard.
- Constant-false `WHERE` — was already working; guard.

## Test plan

- [x] `make installcheck REGRESS=\"cypher_match cypher_merge cypher_with cypher_subquery cypher_vle cypher_union cypher_call cypher_create cypher_set cypher_delete cypher_remove cypher_unwind list_comprehension expr\"` — all 14 suites pass.
- [x] Issue reporter's primary reproducer now returns the expected 3 rows with null-preserved friend.
- [x] Second variant from the issue body (EXISTS referencing outer `p`) now preserves outer rows correctly.
- [x] Non-correlated EXISTS, plain predicates, and `WHERE false` continue to work as before.

### DIFF
--- a/regress/expected/cypher_match.out
+++ b/regress/expected/cypher_match.out
@@ -3835,6 +3835,130 @@ NOTICE:  graph "issue_2193" has been dropped
 (1 row)
 
 --
+-- Issue 2378: OPTIONAL MATCH may incorrectly drop null-preserving outer
+-- rows when its WHERE clause contains a correlated sub-pattern predicate.
+--
+-- Cypher OPTIONAL MATCH semantics: the WHERE applies to the optional
+-- binding; when no right-hand row survives the predicate, the outer row
+-- is still emitted with NULLs in the optional columns.  Before the fix,
+-- a WHERE containing EXISTS { ... } or COUNT { ... } was attached as an
+-- outer filter on the transformed subquery, so it ran after the LATERAL
+-- LEFT JOIN produced null-preserving rows and then incorrectly dropped
+-- them when the predicate evaluated NULL/false on the nulled side.
+--
+SELECT create_graph('issue_2378');
+NOTICE:  graph "issue_2378" has been created
+ create_graph 
+--------------
+ 
+(1 row)
+
+SELECT * FROM cypher('issue_2378', $$
+    CREATE (a:Person {name: 'Alice'}),
+           (b:Person {name: 'Bob'}),
+           (c:Person {name: 'Charlie'}),
+           (a)-[:KNOWS]->(b),
+           (a)-[:KNOWS]->(c)
+$$) AS (v agtype);
+ v 
+---
+(0 rows)
+
+-- Correlated EXISTS referencing the optional variable (friend).
+-- Neither Bob nor Charlie knows anyone, so for every outer p the
+-- predicate fails on all optional matches; expect one row per person
+-- with friend = NULL.
+SELECT * FROM cypher('issue_2378', $$
+    MATCH (p:Person)
+    OPTIONAL MATCH (p)-[:KNOWS]->(friend:Person)
+    WHERE EXISTS { (friend)-[:KNOWS]->(:Person) }
+    RETURN p.name AS name, friend.name AS friend
+    ORDER BY name
+$$) AS (name agtype, friend agtype);
+   name    | friend 
+-----------+--------
+ "Alice"   | 
+ "Bob"     | 
+ "Charlie" | 
+(3 rows)
+
+-- Correlated EXISTS referencing the outer variable (p).
+-- Alice knows someone so her optional matches pass; Bob and Charlie
+-- don't, so they are emitted with NULL friend.
+SELECT * FROM cypher('issue_2378', $$
+    MATCH (p:Person)
+    OPTIONAL MATCH (p)-[:KNOWS]->(friend:Person)
+    WHERE EXISTS { (p)-[:KNOWS]->(:Person) }
+    RETURN p.name AS name, friend.name AS friend
+    ORDER BY name, friend
+$$) AS (name agtype, friend agtype);
+   name    |  friend   
+-----------+-----------
+ "Alice"   | "Bob"
+ "Alice"   | "Charlie"
+ "Bob"     | 
+ "Charlie" | 
+(4 rows)
+
+-- Non-correlated EXISTS (was already working; kept as a regression guard).
+SELECT * FROM cypher('issue_2378', $$
+    MATCH (p:Person)
+    OPTIONAL MATCH (p)-[:KNOWS]->(friend:Person)
+    WHERE EXISTS { MATCH (x:Person) RETURN x }
+    RETURN p.name AS name, friend.name AS friend
+    ORDER BY name, friend
+$$) AS (name agtype, friend agtype);
+   name    |  friend   
+-----------+-----------
+ "Alice"   | "Bob"
+ "Alice"   | "Charlie"
+ "Bob"     | 
+ "Charlie" | 
+(4 rows)
+
+-- Plain scalar predicate on the optional variable (was already working).
+SELECT * FROM cypher('issue_2378', $$
+    MATCH (p:Person)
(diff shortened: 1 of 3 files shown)


MuhammadTahaNaveed:
@gregfelice build is failing

gregfelice:
Ok on it

gregfelice:
Pushed a one-line fix (bc891bfb).

**Root cause:** In `transform_cypher_optional_match_clause` I called `transform_cypher_expr` with `EXPR_KIND_JOIN_ON` when re-attaching the detached WHERE as the LEFT JOIN's ON condition. All other WHERE transforms in `cypher_clause.c` use `EXPR_KIND_WHERE`, and there are three explicit `p_expr_kind == EXPR_KIND_WHERE` guards (`cypher_clause.c:5415`, `5679`, `6597`) that do load-bearing variable resolution for sub-pattern predicates — walking up parent parsestates to rebind variables like `friend` inside `EXISTS { (friend)-[...]->(...) }`. `EXPR_KIND_JOIN_ON` bypassed those guards, so the sub-pattern fell through to the "create new variable" path and produced a structurally invalid parse tree — silently tolerated on a release PG build but caught by `--enable-cassert`.

**Why I didn't catch it:** my local PG18 is the stock Ubuntu package (no `--enable-cassert`), so all 32 regression tests passed locally before I opened the PR. Setting up a dedicated `--enable-cassert` PG build locally now so this class of failure can't slip past me again. Thanks for flagging @MuhammadTahaNaveed.

MuhammadTahaNaveed (approved):

PR #2397, opened 2026-04-16
Fix upgrade test: build default install SQL from HEAD, not initial commit

jrgemignani:
…mmit

The upgrade test previously built age--<CURR>.sql from the initial version-bump commit, meaning CREATE EXTENSION installed 'day-one' SQL. This caused all 31 non-upgrade regression tests to run WITHOUT SQL functions added after the version bump (e.g., age_invalidate_graph_cache, age_prepare_pg_upgrade, age_vertex_stats, etc.). These functions were never registered in pg_proc, so features depending on them (like VLE cache invalidation triggers) were silently disabled during testing.

Fix by inverting the upgrade test direction:

Before (installs incomplete SQL, upgrades to complete):
  age--1.7.0.sql                     built from version-bump commit (incomplete)
  age--1.7.0_upgrade_test.sql        built from HEAD (complete)
  age--1.7.0--1.7.0_upgrade_test.sql stamped from template
  Test: CREATE EXTENSION age -> data -> ALTER EXTENSION UPDATE TO '1.7.0_upgrade_test'

After (installs complete SQL, upgrades from synthetic initial):
  age--1.7.0.sql                     built from HEAD (complete)
  age--1.7.0_initial.sql             built from version-bump commit (synthetic)
  age--1.7.0_initial--1.7.0.sql      stamped from template
  Test: CREATE EXTENSION age VERSION '1.7.0_initial' -> data -> ALTER EXTENSION UPDATE TO '1.7.0'

This ensures:
  - All 31 non-upgrade tests run with every SQL function registered
  - The upgrade template is still validated (initial -> current)
  - The test ends at the default version (clean state for drop test)
  - Tarball builds (no git) work identically (cat sql/sql_files)
  - All synthetic files are cleaned up after the test

Makefile changes:
  - age_sql rule: cat current HEAD sql/sql_files (was: git show from commit)
  - New age_init_sql rule: git show from version-bump commit
  - age_upgrade_test_sql: stamps template as INIT->CURR (was CURR->NEXT)
  - EXTRA_CLEAN, _install_upgrade_test_files: updated filenames
  - Removed AGE_NEXT_VER, age_next_sql; added AGE_INIT_VER, age_init_sql

Test SQL changes:
  - Step 3: CREATE EXTENSION age VERSION '<init>' (was: CREATE EXTENSION age)
  - Step 6: ALTER EXTENSION age UPDATE TO default_version (was: LIKE '%_upgrade_test')
  - Step 7: Check installed = default (was: installed <> default)
[...]

PR #2403, opened 2026-04-21
Fix upgrade test

jrgemignani:
The age_upgrade regression test (added in #2364, improved in #2377, #2397)
was designed to validate the upgrade template (age--<VER>--y.y.y.sql) by
creating graph data before the upgrade and verifying it survived afterward.
This approach had two fundamental problems:

1. It did not detect incomplete upgrade templates. The test verified that
   graph data (vertices, edges, checksums, GIN indexes) survived ALTER
   EXTENSION UPDATE, but never checked whether new SQL objects (functions,
   views, relations, indexes, types, operators, casts, constraints) were
   actually created by the template. A developer could add a new function
   to sql/ and sql_files, forget to add it to the upgrade template, and
   all tests would pass — the function existed via the fresh CREATE
   EXTENSION install that ran before the upgrade test, but would be
   missing for users who upgraded via ALTER EXTENSION UPDATE.

2. The data-integrity checks relied on cypher queries (MATCH/RETURN) within
   the same backend session after DROP EXTENSION + CREATE EXTENSION. This
   caused intermittent failures on some PostgreSQL versions where AGE's
   internal type cache (agtype OID) was not properly refreshed after the
   extension was dropped and recreated, resulting in 'type with OID 0
   does not exist' errors. The data-integrity aspect was also redundant —
   ALTER EXTENSION UPDATE runs DDL statements and does not touch heap data,
   so data survival is guaranteed by PostgreSQL and not a meaningful test.

The fix replaces the entire test with a comprehensive catalog comparison:

  1. Snapshot the ag_catalog schema from the fresh install across seven
     PostgreSQL system catalogs:
       - pg_proc: functions, aggregates, procedures (name, args, and
         properties: volatility, strictness, kind, return type, setof)
       - pg_class: tables, views, sequences, indexes (name, kind)
       - pg_type: types (name, type category)
       - pg_operator: operators (name, left/right operand types)
       - pg_cast: casts involving AGE types (source, target, context)
       - pg_opclass: operator classes (name, access method)
       - pg_constraint: constraints (name, type, table, referenced table)
  2. DROP EXTENSION, CREATE EXTENSION at the synthetic initial version,
     then ALTER EXTENSION UPDATE to the current version via the stamped
     upgrade template.
  3. Snapshot the catalog again after upgrade.
  4. Compare: any object present in the fresh snapshot but missing after
     upgrade means the template is incomplete. Any object present after
     upgrade but not in the fresh snapshot means the template creates
     something unexpected. Function properties (volatility, strictness,
     prokind, return type) are also compared for functions that exist in
     both — catching cases where a CREATE OR REPLACE in the template
     changes a function's signature or behavior.
     
 Additional improvements from code review feedback:

  - Graph cleanup in Step 1 uses a DO block with PERFORM and suppressed
    NOTICEs to produce deterministic output regardless of prior test state.
  - The pg_class snapshot includes indexes (relkind 'i') in addition to
    tables, views, and sequences.
  - Diagnostic output includes relkind/typtype suffixes for actionable diffs.
  - Summary uses boolean equality checks (funcs_match, rels_match, etc.)
    instead of absolute counts, so the expected output does not need
    updating when new objects are added to AGE. Developers who correctly
    add objects to both sql/ and the template will never need to modify
    this test or its expected output.

This approach:
  - Catches the actual failure mode: incomplete upgrade templates.
  - Covers all SQL object categories: functions (including aggregates),
    relations, types, operators, casts, operator classes, and constraints.
  - Detects property changes on existing functions (volatility, strictness,
    kind, return type changes).
  - Uses only plain SQL catalog queries — no cypher, no .so cache issues.
  - Works reliably across all PostgreSQL versions.
  - Reports the exact missing/extra/changed object in the diff output.
  - Is maintenance-free: no expected output changes needed when AGE grows.

Makefile: updated step 5 comment to reflect catalog comparison approach.

All 33 regression tests pass.

modified:   Makefile
modified:   regress/expected/age_upgrade.out
modified:   regress/sql/age_upgrade.sql

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,15 +33,17 @@ age_sql = age--1.7.0.sql
 #   4. Temporarily installs the synthetic files into the PG extension directory
 #      so that CREATE EXTENSION age VERSION '<INIT>' and ALTER EXTENSION
 #      age UPDATE TO '<CURR>' can find them.
-#   5. The age_upgrade regression test exercises the full upgrade path: install
-#      at INIT, create data, ALTER EXTENSION UPDATE to CURR, verify data.
+#   5. The age_upgrade regression test snapshots the ag_catalog schema from
+#      a fresh install, then installs at INIT, upgrades to CURR, and compares
+#      the catalog across seven system catalogs to detect missing, extra,
+#      or changed objects.
 #   6. The test SQL cleans up the synthetic files via a generated shell script.
 #
 # This forces developers to keep the upgrade template in sync: any SQL object
 # added after the version-bump commit must also appear in the template, or the
 # upgrade test will fail (the object will be missing after ALTER EXTENSION UPDATE).
 #
-# Because the default install SQL comes from HEAD, all 31 non-upgrade tests
+# Because the default install SQL comes from HEAD, all non-upgrade tests
 # run with every SQL function registered — no functions are missing.
 #
 # Graceful degradation — the upgrade test is silently skipped when:
--- a/regress/expected/age_upgrade.out
+++ b/regress/expected/age_upgrade.out
@@ -17,36 +17,105 @@
  * under the License.
  */
 --
--- Extension upgrade regression test
+-- Extension upgrade template regression test
 --
--- This test validates the upgrade template (age--<VER>--y.y.y.sql) by:
---   1. Dropping AGE and reinstalling at the synthetic "initial" version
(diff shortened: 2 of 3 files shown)


jrgemignani:
@MuhammadTahaNaveed rebased against latest master.

MuhammadTahaNaveed (approved):

Issue #2383, opened 2026-04-16
`UNWIND [null]` may produce a row whose value prints as `null`, but `count(x)` still counts it as `1`.


Silence6666668:
**Describe the bug**
`UNWIND [null]` may produce a row whose value prints as `null`, but `count(x)` still counts it as `1`.

In Cypher, `count(expr)` should count only non-null values. After `UNWIND [null] AS x`, the single row has `x = null`, so `count(x)` should be `0`.

Instead, Apache AGE returns `1`.

**How are you accessing AGE (Command line, driver, etc.)?**
- PostgreSQL `cypher(...)` wrapper through the local Python differential-testing harness
- Reproducible directly in `psql` inside the Docker container

**What data setup do we need to do?**
No graph data is required beyond creating an empty graph:

```pgsql
SELECT create_graph('fuzz_graph');
```

**What is the necessary configuration info needed?**
- Plain Apache AGE Docker image was enough
- Docker image in local repro: `apache/age`
- AGE extension version: `1.7.0`
- PostgreSQL version: `18.1`
- Graph name used in repro: `fuzz_graph`
- No extra extensions or special configuration were required

**What is the command that caused the error?**
```pgsql
SELECT * FROM cypher('fuzz_graph', $$
  UNWIND [null] AS x
  RETURN count(x) AS c
$$) AS (c agtype);
```

Returned result on AGE:
```text
c
---
1
```

**Expected behavior**
The result should be:

```text
c
---
0
```

Neo4j returns `0` for the equivalent Cypher query:

```cypher
UNWIND [null] AS x
RETURN count(x) AS c
```

**Environment (please complete the following information):**
- Version: Apache AGE `1.7.0`
- PostgreSQL: `18.1`
- Host OS: Windows
- Architecture: x86_64
- Deployment: Docker

**Additional context**
This appears to be a counting/null-semantics issue specifically around `UNWIND`, not a general `count(null)` problem.

Direct control cases on the same AGE instance:

1. Plain `count(null)` behaves correctly:

```pgsql
SELECT * FROM cypher('fuzz_graph', $$
  RETURN count(null) AS c
$$) AS (c agtype);
```

Observed result:
```text
0
```

2. `WITH null AS x RETURN count(x)` also behaves correctly:

```pgsql
SELECT * FROM cypher('fuzz_graph', $$
  WITH null AS x
  RETURN count(x) AS c
$$) AS (c agtype);
```

Observed result:
```text
0
```

[...]

PR #2406, opened 2026-04-22
Propagate null through unnest and single() three-valued logic

### LINKED ISSUE #2393: List-comprehension `WHERE` filters may mishandle `null` elements.
**Describe the bug**
List-comprehension `WHERE` filters may mishandle `null` elements.

In the repro below, Apache AGE drops `null` elements even when the filter explicitly selects them with `x IS NULL`. It also fails to remove `null` elements when the filter is `x IS NOT NULL`.

Both Neo4j and Memgraph return the expected filtered lists.

No graph data is required beyond creating an empty graph:

```pgsql
SELECT create_graph('fuzz_graph');
```

- PostgreSQL version: `18.1`

```pgsql
  RETURN [x IN [null,1] WHERE x IS NULL] AS v
$$) AS (v agtype);
```

Returned result on AGE:
```text
[]
```

**Expected behavior**
The `WHERE x IS NULL` filter should keep the null element:

```text
[null]
```

Neo4j returns `[null]`, and Memgraph returns `[null]`.

- Version: Apache AGE `1.7.0`
- PostgreSQL: `18.1`
- Host OS: Windows
- Architecture: x86_64
- Deployment: Docker

**Additional context**
The same issue appears more clearly with a second null element:

```pgsql
  RETURN [x IN [null,1,null] WHERE x IS NULL] AS v
$$) AS (v agtype);
```

Apache AGE returns:
```text
[]
```

Expected result:
```text
[null, null]
```

Neo4j and Memgraph both return `[null, null]`.

The opposite filter is also wrong on AGE:

```pgsql
  RETURN [x IN [null,1] WHERE x IS NOT NULL] AS v
$$) AS (v agtype);
```

Apache AGE returns:
```text
[null, 1]
```

Expected result:
```text
[1]
```

Neo4j and Memgraph both return `[1]`.

For reference, the plain list comprehension without a filter behaves as expected:

```pgsql
  RETURN [x IN [null,1] | x] AS v
$$) AS (v agtype);
```

Apache AGE returns:
```text
[null, 1]
```

So the issue appears to be specifically in the `WHERE` filtering phase of list comprehension evaluation when `null` elements are present.

MuhammadTahaNaveed:
Two related defects in how AGE handled cypher null inside list-iterating constructs.

age_unnest packaged every iterated element as a non-SQL-NULL agtype datum, even AGTV_NULL scalars. SQL `IS NULL` / `IS NOT NULL` then couldn't see those nulls, so `[x IN [null, 1] WHERE x IS NULL]` dropped the null it was meant to keep, and `WHERE x IS NOT NULL` kept the null it was meant to drop. The same mismatch surfaced in UNWIND. AGE already treats SQL NULL as the row-level representation of cypher null elsewhere (`RETURN null AS v` yields SQL NULL, strict operators short-circuit on it); age_unnest now does the same by emitting the row with `nulls[0] = true` when the element is AGTV_NULL.

single() previously transformed to `SELECT count(*) FROM unnest(list) AS x WHERE pred IS TRUE`, with the grammar wrapping the result as `(subquery) = 1`. With the unnest fix, `[null, 5] WHERE x > 0` left one definite true after the WHERE filter -> count = 1 -> true. Neo4j returns null because the unknown predicate could itself be a second match. Rewritten to a CASE built on `count(*) FILTER (WHERE pred IS TRUE)` and `bool_or(pred IS NULL)`:

  CASE WHEN count(*) FILTER (WHERE pred IS TRUE) >= 2 THEN false
       WHEN bool_or(pred IS NULL)                     THEN NULL
       WHEN count(*) FILTER (WHERE pred IS TRUE) =  1 THEN true
       ELSE false END

The >=2 arm runs first so two definite trues dominate any unknowns. Fits inside the existing make_predicate_case_expr helper alongside all/any/none, removes the special-case transform branch and the grammar `= 1` wrap. A small `make_count_star_filter_agg` helper mirrors the existing `make_bool_or_agg`. Verified against Neo4j for the new edge cases (one-true-plus-null, two-trues-plus-null, all-nulls, mixed-true-false-null).

The predicate_functions regression also picks up the corrected behavior of any/all/none over null elements: `null > 0` now yields SQL NULL instead of being silently treated as true, so the three-valued combinators in those functions produce the openCypher results the comments previously documented as buggy.

### DIFF
--- a/regress/expected/list_comprehension.out
+++ b/regress/expected/list_comprehension.out
@@ -809,6 +809,39 @@ SELECT * FROM cypher('list_comprehension', $$ RETURN [x IN [1, null, 2] | -x] $$
  [-1, null, -2]
 (1 row)
 
+-- Issue 2393 - WHERE filter over null elements should use openCypher's
+-- three-valued logic: IS NULL must keep nulls, IS NOT NULL must drop them.
+SELECT * FROM cypher('list_comprehension', $$ RETURN [x IN [null, 1] WHERE x IS NULL] $$) AS (result agtype);
+ result 
+--------
+ [null]
+(1 row)
+
+SELECT * FROM cypher('list_comprehension', $$ RETURN [x IN [null, 1, null] WHERE x IS NULL] $$) AS (result agtype);
+    result    
+--------------
+ [null, null]
+(1 row)
+
+SELECT * FROM cypher('list_comprehension', $$ RETURN [x IN [null, 1] WHERE x IS NOT NULL] $$) AS (result agtype);
+ result 
+--------
+ [1]
+(1 row)
+
+SELECT * FROM cypher('list_comprehension', $$ RETURN [x IN [1, 2, 3] WHERE x IS NULL] $$) AS (result agtype);
+ result 
+--------
+ []
+(1 row)
+
+SELECT * FROM cypher('list_comprehension', $$ UNWIND [null, 1] AS x RETURN x, x IS NULL, x IS NOT NULL $$) AS (x agtype, a agtype, b agtype);
+ x |   a   |   b   
+---+-------+-------
+   | true  | false
+ 1 | false | true
+(2 rows)
+
 -- Clean up
 SELECT * FROM drop_graph('list_comprehension', true);
 NOTICE:  drop cascades to 4 other objects
--- a/regress/expected/predicate_functions.out
+++ b/regress/expected/predicate_functions.out
@@ -194,20 +194,20 @@ $$) AS (result agtype);
 --
 -- NULL predicate results: three-valued logic
 --
--- Note: In AGE's agtype, null is a first-class value.  The comparison
--- agtype_null > agtype_integer evaluates to true (not SQL NULL).
--- Three-valued logic only applies when the predicate itself is a
--- literal null constant, which becomes SQL NULL after coercion.
--- agtype null in list: null > 0 = true in AGE, so any() = true
+-- Null list elements arrive at the predicate as SQL NULL, so the usual
+-- strict-operator short-circuit applies: `null > 0` yields NULL, and the
+-- predicate functions combine NULLs with the openCypher three-valued
+-- logic (true trumps null in any(), false trumps null in all(), etc.).
+-- [null]: only null predicate, no true -> any() = NULL
 SELECT * FROM cypher('predicate_functions', $$
     RETURN any(x IN [null] WHERE x > 0)
 $$) AS (result agtype);
  result 
 --------
- true
+ 
 (1 row)
 
--- agtype null + real values: all comparisons are true
+-- one true (1 > 0) is enough: any() = true
 SELECT * FROM cypher('predicate_functions', $$
     RETURN any(x IN [null, 1, 2] WHERE x > 0)
 $$) AS (result agtype);
@@ -226,16 +226,16 @@ $$) AS (result agtype);
  
 (1 row)
 
--- agtype null in list: null > 0 = true in AGE, so all() = true
+-- no false, but one null -> all() = NULL
 SELECT * FROM cypher('predicate_functions', $$
     RETURN all(x IN [1, null, 2] WHERE x > 0)
 $$) AS (result agtype);
  result 
 --------
- true
+ 
 (1 row)
 
--- -1 > 0 = false, so all() = false
+-- -1 > 0 = false dominates the null -> all() = false
 SELECT * FROM cypher('predicate_functions', $$
     RETURN all(x IN [1, null, -1] WHERE x > 0)
 $$) AS (result agtype);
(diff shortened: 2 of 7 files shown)


jrgemignani:
@MuhammadTahaNaveed It looks like you need to rebase this PR

gregfelice (commented):
Reviewed the `age_unnest` change, the `single()` CASE logic, and the new aggregate helpers. The approach is sound and I'd endorse the direction; a few items below before it's approve-ready.

**Approach.** Emitting a SQL-NULL row (`nulls[0]=true`) for `AGTV_NULL` elements in `age_unnest` is the right fix — it makes the iterated element match how AGE represents cypher null everywhere else, so SQL `IS NULL`/`IS NOT NULL` and strict aggregates behave per openCypher. One change covers list-comprehension filters (#2393), `UNWIND [null]` count (#2383), and `single()` three-valued logic together.

**`single()` 3VL — verified correct.** I worked the cases by hand:
- ≥2 definite trues → false (and this arm is evaluated first, so definite trues correctly dominate any unknown/null elements);
- exactly 1 true with no nulls → true; exactly 1 true with a null → NULL (the unknown could be a second match);
- 0 trues with a null → NULL; all-false → false; empty list → false.

That matches Neo4j, including the empty-list and the one-true-plus-null edge cases.

**Overlap with #2411 (worth a maintainer decision).** Both PRs fix the same root cause in `age_unnest`. This PR changes `age_unnest` in place, so it fixes UNWIND **and** list comprehension **and** single() in one go; #2411 adds a separate UNWIND-specific function and fixes only the `UNWIND [null]` count case. They're mutually redundant — picking this one would subsume #2411.

**Minor (the three Copilot inline notes are valid):**
1. `makeConst(INT8OID, ..., FLOAT8PASSBYVAL)` should use `INT8PASSBYVAL`. It's functionally identical on all supported platforms (both follow `USE_FLOAT8_BYVAL`), but it's the wrong macro for an int8 constant — worth fixing for clarity/correctness.
2. The empty-list comment is inaccurate: `count(*)` (incl. `count(*) FILTER (...)`) returns 0, not NULL, on zero rows — only `bool_or(...)` returns NULL. The logic is correct; just the comment.
3. `make_count_star_filter_agg` hard-codes `INT8OID` as the return type — fine for `count(*)`, slightly more robust to derive it, but optional.

With the rebase @jrgemignani asked for plus the byval/comment tidy-ups, this looks good to me.

gregfelice:
@jrgemignani — a couple of cross-cutting notes from going through the open parser/agtype PRs, in case they help with sequencing:

**1. #2406 and #2411 are competing fixes for the same `UNWIND [null]` bug (#2383).**
They take opposite approaches and will conflict — whichever lands first forces a rework of the other:

- **#2411** is surgical: a separate `age_unwind` SRF, with only `transform_cypher_unwind` repointed to it. `age_unnest` is untouched, so list comprehensions and predicate functions can't regress. The trade-off is it *intentionally* leaves the related list-comprehension null-filter bug (#2393) and the predicate-function three-valued logic unfixed, since those callers currently rely on `agtype_null > agtype_integer` being truthy.
- **#2406** (this PR) is root-cause: it fixes `age_unnest` itself and rewrites the predicate-function 3VL so that truthy-null quirk becomes proper three-valued logic — fixing UNWIND, list comprehensions (#2393), and `single()` together.

Per openCypher, `null > 0` should be `null` rather than truthy, so this PR is the more correct end state; #2411 is the smaller, lower-risk step. Both are sound for their stated scope — it's really a scoping call on whether to fix the whole null-handling family at once or just the UNWIND symptom. Flagging so they don't both get merged independently.

**2. A few of these are blocked more on process than on code.**

- **No CI run:** #2411 and #2412 don't show any check runs — the contributor branches likely need a maintainer to kick off the workflow (or a push) before they can be evaluated for merge.
- **Need a rebase:** #2295 and #2411 are currently CONFLICTING with master. #2295 in particular dates to Dec 2025 and touches parser/agtype areas that have moved a lot since, so it'll want a rebase + fresh green CI before a real merge pass.

Happy to take another look at any of these once they're rebased / CI is green.

gregfelice:
@MuhammadTahaNaveed The approach here is solid and I'd like to get it in — it's currently blocked only on a rebase plus two small tidy-ups. If you can:

1. **Rebase onto latest `master`** (it's conflicting now, and @jrgemignani asked back on Apr 30) so CI runs green again.
2. **`INT8PASSBYVAL` fix** — the two `single()` `makeConst(INT8OID, …)` calls use `FLOAT8PASSBYVAL` for the byval flag; they should use `INT8PASSBYVAL`. Functionally identical on all supported platforms, but it's the correct macro for an int8 const.
3. **Empty-list comment** — the revised comment still says "all aggregates return NULL on zero rows"; `count(*)` (incl. `count(*) FILTER (…)`) returns `0`, not NULL. The logic is right, just the comment — worth a one-line correction.

(The `make_count_star_filter_agg` hardcoded `INT8OID` return is fine to leave as-is.)

Once that's done and CI's green, I'll approve. Thanks for working the full three-valued-logic fix rather than just the UNWIND symptom.

MuhammadTahaNaveed:
@gregfelice fixed.

gregfelice (approved):
All three items addressed — rebased onto master (mergeable, CI green), `INT8PASSBYVAL` applied to the `single()` consts (with a portability guard), and the empty-list comment corrected. Verified the three-valued logic for single()/any()/all()/none() over null elements against the openCypher/Neo4j semantics. LGTM, thanks @MuhammadTahaNaveed.

PR #2407, opened 2026-04-22
Critical: fix inverted logic here

sfc-gh-dachristensen:
strcmp(str, "") returns 0 (false) when str is empty, meaning the check is inverted: it returns NULL when parsing succeeds and continues when parsing fails. This allows non-numeric strings to pass through as array indices, leading to type confusion and potentially incorrect memory access.

The strcmp logic handles most cases correctly (non-numeric strings return NULL, valid integers pass through). However, the empty string "" is accepted as a valid array index of 0: [10, 20, 30] #> '[""]' returns 10 instead of NULL. This occurs because strtol("") sets lindex=0 and str="", so strcmp("", "") returns 0, bypassing the error check.

### DIFF
--- a/src/backend/utils/adt/agtype_ops.c
+++ b/src/backend/utils/adt/agtype_ops.c
@@ -2098,7 +2098,7 @@ static Datum get_agtype_path_all(FunctionCallInfo fcinfo, bool as_text)
                 char* str = NULL;
                 lindex = strtol(cur_key->val.string.val, &str, 10);
 
-                if (strcmp(str, ""))
+                if (strcmp(str, "") != 0)
                 {
                     PG_RETURN_NULL();
                 }


jrgemignani:
@sfc-gh-dachristensen Please fix the above Copilot issues.

jrgemignani (changes requested):
See copilot review

gregfelice:
@sfc-gh-dachristensen — I think this change is a no-op, and the title overstates it. Recommending we close unless I'm missing something.

The diff is:

```c
-                if (strcmp(str, ""))
+                if (strcmp(str, "") != 0)
```

These are identical in C. `strcmp()` returns `0` when the strings are equal and non-zero otherwise, and `if (x)` already tests `x != 0`. So both forms mean "if `str` is **not** empty" and compile to the same thing. There is no inverted logic here — the existing code is already correct.

The semantics of the surrounding block are also right as-is: `strtol()` sets `str` to the first unparsed character, so `strcmp(str, "")` being true means there was trailing garbage after the number (e.g. `"1abc"`), and the code correctly bails with `PG_RETURN_NULL()`. An empty leftover means the whole key parsed cleanly as an integer.

If the actual concern is the one Copilot describes — that an **empty key** `""` is accepted and silently treated as index `0`, because `strtol("")` returns 0 with `str` pointing at the terminator so the guard passes — then that is a real issue, but this diff does not fix it. That would need an explicit check that the key is non-empty (and ideally an `errno`/`ERANGE` check around `strtol`) before using `lindex`.

Happy to look at a revised PR that addresses the empty-key case with a regress test. As it stands I'd close this one.

sfc-gh-dachristensen:
@gregfelice on re-review, agreed this PR as currently written is nonsense—will close and review to see what the original report I was responding to was referencing.  This code block clearly intends to return NULL if we have any extra chars in the parsed string, which it does do properly.  Sorry for the noise.

PR #2408, opened 2026-04-22
Critical: check for Size overflow when using palloc

sfc-gh-dachristensen:
Multiple locations compute allocation sizes via unchecked integer multiplication:

Line 1027: palloc(sizeof(agtype_value) * (*pstate)->size) — size doubles repeatedly (lines 1122, 1169); near 0x40000000, multiplication overflows to a small value, leading to heap buffer overflow on subsequent writes. Line 1038: palloc(sizeof(agtype_pair) * (*pstate)->size) — same pattern. Lines 2267, 2366: sizeof(agtentry) * num_elems and sizeof(agtentry) * num_pairs * 2 — chained multiplication without overflow check. Line 1397: it->num_elems * sizeof(agtentry) * 2

Fix this by making explicit checks.

### DIFF
--- a/src/backend/utils/adt/agtype_util.c
+++ b/src/backend/utils/adt/agtype_util.c
@@ -1023,6 +1023,8 @@ static agtype_value *push_agtype_value_scalar(agtype_parse_state **pstate,
         {
             (*pstate)->size = 4;
         }
+        if ((*pstate)->size > SIZE_MAX / sizeof(agtype_value))
+            ereport(ERROR, "allocation request exceeds size limits");
         (*pstate)->cont_val.val.array.elems =
             palloc(sizeof(agtype_value) * (*pstate)->size);
         (*pstate)->last_updated_value = NULL;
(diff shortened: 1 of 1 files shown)


jrgemignani:
@sfc-gh-dachristensen I'm going to reject and close this PR as it isn't necessary -

Below is from Opus 4.7, but I have also reviewed the work. Additionally, it doesn't follow our coding standards: all blocks are always wrapped with braces. While not all of our code is this way, originally, all new code is.

## Summary

**Verdict: the PR is misdiagnosed and the patch is broken. Should be rejected as written.**

The PR claims integer-overflow → heap-buffer-overflow via palloc size-multiplication in `agtype_util.c`. After checking each cited site against the source, the conclusion is:

1. **The vulnerabilities described are not reachable** — pre-existing guards block them.
2. **The patch as posted does not compile.**
3. **Some checks the patch adds are dead code.**
4. **Several sites the description mentions are not actually patched in the diff.**
5. **Where the patch is correct, it is hardening, not a fix.**

Issue #2349, opened 2026-03-01
Docs: clarify VLE semantics and cost model


gregfelice:
## Summary

The variable-length edge (VLE) expansion algorithm used for patterns like `[*1..3]` exhibits O(n^k) time complexity because it does not perform cycle detection or deduplication during recursive expansion. On graphs with cycles (which are common in real-world data), this causes exponential path explosion.

## Reproduction

Setup: 100K nodes with edges forming natural cycles.

```sql
-- VLE query: find all nodes reachable within 1-3 hops
SELECT * FROM cypher('benchmark', $$
  MATCH (n:Node {bench_id: 42})-[*1..3]->(m)
  RETURN count(m)
$$) AS (cnt agtype);
```

On graphs with cycles, the expansion visits the same nodes exponentially many times because `UNION ALL` semantics are used internally (no deduplication).

## Performance Data

Benchmark W06 (variable-length edge traversal):

| Approach | 10K scale | 100K scale |
|---|---|---|
| Cypher VLE (before SDK workaround) | ~23ms | ~23ms |
| Recursive CTE with UNION (dedup) | 0.08ms | 0.34ms |

The SDK workaround uses a recursive CTE with `UNION` (not `UNION ALL`) which provides implicit cycle detection through set deduplication, preventing exponential blowup:

```sql
WITH RECURSIVE vle AS (
  SELECT e.end_id AS node_id, 1 AS depth
  FROM benchmark."EDGE" e
  WHERE e.start_id = (SELECT id FROM benchmark."Node" WHERE _bench_id = 42)
  UNION  -- implicit dedup = cycle detection
  SELECT e.end_id, v.depth + 1
  FROM vle v
  JOIN benchmark."EDGE" e ON e.start_id = v.node_id
  WHERE v.depth < 3
)
SELECT count(*) FROM vle WHERE depth >= 1;
```

## Suggested Fix

Add a visited-node set to the VLE expansion in the executor. When expanding `[*min..max]`, track which `(node_id, depth)` pairs have already been explored and skip duplicates. This is equivalent to using `UNION` instead of `UNION ALL` in the recursive expansion.

## Environment

- PostgreSQL 18.2
- Apache AGE 1.7.0 (built from source, branch `release/PG18/1.7.0`)
- 32-core, 64GB RAM, Linux 6.17

jrgemignani:
@gregfelice The following is from Opus, because it can explain it better than I, as it has been a while since I wrote the VLE.

[...]

PR #2409, opened 2026-04-22
Critical: add recursive limit to prevent resource exhaustion

sfc-gh-dachristensen:
A Cypher query like MATCH (a)-[*]->(b) RETURN * explores all paths of any length, consuming unbounded memory and CPU. In VLE_FUNCTION_PATHS_ALL mode, every possible path is materialized. The DFS loops (dfs_find_a_path_between, dfs_find_a_path_from) also lack CHECK_FOR_INTERRUPTS(), making runaway queries uncancellable. When uidx_infinite is true, traversal on cyclic graphs runs indefinitely. Unbounded stack growth via push_graphid_stack() calls (lines 1006, 1134) can consume all available process memory.

In order to address this issue we add a GUC to limit the resources.  We also add CHECK_FOR_INTERRUPTS() to the call to allow the cancelation of a high-cycle graph computation.

### DIFF
--- a/src/backend/utils/adt/age_vle.c
+++ b/src/backend/utils/adt/age_vle.c
@@ -18,6 +18,7 @@
  */
 
 #include "postgres.h"
+#include "miscadmin.h"
 
 #include "common/hashfn.h"
 #include "funcapi.h"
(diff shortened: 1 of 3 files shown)


jrgemignani:
@sfc-gh-dachristensen Please see https://github.com/apache/age/issues/2349

PR #2410, opened 2026-04-22
Critical: fix stack overflow from unbounded sprintf()

sfc-gh-dachristensen:
The %f format specifier for doubles can produce over 300 characters (e.g., DBL_MAX formatted with %f). The 64-byte stack buffer is insufficient. This is a stack buffer overflow exploitable via crafted agtype float values.

Fix this by both bumping the buffer size (to allow full precision) and changing all sprintf() call sites to be snprintf() calls instead.

### DIFF
--- a/src/backend/utils/adt/agtype.c
+++ b/src/backend/utils/adt/agtype.c
@@ -6556,7 +6556,7 @@ Datum age_tofloatlist(PG_FUNCTION_ARGS)
     int i;
     bool is_valid = false;
     float8 float_num;
-    char buffer[64];
+    char buffer[400];
 
     /* check for null */
     if (PG_ARGISNULL(0))
@@ -6618,7 +6618,7 @@ Datum age_tofloatlist(PG_FUNCTION_ARGS)
 
             float_elem.type = AGTV_FLOAT;
             float_num = elem->val.float_value;
-            sprintf(buffer, "%f", float_num);
+            snprintf(buffer, sizeof(buffer), "%f", float_num);
             string = buffer;
             float_elem.val.float_value = float8in_internal_null(string, NULL, "double precision", string, &is_valid);
             agis_result.res = push_agtype_value(&agis_result.parse_state, WAGT_ELEM, &float_elem);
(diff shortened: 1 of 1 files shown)


jrgemignani:
@sfc-gh-dachristensen Please add the following -

- Add a regression test - without this, a future refactor could reintroduce the bug.
- The value 400 seems a bit arbitrary. Why not some constant or number that feels less like a magic number? Fx, 512.
- Add a comment at the buffer declaration explaining the need.

gregfelice:
Confirming the vulnerability is real and query-reachable: at `age_tofloatlist` (`agtype.c`, the `AGTV_FLOAT` case), `sprintf(buffer, "%f", float_num)` formats a double's full integer part into a 64-byte stack buffer — `±DBL_MAX` is ~309 integer digits + `.000000` ≈ 317 chars, so `RETURN toFloatList([1.0e308])` overflows it. The `snprintf` + larger buffer fixes the overflow.

One suggestion that I think is strictly better than enlarging the buffer, and it also resolves @jrgemignani's "400 is a magic number" point by removing the buffer from that path entirely:

The value being formatted is **already** a `float8` (`elem->val.float_value`). The current code formats it with `%f` and re-parses the string back into a float:

```c
float_num = elem->val.float_value;
sprintf(buffer, "%f", float_num);                 // %f = 6 fractional digits → lossy
string = buffer;
float_elem.val.float_value = float8in_internal_null(string, ...);
```

Besides the overflow, `%f` truncates to 6 decimal places, so this round-trip is **lossy** — `toFloatList([0.123456789])` currently returns `0.123457`. Since the value is already a double, the format-and-reparse is unnecessary; assigning directly:

```c
float_elem.type = AGTV_FLOAT;
float_elem.val.float_value = elem->val.float_value;
```

fixes the stack overflow, the precision loss, and the magic-buffer-size question in one move (no buffer, no `sprintf`/`snprintf`, no re-parse). The `snprintf` conversions in `age_tostringlist` are still worth keeping as defensive hardening (those use bounded `%.*g`/`%ld` formats, so they were never overflow-prone — the 64-byte buffer there is fine).

If you'd prefer to keep the minimal critical-hotfix shape (`snprintf` + bigger buffer) and address the precision round-trip separately, that's reasonable too — just flagging that the direct assignment is the smaller and more correct end state.

gregfelice:
Thanks for catching and reporting this, @sfc-gh-dachristensen — the stack overflow is real and query-reachable, exactly as you identified.

We've opened #2451 to land the fix for the 1.8.0 release. It takes a slightly different end state: since the element is already a `float8`, it drops the `sprintf`/reparse entirely and assigns the value directly. That resolves the overflow **and** a latent precision-loss bug (`"%f"` truncates to 6 fractional digits), and removes the need for a fixed buffer at all. It also adds regression coverage, which was the main thing blocking this from merging as-is.

Your original report and fix are credited via `Co-authored-by` on that PR. Closing this one as superseded by #2451 — thank you again for the find.

Issue #2355, opened 2026-03-17
Usind reserved keywords as aliases in Cypher queries


GehriJan:
**Describe the bug**
Using reserved keywords or function names (e.g., `count`, `exists`, `coalesce`, `match`, `set`, etc.) as aliases in Cypher queries (e.g., `RETURN count(n) AS count`) causes parsing errors. The grammar does not distinguish between identifiers and reserved keywords in alias positions, leading to ambiguity and rejection of valid queries.

Specifically, all the keywords specified in `safe_keywords` in the file `src/.../parser/cypher_gram.y` on line `2467` and following (and for my testing also _only_ these keywords) lead to this error.

**How are you accessing AGE (Command line, driver, etc.)?**
This is not important to the issue, but the database was accessed through pgAdmin.

**What data setup do we need to do?**
```pgsql
SELECT * from cypher('my_graph_name', $$
  CREATE (a:Part {part_num: '123'}), 
         (b:Part {part_num: '345'}), 
         (c:Part {part_num: '456'}), 
         (d:Part {part_num: '789'})
$$) as (a agtype);
```

**What is the necessary configuration info needed?**
There is no special configuration needed for this.

**What is the command that caused the error?**
```pgsql
SELECT * from cypher('my_graph_name', $$
  RETURN 1 AS count
$$) as (a agtype);
```
```
ERROR:  syntax error at or near "count"
LINE 2:  RETURN 1 AS count
```

**Expected behavior**
Aliases should allow any valid identifier, including function names, unless explicitly forbidden. Neo4j, for example, allows the above queries to be executed.

**Environment (please complete the following information):**
- PostgreSQL version: PostgreSQL 16.13 (Debian 16.13-1.pgdg13+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 14.2.0-19) 14.2.0, 64-bit
- AGE Version: 1.16.0

**Additional context**

crprashant:
Hi @rafsun42 — I have a fix up for this in **#2415**.

[...]

PR #2415, opened 2026-04-27
Allow safe Cypher reserved keywords in alias positions (#2355)

crprashant:
# PR: Allow safe Cypher reserved keywords in alias positions (#2355)

**Branch:** `crprashant:fix/2355-reserved-keywords-as-aliases` → `apache:master`
**Open URL:** https://github.com/crprashant/age/pull/new/fix/2355-reserved-keywords-as-aliases
**Closes:** #2355

---

## Summary

`RETURN 1 AS count` (and equivalents using any of 49 non-conflicting reserved
keywords) failed with `syntax error at or near "count"`. This PR fixes it by
introducing a tightly-scoped `var_name_alias` grammar rule used only at the
alias-binding sites of `RETURN`, `WITH`, `YIELD`, and `UNWIND`. The set of
keywords newly accepted as aliases is exactly `safe_keywords` — the same set
already accepted in `func_name` and `schema_name`.

## Reproducer (from the issue)

```sql
LOAD 'age';
SET search_path = ag_catalog;
SELECT * FROM create_graph('g');
SELECT * FROM cypher('g', $$ RETURN 1 AS count $$) AS (a agtype);
```

| | Before | After |
|--|--|--|
| Result | `ERROR: syntax error at or near "count"` | `1` (1 row) |

## Root cause

In `src/backend/parser/cypher_gram.y` the productions

```
yield_item   : expr AS var_name
return_item  : expr AS var_name
unwind       : UNWIND expr AS var_name
```

referenced `var_name`, which expands only to `symbolic_name` (i.e. plain
`IDENTIFIER`). Reserved keywords — even those that pose no parsing ambiguity —
were therefore rejected as aliases, although `schema_name` already includes
`reserved_keyword` and `func_name` already includes `safe_keywords`.

## Fix

A new non-terminal `var_name_alias` is introduced and used only in the three
alias-binding sites above:

```yacc
var_name_alias:
    var_name
    | safe_keywords
    {
        $$ = pstrdup((char *) $1);
    }
    ;
```

`var_name` itself is **not** broadened, intentionally. Allowing
`safe_keywords` everywhere `var_name` is used produces **156 shift/reduce
conflicts** in bison (verified locally) because keyword tokens collide with
their syntactic roles inside expressions and patterns. Restricting the
broadening to `AS`-bound alias positions removes that ambiguity entirely —
the build is conflict-free.

`conflicted_keywords` (`END`, `NULL`, `TRUE`, `FALSE`) remain rejected
everywhere; they are genuinely ambiguous with literal / CASE productions.

## Behavior matrix

| Form | Before | After | Why |
|---|---|---|---|
| `RETURN 1 AS count` | error | ✅ ok | `count` ∈ safe_keywords |
| `RETURN 1 AS exists / coalesce / match / where / order / limit / distinct / optional / detach / contains / starts / ends / in / is / not / yield / call / ...` | error | ✅ ok | safe_keywords |
| `RETURN 1 AS count, 2 AS exists, 3 AS where` | error | ✅ ok | multiple aliases |
| `WITH 1 AS count RETURN 1 AS x` | error | ✅ ok | WITH alias binding |
| `UNWIND [1,2,3] AS row RETURN 1 AS x` | error | ✅ ok | UNWIND alias binding |
| `RETURN 1 AS null / true / false / end` | error | ❌ still error | conflicted_keywords (intentional) |
| `MATCH (count) RETURN 1 AS x` | error | ❌ still error | pattern bindings unchanged (intentional) |
| `RETURN 1 AS my_alias` | ✅ ok | ✅ ok | unaffected |

### Known limitation (intentional, scope-bounded)

Even with this PR, **referencing** an alias whose name is a keyword (e.g.
`WITH 1 AS count RETURN count`) still fails, because `expr_var` reads through
`var_name`, which is unchanged. Broadening `expr_var` to accept keywords
re-introduces the 156 grammar conflicts. The literal repro from the issue
(`RETURN 1 AS count`) is resolved; reading keyword-named aliases back is a
deeper grammar refactor and out of scope here. This is documented inline
in the regression test.

## Testing

* **Build:** `make PG_CONFIG=/usr/lib/postgresql/18/bin/pg_config -j$(nproc)` —
  clean, no new bison warnings, no shift/reduce conflicts.
* **Regression suite:** `make installcheck` on PostgreSQL 18.
  Result: **32/34 tests pass**, including the new `reserved_keyword_alias` test.
  The remaining failure (`age_upgrade`) is **pre-existing on master** and
  unrelated to this change.
* **PostgreSQL version coverage:** the change is parser-only and
  PG-version-agnostic. The fix lives entirely in `cypher_gram.y` and the
  generated `cypher_gram.c`; no PG-version-specific APIs are touched. The
  reporter's environment was AGE 1.16.0 on PostgreSQL 16.13. Building
  `master` against PG 16 currently fails for unrelated reasons
  (`cypher_set.c`/`age.c` use PG18-only APIs such as `index_close` and
  `age_shmem_startup_hook`); this is independent of #2355 and is not
  addressed here.

### New regression test

`regress/sql/reserved_keyword_alias.sql` covers:

1. The exact reproducer from the issue.
2. 18 representative safe_keywords as `RETURN ... AS <kw>`.
3. Multiple keyword aliases in one projection.
4. `WITH ... AS <kw>` alias binding.
5. `UNWIND ... AS <kw>` alias binding.
6. Negative tests for `END / NULL / TRUE / FALSE` (must still error).
7. Negative test for keyword in pattern-variable position (must still error).
8. A normal-identifier sanity check.

## Risk analysis

* **Grammar conflicts:** zero new conflicts (verified — bison was clean,
  build had no `-Wconflicts-sr` warning).
* **Call-site impact:** only three productions reference `var_name_alias`
  (yield_item, return_item, unwind). All other 6 use sites of `var_name`
  (named paths, pattern variables, edge bindings, `expr_var`,
  `var_name_opt`) are untouched.
* **Output / semantics:** only the parser accepts a strictly larger input
  language. Accepted-before queries continue to behave identically.

## Files changed

| File | Notes |
|---|---|
| `src/backend/parser/cypher_gram.y` | New `var_name_alias` rule; alias-binding productions switched to it. |
| `Makefile` | New test `reserved_keyword_alias` added to `REGRESS` between `security` and `drop`. |
| `regress/sql/reserved_keyword_alias.sql` | New regression test. |
| `regress/expected/reserved_keyword_alias.out` | Expected output. |

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,7 +177,8 @@ REGRESS = scan \
           predicate_functions \
           map_projection \
           direct_field_access \
-          security
+          security \
+          reserved_keyword_alias
 
 ifneq ($(EXTRA_TESTS),)
   REGRESS += $(EXTRA_TESTS)
--- a/regress/expected/reserved_keyword_alias.out
+++ b/regress/expected/reserved_keyword_alias.out
@@ -1,0 +1,248 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Regression coverage for issue #2355:
+ *   Cypher reserved keywords from the `safe_keywords` set must be accepted
+ *   in alias positions (RETURN/WITH/YIELD ... AS <name>, UNWIND ... AS <name>).
+ *   Conflicting tokens (END / NULL / TRUE / FALSE) must remain rejected.
+ *   Pattern variable bindings are intentionally still restricted to plain
+ *   identifiers.
+ */
+LOAD 'age';
+SET search_path TO ag_catalog;
+SELECT * FROM create_graph('issue_2355');
+NOTICE:  graph "issue_2355" has been created
+ create_graph 
+--------------
+ 
+(1 row)
+
+-- The exact reproducer from the issue (previously failed with
+-- "syntax error at or near "count"").
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS count $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+-- Representative coverage across the safe_keywords set as RETURN aliases.
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS exists      $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS coalesce    $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS match       $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS return      $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS where       $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS order       $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS limit       $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS distinct    $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS optional    $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS detach      $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS contains    $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS starts      $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS ends        $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS in          $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS is          $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS not         $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS yield       $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+SELECT * FROM cypher('issue_2355', $$ RETURN 1 AS call        $$) AS (a agtype);
+ a 
+---
+ 1
+(1 row)
+
+-- Multiple keyword aliases in one projection.
+SELECT * FROM cypher('issue_2355',
+    $$ RETURN 1 AS count, 2 AS exists, 3 AS where $$
+) AS (count agtype, ex agtype, w agtype);
+ count | ex | w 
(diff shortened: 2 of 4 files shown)


jrgemignani:
@crprashant Please address these minor issues -

## Recommendation

1. **Drop the redundant `pstrdup` in the new action body.**
   The `safe_keywords` rule already runs `KEYWORD_STRDUP($1)` on every alternative, which is `pnstrdup`-backed. The extra `pstrdup((char *) $1)` in `var_name_alias` is a second heap allocation per parse with no purpose. Replace it with `$$ = (char *) $1;` to match the established pattern in `schema_name`:
   ```yacc
   schema_name:
       symbolic_name
       | reserved_keyword
           {
               /* we don't need to copy it, as it already has been */
               $$ = (char *) $1;
           }
       ;
   ```
   This is a one-line cosmetic improvement that also makes the code symmetric with neighbouring rules and slightly reduces parser allocations on alias-heavy queries.

2. **Add an inline comment in `var_name_alias` documenting the unreadable-alias limitation, with an issue link.**
   The PR description correctly notes that `WITH 1 AS count RETURN count` still fails because `expr_var` reads through `var_name`, which is unchanged. That asymmetry (writable but not readable) will surprise the next maintainer who reads only the grammar file. Add 2–3 lines pointing to the follow-up issue, e.g.:
   ```c
   /*
    * NOTE: Reading a keyword-named alias back (e.g. WITH 1 AS count RETURN count)
    * is intentionally still rejected — broadening expr_var to accept safe_keywords
    * reintroduces ~156 shift/reduce conflicts. Tracked in issue #NNNN.
    */
   ```
   File the follow-up issue and reference its number here so the limitation is discoverable from the source, not just the merged PR description.

3. **Add two regression cases to round out coverage.**
   - **Backtick-quoted alias positive case:** `RETURN 1 AS \`count\`` — confirms that quoting forces the `IDENTIFIER` token path and continues to work, so future grammar refactors don't accidentally regress quoted identifiers.
   - **Known-limitation negative case:** `WITH 1 AS count RETURN count` — currently fails. Adding it with the failing expected output gives the next contributor a precise file to update when they fix `expr_var`, and serves as permanent self-documentation in the regress tree. Mark it clearly in the test file with a comment like `-- Known limitation: see issue #NNNN`.

4. **Fix the doc nit in the `var_name_alias` comment.**
   The current comment says the new rule permits "the same set of non-conflicting reserved keywords accepted by `safe_keywords` (and already accepted in `func_name` / `schema_name`)." That is slightly inaccurate: `func_name` accepts `safe_keywords`, but `schema_name` accepts the broader `reserved_keyword` (which includes `conflicted_keywords`). Tighten the wording to "(already accepted in `func_name`; `schema_name` accepts the broader `reserved_keyword`)" so a future reader doesn't infer that `schema_name` is a 1:1 precedent for the new rule.

crprashant:
@jrgemignani thanks for the careful review — pushed [49a424e](https://github.com/apache/age/pull/2415/commits/49a424ead5a46eaae272b8496af4de2f1d8ed54e) addressing all four points:

1. **Redundant `pstrdup` removed.** `safe_keywords` already heap-copies via the `KEYWORD_STRDUP` macro (`pnstrdup((kw), strlen(kw))`), so the action is now `$$ = (char *) $1;` — the same pattern `schema_name` uses for its `reserved_keyword` branch. An inline comment explains why no further copy is needed.
2. **Read-back limitation documented in code.** Added a `NOTE:` block above `var_name_alias` explaining that `WITH 1 AS count RETURN count` is intentionally still rejected because broadening `expr_var`/`var_name` reintroduces ~156 shift/reduce conflicts, and linked it to the new follow-up issue **#2416**.
3. **Doc comment tightened.** Replaced "(and already accepted in `func_name` / `schema_name`)" with "(already accepted in `func_name`; `schema_name` accepts the broader `reserved_keyword`)" — accurately reflects that `schema_name` is a strict superset of what `var_name_alias` accepts.
4. **Regression coverage extended.** `regress/sql/reserved_keyword_alias.sql` now also exercises:
   - **Positive:** ``RETURN 1 AS `count` `` and ``WITH 1 AS `count` RETURN `count` `` — the backtick-quoted IDENTIFIER path, so future grammar refactors can't silently regress quoted identifiers when the unquoted form is also a keyword.
   - **Negative known-limitation:** `WITH 1 AS count RETURN count` is captured with its `syntax error at end of input` output and a comment pointing at #2416, giving the next contributor who tackles `expr_var` a precise file to update.

Validation re-run on the amended commit:

- Build/Regression (PG18, `COPT=-Werror`, `EXTRA_TESTS="pgvector fuzzystrmatch pg_trgm"`): **36/37** — only the pre-existing `age_upgrade` failure remains (same as `master`).
- Python driver: all 4 suites green.
- Go driver: all tests green.
- Node.js driver: 31/31 green.
- JDBC driver (`gradle clean test`): `BUILD SUCCESSFUL`, every `*Test PASSED`.
- No new bison shift/reduce conflicts.

Happy to keep iterating if anything else stands out.

jrgemignani:
@crprashant Can you address the copilot messages above about CRLF? These files are for linux and need to conform to that standard. 

>This new expected output file appears to be committed with CRLF (Windows) line endings (existing regress/expected/*.out files use LF). Please convert it to LF-only line endings so pg_regress comparisons are stable across platforms.

crprashant:
@jrgemignani good catch — thanks! Pushed [f686e93](https://github.com/apache/age/pull/2415/commits/f686e93423fc42e0ffe78d993997466ad75e4bbd) with both new regression files normalized to LF-only line endings, matching the rest of the `regress/sql/*.sql` and `regress/expected/*.out` tree. The earlier amend accidentally retained CRLF from a Windows-side scratch file.

Verified:

```
$ file regress/sql/reserved_keyword_alias.sql regress/expected/reserved_keyword_alias.out
regress/sql/reserved_keyword_alias.sql:      ASCII text
regress/expected/reserved_keyword_alias.out: ASCII text
```

(`file` reports `with CRLF line terminators` when CR is present; both now report plain `ASCII text` ⇒ LF-only.)

The expected output was regenerated from a fresh `installcheck` run against the LF `.sql`, so the captured `SET search_path`, `LOAD`, and other echoed lines are all LF-only — no mixed endings remain. Build/Regression on PG18 with `COPT=-Werror EXTRA_TESTS="pgvector fuzzystrmatch pg_trgm"` is still 36/37 (only the pre-existing `age_upgrade` fails). No driver test impact since no driver code or fixtures changed.

jrgemignani (approved):

PR #2419, opened 2026-05-01
Fix VLE [*0..N] zero-hop self-binding when edge label is missing (#2382)

### LINKED ISSUE #2382: Variable-length relationship patterns with a zero lower bound may fail to produce the expected zero-hop self-binding.
**Describe the bug**
Variable-length relationship patterns with a zero lower bound may fail to produce the expected zero-hop self-binding.

In the repro below, the graph contains a single `Person` node and no `FRIEND` relationships. Under Cypher semantics, the pattern `[:FRIEND*0..1]` should still match with length `0`, so `f` should bind to the same node as `p`.

Instead, Apache AGE returns `f = null`, and `collect(f.name)` becomes `[]` instead of `['Alice']`.

```pgsql
  CREATE (:Person {name: 'Alice'})
$$) AS (v agtype);
```

- PostgreSQL version: `18.1`

```pgsql
  MATCH (p:Person {name: 'Alice'})
  OPTIONAL MATCH (p)-[:FRIEND*0..1]-(f:Person)
  RETURN p.name AS person, f.name AS friend
$$) AS (person agtype, friend agtype);
```

Returned result on AGE:
```text
person | friend
-------+-------
Alice  | null
```

**Expected behavior**
The zero-hop match should bind `f` to the same node as `p`, so the expected result is:

```text
person | friend
-------+-------
Alice  | Alice
```

Neo4j returns exactly that result for the equivalent Cypher query.

- Version: Apache AGE `1.7.0`
- PostgreSQL: `18.1`
- Host OS: Windows 10
- Architecture: x86_64
- Deployment: Docker

crprashant:
## Summary

Fixes [#2382](https://github.com/apache/age/issues/2382): variable-length relationship patterns with a zero lower bound (`[*0..N]`) failed to produce the zero-hop self-binding when the edge label did not exist in the graph.

Per Neo4j/openCypher semantics, a pattern like `(p)-[:FRIEND*0..1]-(f)` must always include the zero-length match where `f` binds to the same node as `p`, regardless of whether any `FRIEND` edge exists. Apache AGE was treating any missing edge label as fatal for the entire MATCH.

### Repro (before this PR)

```sql
SELECT * FROM cypher('fuzz_graph', $$ CREATE (:Person {name:'Alice'}) $$) AS (v agtype);

  MATCH (p:Person {name: 'Alice'})
  OPTIONAL MATCH (p)-[:FRIEND*0..1]-(f:Person)
  RETURN p.name AS person, f.name AS friend
$$) AS (person agtype, friend agtype);

-- Before:  Alice | NULL
-- After:   Alice | Alice   (matches Neo4j)
```

### Root cause

Two separate label-validity gates short-circuited the MATCH when an edge label was not in the cache:

- `match_check_valid_label()` injected a `WHERE false` (or volatile equivalent) at the MATCH level.
- `path_check_valid_label()` made the planner emit `NULL::agtype` constants in place of vertex bindings even when the vertex labels themselves were valid.

Neither check distinguished fixed-length patterns (which legitimately require an edge of the label) from VLE patterns with a lower bound of 0 (which don't).

There was also a latent runtime bug in `is_an_edge_match()`: when the user requested a label that did not exist, the InvalidOid check fell through the existing "no constraints → match all" fast-path. Once the parser-side gate was relaxed, `[:NOEXIST*0..N]` would have begun traversing arbitrary other-label edges. This PR fixes that path too so the zero-hop case is the only thing that matches.

### Fix

1. **`src/backend/parser/cypher_clause.c`** — new helper `is_zero_lower_bound_vle()` that defensively inspects the `vle()` FuncCall built by `build_VLE_relation()` (in `cypher_gram.y`). Both `match_check_valid_label()` and `path_check_valid_label()` use it to skip the false-where short-circuit only for relationships that are zero-bound VLEs. Mixed patterns where another segment is genuinely unsatisfiable still short-circuit correctly.
2. **`src/backend/utils/adt/age_vle.c`** — `is_an_edge_match()` returns false early when the requested label does not exist. The zero-hop self-binding is generated by `build_VLE_zero_container()` and never calls `is_an_edge_match()`, so this fix does not affect the zero-hop semantics.
3. **`regress/sql/cypher_vle.sql`** — seven regression cases:
   - plain MATCH `[:NOEXIST*0..1]` → zero-hop only (no leakage onto other-label edges in the graph)
   - OPTIONAL MATCH form (issue's exact repro shape)
   - `[*0..0]` → zero-hop only
   - fixed-length `[:NOEXIST*1..1]` → 0 rows (preserved)
   - OPTIONAL fixed-length missing → outer row preserved with NULL
   - mixed pattern `(a)-[:NOEXIST*0..1]-(b)-[:STILL_MISSING]-(c)` → 0 rows
   - sanity: zero-bound on an existing label (unchanged behaviour)

### Validation

- `make installcheck EXTRA_TESTS="pgvector fuzzystrmatch pg_trgm"`: 36/37 pass on PG18. The single failure (`age_upgrade`) is pre-existing on `master` at `54e19fac` and is unrelated to this change.
- All 7 new regression cases pass.
- `cypher_vle` (which is the suite the new tests live in) passes cleanly.
- Manual rebuild + `EXPLAIN (VERBOSE, ANALYZE)` confirms the planner no longer constant-folds `f.name` to `NULL::agtype` and the query returns exactly the expected zero-hop row.

### Notes for reviewers

- The helper validates that `rel->varlen` is a `FuncCall` named `"vle"` with at least 5 args before reading `args[3]`; any deviation falls back to the existing short-circuit, so future parser refactors will be safe.
- `A_Const` integer inspection follows the same direct field access (`val.ival.type` / `val.ival.ival`) used elsewhere in `cypher_gram.y` (e.g. line 696).
- No bison or grammar changes; no new shift/reduce conflicts.

Closes #2382.

### DIFF
--- a/regress/expected/cypher_vle.out
+++ b/regress/expected/cypher_vle.out
@@ -1220,5 +1220,120 @@ NOTICE:  graph "cypher_vle" has been dropped
 (1 row)
 
 --
+-- Issue #2382: variable-length relationships with a zero lower bound must
+-- still produce the zero-hop self-binding even when the edge label does not
+-- exist in the graph (Neo4j/openCypher semantics).
+--
+SELECT create_graph('issue_2382');
+NOTICE:  graph "issue_2382" has been created
+ create_graph 
+--------------
+ 
+(1 row)
+
+SELECT * FROM cypher('issue_2382', $$
+    CREATE (:Person {name: 'Alice'})-[:KNOWS]->(:Person {name: 'Bob'})
+$$) AS (v agtype);
+ v 
+---
+(0 rows)
+
+-- Plain MATCH on a non-existent edge label with [*0..N] must return the
+-- zero-hop self-binding row (Alice -> Alice). It must NOT match arbitrary
+-- edges of other labels (e.g. KNOWS).
+SELECT * FROM cypher('issue_2382', $$
+    MATCH (p:Person {name: 'Alice'})
+    MATCH (p)-[:NOEXIST*0..1]-(f:Person)
+    RETURN p.name AS person, f.name AS friend
+$$) AS (person agtype, friend agtype);
+ person  | friend  
+---------+---------
+ "Alice" | "Alice"
+(1 row)
+
+-- OPTIONAL MATCH form (the exact shape from the issue report).
+SELECT * FROM cypher('issue_2382', $$
+    MATCH (p:Person {name: 'Alice'})
+    OPTIONAL MATCH (p)-[:NOEXIST*0..1]-(f:Person)
+    RETURN p.name AS person, f.name AS friend
+$$) AS (person agtype, friend agtype);
+ person  | friend  
+---------+---------
+ "Alice" | "Alice"
+(1 row)
+
+-- [*0..0] still emits exactly the zero-hop self-binding.
+SELECT * FROM cypher('issue_2382', $$
+    MATCH (p:Person {name: 'Alice'})
+    MATCH (p)-[:NOEXIST*0..0]-(f:Person)
+    RETURN p.name AS person, f.name AS friend
+$$) AS (person agtype, friend agtype);
+ person  | friend  
+---------+---------
+ "Alice" | "Alice"
+(1 row)
+
+-- Fixed-length (lower bound > 0) on a missing label must still return zero
+-- rows: there is no edge of that label, so the pattern is unsatisfiable.
+SELECT * FROM cypher('issue_2382', $$
+    MATCH (p:Person {name: 'Alice'})
+    MATCH (p)-[:NOEXIST*1..1]-(f:Person)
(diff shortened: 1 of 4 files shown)


jrgemignani:
@crprashant This PR will need to wait on #2420 #2421 and #2422 (which the others are waiting on). As the prior 2 are significant VLE rewrites, it is best to wait for those.

jrgemignani:
@crprashant Could you rebase this one as the master has changed significantly.

crprashant:
@jrgemignani Rebased onto the latest master (23cbe57a). The fix re-applied cleanly to the rewritten VLE code; full regression suite passes (36/36) and cypher_vle including the new #2382 cases is green.

jrgemignani (approved):

PR #2420, opened 2026-05-01
perf: VLE terminal-qual rewrite

jrgemignani:
perf: VLE terminal-qual rewrite — emit endpoint equalities instead of SRF qual functions.

Removes the per-row age_match_vle_terminal_edge and age_match_two_vle_edges qual functions from VLE query plans. The cypher transformer now emits the endpoint match as a plain graphid/int8 equality on new SRF output columns, evaluated by the planner like any other join clause — no detoasting, no per-row C function dispatch. Stages land as one commit:

  S1  Inline start_vid/end_vid in VLE_path_container header
  S2  Read VLE qual endpoints from header-only TOAST slice
  S4  Emit start_id/end_id as scalar SRF output columns
      (age_vle now RETURNS SETOF record with edges/start_id/end_id)
  S5  Cypher transformer rewrites terminal-edge match quals as
      integer equalities (drops age_match_vle_terminal_edge call)
  S6  Cypher transformer emits graphid equality for two-VLE-edge
      joins (drops age_match_two_vle_edges call)

Performance (SF3 LDBC SNB, 5 runs/3 warmup, vs clean master baseline_v2):

  IC sum    198,958 → 109,322 ms   −45.05 %  (1.82× end-to-end speedup)
  IC1   8,625 →  4,600 ms  −46.67 %
  IC3  21,239 →  9,784 ms  −53.93 %
  IC5  21,051 →  5,696 ms  −72.94 %
  IC6  15,916 →  4,447 ms  −72.06 %
  IC9  44,839 → 21,161 ms  −52.81 %
  IC10 13,104 →  2,432 ms  −81.44 %
  IC11 11,676 →    241 ms  −97.93 %  (48× speedup)
  IC2/4/7/8/12: parity (within ±3.3 %; IC4 is −2.47 %, no regression)
  IS sum:  1,009 → 1,004 ms   −0.51 %  (no VLE traffic)
  IU sum:     77 →    71 ms   −8.38 %  (IU1 −16.09 %; incidental)

Memory: header-only TOAST slice for VLE qual evaluation avoids detoasting full path containers on every row; reduces per-call palloc/pfree churn in long DFS paths. No measured RSS change.

Dead-code removal:
  - Bodies of age_match_vle_terminal_edge and age_match_two_vle_edges are gone from age_vle.c (~225 lines).  C entry points remain as error-raising stubs solely so the upgrade-test snapshot loader (which sources an older 1.7.0_initial SQL against the current age.so) can resolve the symbols before the immediate ALTER EXTENSION UPDATE drops them.  No regress test references either function.
  - SQL CREATE FUNCTION declarations removed from fresh install (sql/agtype_typecast.sql).
  - DROP FUNCTION IF EXISTS for both qual functions added to the upgrade script (age--1.7.0--y.y.y.sql).

API change: ag_catalog.age_vle(...) now RETURNS SETOF record with output columns (edges agtype, start_id graphid, end_id graphid) instead of RETURNS SETOF agtype.  Both 7-arg and 8-arg overloads are updated in fresh-install (sql/agtype_typecast.sql) and upgrade (age--1.7.0--y.y.y.sql) paths.  age_match_vle_terminal_edge and age_match_two_vle_edges are dropped on upgrade and absent from fresh installs.  Internal AGE callers are unaffected; external SQL that called any of these directly must adapt.

Tested on PostgreSQL 18.3 (REL_18_STABLE): all 34 regression tests pass (installcheck), warning-free build.

modified:   regress/expected/cypher_match.out
modified:   regress/expected/expr.out
modified:   sql/agtype_typecast.sql
modified:   src/backend/parser/cypher_transform_entity.c
modified:   src/include/parser/cypher_transform_entity.h


jrgemignani:
@MuhammadTahaNaveed Rebased and addressed copilots concerns.

MuhammadTahaNaveed (approved):

PR #2422, opened 2026-05-01
Fix upgrade test: allow function removal

jrgemignani:
Fix upgrade test: allow function removal and detect more deficiencies.

The age_upgrade regression test (added in #2364, refined in #2377, #2397, install and a synthetic-initial -> current upgrade. Three gaps surfaced in practice:

1. Function removal forced permanent C stubs. The synthetic '_initial' install is built from a fixed historical commit. CREATE EXTENSION resolves every CREATE FUNCTION ... AS '$libdir/age', '<symbol>' via dlsym at install time when check_function_bodies is on (the default). If a developer retires a C entry point in HEAD's age.so, step 10 aborts with "could not find function ... in file age.so" -- even though the immediately-following ALTER EXTENSION UPDATE would DROP that SQL declaration. The only way to keep the test green was to leave a permanent error-raising stub in age.so, and to remember to add a DROP to the upgrade template.

2. Modifications were under-detected. The function-property-change query did not compare probin or prosrc, so a C function whose symbol was renamed in the upgrade template, or a SQL/plpgsql function whose body changed in either path, slipped through.

3. Extension membership was not checked. A template that CREATEs an object but never ALTER EXTENSION ADDs it leaves a row in pg_proc/pg_class but no pg_depend deptype='e' link. pg_dump --extension would diverge, but the existing per-catalog diff queries all returned 0 rows.

Changes (regress/sql/age_upgrade.sql + regress/expected/age_upgrade.out):

* Step 10 wraps the synthetic CREATE EXTENSION in SET check_function_bodies = off; ... RESET check_function_bodies; Symbol resolution is deferred to call time. Step 11's ALTER EXTENSION UPDATE then DROPs any retired functions before any plan can call them. Step 35's fresh CREATE EXTENSION runs at the GUC default, so HEAD's sql/ <-> HEAD's age.so consistency is still enforced on the production install path.

* Steps 2 and 13 add probin and prosrc to the function snapshot. Step 21 reports probin and prosrc divergences alongside the existing property-change columns.

* Steps 7b and 18b add an extension-membership snapshot from pg_depend deptype='e' filtered to the AGE extension OID. Every member is labeled by stable identity (regprocedure, regtype, regoperator, opfname+strategy+types, etc.), never by raw OID, so OID drift between fresh and upgrade installs cannot produce false positives. Steps 33a and 33b report MISSING / EXTRA members. Step 34 adds extmembers_match to the summary row.

* Section-header step ranges updated to include the new sub-steps.

The change is fully self-contained: only regress/sql/age_upgrade.sql and regress/expected/age_upgrade.out are modified. No production C, SQL, build, or test files are touched. All 34 regression tests pass on a clean tree.

Mutation-tested with 8 cases against the unmutated tree: baseline pass; remove-function-with-DROP pass (no stub needed); remove-function-forget- DROP fail; add-function-with-CREATE pass; add-function-forget-CREATE fail; volatility-change-no-template fail; volatility-change-with-CREATE- OR-REPLACE pass; C-symbol-rename-no-template fail. All eight expected outcomes observed.

All 34 regression tests pass.

modified:   regress/expected/age_upgrade.out
modified:   regress/sql/age_upgrade.sql

### DIFF
--- a/regress/expected/age_upgrade.out
+++ b/regress/expected/age_upgrade.out
@@ -27,15 +27,29 @@
 -- upgrade template.
 --
 -- Compared catalogs:
---   pg_proc     — functions, aggregates, procedures (name, args, properties)
+--   pg_proc     — functions, aggregates, procedures (name, args, properties
+--                 including probin/prosrc to catch C-symbol renames and
+--                 SQL-body changes)
 --   pg_class    — tables, views, sequences, indexes (name, kind)
 --   pg_type     — types (name, type category)
 --   pg_operator — operators (name, left/right types)
 --   pg_cast     — casts involving AGE types (source, target, context)
 --   pg_opclass  — operator classes (name, access method)
 --   pg_constraint — constraints (name, type, table, referenced table)
+--   pg_depend   — extension membership (every AGE-owned object must be
+--                 linked back to the extension via deptype='e'; catches
+--                 a missing ALTER EXTENSION age ADD ... in the template)
 --
 -- All comparison queries should return 0 rows.
+--
+-- Note on synthetic-initial install (step 10): the synthetic '*_initial'
+-- snapshot is built from a fixed historical commit, so its CREATE FUNCTION
+-- statements may reference C symbols that have since been removed from the
+-- current age.so. Step 10 disables check_function_bodies so that dlsym is
+-- deferred to call time; the immediately-following ALTER EXTENSION UPDATE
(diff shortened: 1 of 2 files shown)


muhammadshoaib:
Looks fine.

jrgemignani:
> Looks fine.

Looks fine approval? or merge? ;)

muhammadshoaib (approved):
Looks fine

PR #2423, opened 2026-05-02
Zero-initialize parent_cpstate in analyze_cypher

hari90:
cypher_parsestate parent_cpstate is declared on the stack in analyze_cypher() and only pstate, graph_name, and params are explicitly set before it is passed to make_cypher_parsestate(). The latter reads parent_cpstate->subquery_where_flag (and other fields) in cypher_parse_node.c, which leaves them with indeterminate values. Yugabyte UBSan flagged the garbage bool (value 8) and aborted the backend.

Use MemSet to zero the struct before populating the explicit fields so all remaining members start with a defined value.


jrgemignani (approved):

hari90:
Thank you

PR #2376, opened 2026-04-09
VLE cache: replace snapshot invalidation with per-graph

jrgemignani:
Replace snapshot-based VLE cache invalidation with per-graph version
counters and add performance optimizations for VLE traversal.

Cache invalidation:
- Add DSM (PG 17+) and shmem (PG <17) per-graph monotonic version
  counters for cross-backend cache invalidation
- Replace snapshot xmin/xmax/curcid comparison with version counter
  check in is_ggctx_invalid(), with snapshot fallback for safety
- Add executor hooks in CREATE/DELETE/SET/MERGE to increment the
  graph version counter on mutations
- Add SQL trigger function (age_invalidate_graph_cache) for catching
  SQL-level mutations (INSERT/UPDATE/DELETE/TRUNCATE)
- Auto-install trigger on new label tables via label_commands.c
  with LookupFuncName check for backward compatibility
- Add TRUNCATE interception in ProcessUtility hook (ag_catalog.c)
- Add shmem_request/startup hooks for PG <17 (age.c)

Thin entries (lazy property fetch):
- Replace Datum edge_properties/vertex_properties with 6-byte
  ItemPointerData TID in vertex_entry and edge_entry
- Add get_vertex_entry_properties() and get_edge_entry_properties()
  that do heap_fetch via stored TID on demand
- Add is_an_edge_match() fast path: skip property access entirely
  for label-only VLE patterns (the common case)

Edge cleanup list removal:
- Remove ggctx->edges linked list from GRAPH_global_context. With
  thin entries, edge_entry has no palloc'd sub-structures to free
  (TIDs are inline). The cleanup list existed to pfree datumCopy'd
  edge_properties Datums, which no longer exist. hash_destroy()
  handles all edge hash table memory. Saves ~5.6 GB at SF10.

Performance optimizations:
- Reduce VERTEX/EDGE_HTAB_INITIAL_SIZE from 1,000,000 to 10,000.
  PG dynahash grows automatically; large initial size wastes memory.
- Eliminate extract_variadic_args in age_match_vle_terminal_edge:
  direct PG_GETARG_DATUM + cached arg types via fn_extra
- Eliminate extract_variadic_args in age_match_vle_edge_to_id_qual:
  same pattern
- Add agtype_access_operator 2-arg fast path: bypasses
  extract_variadic_args_min for the common property access case
- Add age_tointeger 1-arg fast path: bypasses extract_variadic_args
  with cached arg type
- Add GraphIdStack (flat array-based) DFS stacks replacing
[...]

Issue #1878, opened 2024-05-15
Flaky test age_global_graph fails on slow machines


saygoodbyye:
**Describe the bug**
Test age_global_graph sometimes fails. Mainly appears when the machine is running slowly.

**How are you accessing AGE (Command line, driver, etc.)?**
Accessing AGE through command line.

**What data setup do we need to do?**
Apache AGE (master branch) with PostgreSQL (REL_16_STABLE).

**What is the necessary configuration info needed?**
```
./configure CFLAGS=" -Og" --enable-tap-tests --enable-debug  --enable-cassert
```

**What is the command that caused the error?**
To reproduce fail several times faster I do this Makefile patch (Only to reproduce fail faster. This fail may occur even without these changes):
```
diff --git a/Makefile b/Makefile
index 1224fc2a..0d51a9e0 100644
--- a/Makefile
+++ b/Makefile
@@ -85,35 +85,7 @@ SQLS := $(addsuffix .sql,$(SQLS))
 DATA_built = $(age_sql)
 
 # sorted in dependency order
-REGRESS = scan \
-          graphid \
-          agtype \
-          catalog \
-          cypher \
-          expr \
-          cypher_create \
-          cypher_match \
-          cypher_unwind \
-          cypher_set \
-          cypher_remove \
-          cypher_delete \
-          cypher_with \
-          cypher_vle \
-          cypher_union \
-          cypher_call \
-          cypher_merge \
-          cypher_subquery \
-          age_global_graph \
-          age_load \
-          index \
-          analyze \
-          graph_generation \
-          name_validation \
-          jsonb_operators \
-          list_comprehension \
-          map_projection \
-          drop
-
+REGRESS=$(shell printf "age_global_graph %.0s" `seq 100` )
 srcdir=`pwd`
 
 ag_regress_dir = $(srcdir)/regress
```
Then i run make installcheck:
```
for i in `seq 1000`;do make installcheck || break;done
```
And get:
```
<...>
ok 60        - age_global_graph                         1440 ms
not ok 61    - age_global_graph                         1126 ms
ok 62        - age_global_graph                         1181 ms
ok 63        - age_global_graph                         1037 ms
<...>
```
regression.diffs:
```
diff -U3 /home/test/work/age/regress/expected/age_global_graph.out /home/test/work/age/regress/results/age_global_graph.out
--- /home/test/work/age/regress/expected/age_global_graph.out   2024-05-14 13:18:45.498945171 +0700
[...]

PR #2433, opened 2026-06-09
fix: remove pthread_mutex that causes self-deadlock on VLE queries

### LINKED ISSUE #2432: pthread_mutex in manage_GRAPH_global_contexts causes permanent self-deadlock on VLE queries
**Describe the bug**

`pthread_mutex` in `manage_GRAPH_global_contexts()` causes permanent self-deadlock on VLE queries. When `ereport(ERROR)` is raised while the mutex is held (e.g., statement timeout, query cancellation, OOM), PostgreSQL's `siglongjmp` jumps to the error handler, skipping `pthread_mutex_unlock()`. The mutex remains permanently locked. Any subsequent VLE query on the same backend connection deadlocks on itself — the process hangs forever in `pthread_mutex_lock()` with `__owner == own PID`.

- psql (command line), but the bug affects any client/driver.

```pgsql
LOAD 'age';

SELECT create_graph('test_deadlock');

SELECT * FROM cypher('test_deadlock', $$
    UNWIND range(1, 50000) AS i
    CREATE (:Node {id: i})
$$) AS (v agtype);

SELECT * FROM cypher('test_deadlock', $$
    MATCH (a:Node), (b:Node)
    WHERE b.id = a.id + 1
    CREATE (a)-[:LINK {weight: a.id}]->(b)
$$) AS (e agtype);

-- Load graph context into cache first
SELECT * FROM cypher('test_deadlock', $$
    MATCH path = (a)-[r*1..2]->(b)
    RETURN path LIMIT 1
$$) AS (path agtype);

-- Invalidate cached context by modifying the graph
SELECT * FROM cypher('test_deadlock', $$
    CREATE (:Dummy {x: 1})
$$) AS (v agtype);
```

- Any AGE version that includes PR #1881 
- PostgreSQL 16, 17, or 18
- No special configuration needed

Repeat cache-invalidate + timeout in a loop. The timeout must hit during graph context reload (while the mutex is held). It may take a few iterations depending on machine speed.

```pgsql
-- Repeat: invalidate cache, then cancel VLE query via statement_timeout.
-- Each round has a chance of hitting the mutex-held window.
-- Once it hits, every subsequent VLE query on this connection hangs forever.

-- Round 1
SELECT * FROM cypher('test_deadlock', $$ CREATE (:T1 {x: 1}) $$) AS (v agtype);
SET statement_timeout = '1ms';
SELECT * FROM cypher('test_deadlock', $$
    MATCH path = (a)-[r*1..3]->(b) RETURN path LIMIT 1
$$) AS (path agtype);
RESET statement_timeout;

-- Round 2
SELECT * FROM cypher('test_deadlock', $$ CREATE (:T2 {x: 2}) $$) AS (v agtype);
SET statement_timeout = '1ms';
SELECT * FROM cypher('test_deadlock', $$
    MATCH path = (a)-[r*1..3]->(b) RETURN path LIMIT 1
$$) AS (path agtype);
RESET statement_timeout;

-- Round 3
SELECT * FROM cypher('test_deadlock', $$ CREATE (:T3 {x: 3}) $$) AS (v agtype);
SET statement_timeout = '1ms';
SELECT * FROM cypher('test_deadlock', $$
    MATCH path = (a)-[r*1..3]->(b) RETURN path LIMIT 1
$$) AS (path agtype);
RESET statement_timeout;

-- (add more rounds if needed)

-- Final test: if any round above hit the mutex window,
-- this query hangs forever (self-deadlock).
SELECT * FROM cypher('test_deadlock', $$
    MATCH path = (a)-[r*1..2]->(b) RETURN path LIMIT 1
$$) AS (path agtype);
-- If it returns results, add more rounds above and retry.
```

To confirm with GDB:
```bash
gdb -batch -p <hung_pid> \
  -ex "print global_graph_contexts_container.mutex_lock.__data.__owner"
# Output: $1 = <hung_pid>   (owner == self → self-deadlock)
```

**Expected behavior**

VLE queries should continue to work normally after a query error or cancellation. A statement timeout on one query should not permanently break the backend connection.

- AGE Version: master (also affects PG16, PG17, PG18 branches — any version with PR #1881)
- PostgreSQL Version: 16, 17, 18

**Additional context**

The mutex was introduced in PR #1881 (fix for issue #1878). However, it is both unnecessary and harmful:

1. **Unnecessary:** The protected variable is a process-local `static` — no concurrent access exists. The test failure in #1878 was a catalog-level race, already fixed by the Assert→runtime check and `strndup` in the same PR. For cross-backend cache invalidation, PostgreSQL syscache uses `sinval` callbacks, and AGE PR #2376 already uses lock-free `pg_atomic_uint64` version counters in shared memory for this.

2. **Harmful:** `pthread_mutex` is incompatible with PostgreSQL's error handling. `ereport(ERROR)` uses `siglongjmp` to jump directly to the error handler, bypassing all code between the error site and the handler — including `pthread_mutex_unlock()`. Once skipped, the mutex is permanently locked for that backend process, and any subsequent VLE query self-deadlocks.

We will submit a PR with a fix.

crdv7:
Remove `pthread_mutex` from `age_global_graph.c` — it causes permanent self-deadlock on VLE queries.

`ereport(ERROR)` (statement timeout, query cancellation, OOM, etc.) triggers `siglongjmp` which skips `pthread_mutex_unlock()`. The mutex remains locked, and every subsequent VLE query on the same backend hangs forever in `pthread_mutex_lock()` with `__owner == own PID`.

Closes #2432

## Analysis

The mutex was introduced in PR #1881 (fix for #1878). It is both unnecessary and harmful:

1. **Unnecessary:** The protected variable is a process-local `static` — no concurrent access exists. The test failure in #1878 was a catalog-level race, already fixed by the Assert→runtime check and `strndup` in the same PR. For cross-backend cache invalidation, PostgreSQL syscache uses `sinval` callbacks, and PR #2376 already uses lock-free `pg_atomic_uint64` version counters in shared memory for this.

2. **Harmful:** `pthread_mutex` is incompatible with PostgreSQL's `ereport(ERROR)` / `siglongjmp` error handling. Any ERROR during the mutex-held window permanently locks it. This is the only use of `pthread` in the entire AGE codebase.

## Changes

`src/backend/utils/adt/age_global_graph.c` (1 file, +13 −50):

- Remove `#include <pthread.h>`
- Remove `GRAPH_global_context_container` struct (with `pthread_mutex_t` field)
- Revert to `static GRAPH_global_context *global_graph_contexts = NULL`
- Remove all `pthread_mutex_lock/unlock` calls (20 references)

The other fixes from PR #1881 (Assert→runtime check + `strndup`) remain untouched.

## Verification

- All regression tests pass (36/36)
- Deadlock reproduction via `statement_timeout`: no longer hangs after fix

### DIFF
--- a/src/backend/utils/adt/age_global_graph.c
+++ b/src/backend/utils/adt/age_global_graph.c
@@ -45,7 +45,6 @@
 #include "catalog/ag_label.h"
 #include "utils/ag_cache.h"
 
-#include <pthread.h>
 
 /* defines */
 #define VERTEX_HTAB_NAME "Vertex to edge lists " /* added a space at end for */
@@ -152,18 +151,8 @@ typedef struct GRAPH_global_context
     struct GRAPH_global_context *next; /* next graph */
 } GRAPH_global_context;
 
-/* container for GRAPH_global_context and its mutex */
-typedef struct GRAPH_global_context_container
-{
-    /* head of the list */
-    GRAPH_global_context *contexts;
-
-    /* mutex to protect the list */
-    pthread_mutex_t mutex_lock;
-} GRAPH_global_context_container;
-
 /* global variable to hold the per process GRAPH global contexts */
-static GRAPH_global_context_container global_graph_contexts_container = {0};
+static GRAPH_global_context *global_graph_contexts = NULL;
 
 /*
  * VertexEdgeArray helpers — flat-array adjacency container used by
@@ -1054,12 +1043,10 @@ GRAPH_global_context *manage_GRAPH_global_contexts(char *graph_name,
      *     5) One or more other contexts do exist but, one or more are invalid.
      */
 
-    /* lock the global contexts list */
-    pthread_mutex_lock(&global_graph_contexts_container.mutex_lock);
 
     /* free the invalidated GRAPH global contexts first */
     prev_ggctx = NULL;
-    curr_ggctx = global_graph_contexts_container.contexts;
+    curr_ggctx = global_graph_contexts;
     while (curr_ggctx != NULL)
     {
         GRAPH_global_context *next_ggctx = curr_ggctx->next;
@@ -1076,7 +1063,7 @@ GRAPH_global_context *manage_GRAPH_global_contexts(char *graph_name,
              */
             if (prev_ggctx == NULL)
             {
-                global_graph_contexts_container.contexts = next_ggctx;
+                global_graph_contexts = next_ggctx;
             }
             else
             {
@@ -1089,8 +1076,6 @@ GRAPH_global_context *manage_GRAPH_global_contexts(char *graph_name,
             /* if it wasn't successfull, there was a missing vertex entry */
             if (!success)
             {
-                /* unlock the mutex so we don't get a deadlock */
-                pthread_mutex_unlock(&global_graph_contexts_container.mutex_lock);
 
                 ereport(ERROR, (errcode(ERRCODE_DATA_EXCEPTION),
                                 errmsg("missing vertex or edge entry during free")));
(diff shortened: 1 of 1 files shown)


jrgemignani:
@crdv7 Please see the above Copilot message.  *edit* nvm, this is pre-existing and needs to be addressed separately.

jrgemignani (approved):

PR #2440, opened 2026-06-17
Make ag_catalog ownership and built-in resolution explicit

jrgemignani:
AGE places all of its objects in the ag_catalog schema. Make the assumptions around that schema explicit so installs and upgrades behave predictably regardless of how a database is provisioned:

- Ownership-checked install: CREATE EXTENSION age installs into ag_catalog only when that schema does not already exist under a different owner, keeping ownership of AGE's catalog well-defined.
- Deterministic name resolution: the pg_upgrade helper functions resolve built-ins from pg_catalog first and schema-qualify their format()/hashtext() calls, so their behavior does not depend on what else is defined in ag_catalog.
- README note describing ag_catalog ownership and the install-time check.

The upgrade script applies the same helper changes so existing installations get them on ALTER EXTENSION UPDATE. Adds an extension_security regression test covering the ownership check and the qualified-call / search_path properties.

Assisted-by: GitHub Copilot (Claude Opus 4.8)

modified:   Makefile
modified:   README.md
new file:   regress/expected/extension_security.out
new file:   regress/sql/extension_security.sql
modified:   sql/age_main.sql
modified:   sql/age_pg_upgrade.sql

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -220,7 +220,8 @@ REGRESS = scan \
           reserved_keyword_alias \
           agtype_jsonb_cast \
           containment_selectivity \
-          subgraph
+          subgraph \
+          extension_security
 
 ifneq ($(EXTRA_TESTS),)
   REGRESS += $(EXTRA_TESTS)
--- a/README.md
+++ b/README.md
@@ -215,6 +215,16 @@ LOAD 'age';
 SET search_path = ag_catalog, "$user", public;
 ```
 
+### Note on `ag_catalog` ownership
+
+AGE installs all of its objects into the `ag_catalog` schema. Install AGE
+(`CREATE EXTENSION age`) **before** granting the `CREATE` privilege on the
+database to other roles. A role that can create schemas could otherwise
+pre-create `ag_catalog` and own it; `CREATE EXTENSION age` therefore refuses to
+install when `ag_catalog` already exists and is owned by a different role. If you
+hit that error, drop the stray schema (`DROP SCHEMA ag_catalog CASCADE`) or
+transfer its ownership to the installing role, then retry.
+
 <h2><img height="20" src="/img/contents.svg">&nbsp;&nbsp;Using AGE with Non-Autocommit Clients (psycopg, JDBC, etc.)</h2>
 
 If you are using AGE from a database client that does **not** default to autocommit — most commonly `psycopg` v3 or JDBC — you must understand how PostgreSQL's transaction semantics apply to AGE's setup and DDL-like functions. Otherwise, you may see graphs or labels that appear to be created successfully, but are not visible from new connections.
--- a/age--1.7.0--y.y.y.sql
+++ b/age--1.7.0--y.y.y.sql
@@ -41,7 +41,10 @@
 CREATE FUNCTION ag_catalog.age_prepare_pg_upgrade()
     RETURNS void
     LANGUAGE plpgsql
-    SET search_path = ag_catalog, pg_catalog
+    -- Resolve built-in functions and operators from pg_catalog first so they
+    -- are not overridden by same-named objects defined in ag_catalog. The
+    -- ag_catalog objects referenced here are schema-qualified.
+    SET search_path = pg_catalog, ag_catalog
     AS $function$
 DECLARE
     graph_count integer;
@@ -108,7 +111,10 @@ COMMENT ON FUNCTION ag_catalog.age_prepare_pg_upgrade() IS
 CREATE FUNCTION ag_catalog.age_finish_pg_upgrade()
     RETURNS void
     LANGUAGE plpgsql
-    SET search_path = ag_catalog, pg_catalog
+    -- Resolve built-in functions and operators from pg_catalog first so they
+    -- are not overridden by same-named objects defined in ag_catalog. The
+    -- ag_catalog objects referenced here are schema-qualified.
+    SET search_path = pg_catalog, ag_catalog
     AS $function$
 DECLARE
     mapping_count integer;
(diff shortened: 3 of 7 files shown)


gregfelice:
Reviewed the full diff and traced the resolution/ownership changes against the actual function bodies and the control file. This is a clean, well-scoped hardening change — I'd merge it. (Read carefully; did not rebuild locally — the 38/38 installcheck is the author's.)

### Verified

- **The `search_path` flip is safe.** Changing `ag_catalog, pg_catalog` -> `pg_catalog, ag_catalog` in the four pg_upgrade helpers only matters if some unqualified ag_catalog object reference relied on ag_catalog being first. I checked all four bodies: every `ag_catalog` table is already schema-qualified (`ag_catalog.ag_graph`, `ag_catalog.ag_label`, ...), there are no unqualified AGE function/operator calls, and `_graphid_mapping` is a TEMP table (resolves via the implicit `pg_temp` entry regardless of the explicit path). So the flip can only change built-in resolution, which is the intent.
- **The install guard is reachable and effective.** With `schema = 'ag_catalog'` in `age.control` and no explicit `CREATE SCHEMA` in `age_main.sql`, PostgreSQL creates ag_catalog (owned by the installer) when it does not pre-exist, so the guard is a no-op on the common path and fires only for a *pre-existing, foreign-owned* ag_catalog — exactly the case it is meant to catch. Placing it at the top of `age_main.sql`, before any object creation, is correct.
- **Guard is correctly absent from the upgrade script.** It lives only in the fresh-install path; the upgrade template gets only the qualification/`search_path` changes, so an `ALTER EXTENSION ... UPDATE` on an existing (possibly foreign-owned) ag_catalog won't be broken by it. Good call.
- **Ownership compared by OID, not role membership** — correct for the superuser-is-implicitly-a-member-of-every-role case, and the comment documents the intent.
- Nice touch: the regress test asserts via `prosrc ~ '...'` that no unqualified `format(`/`hashtext(` calls survive.

### Suggestions (non-blocking)

1. **The guard's actual firing is untested end-to-end.** The regress test exercises the *detection predicate* (the `EXISTS`/`nspowner <>` query) in isolation, but never drives a real `CREATE EXTENSION age` against a foreign-owned ag_catalog, so the `RAISE EXCEPTION` path and its HINT are not covered. Understandable given the regress harness already has the extension installed — but a TAP test (or a scripted install into a fresh DB owned by a second role) would close the loop on the one behavior this PR actually adds.
2. **Minor:** the `[^.]\mformat\s*\(` regex requires a non-dot character before the keyword, so a `format(` at the very start of `prosrc` would slip through; in practice there is always leading whitespace, so this is negligible.

Solid work — the OID-based ownership comparison and the upgrade-path exclusion show good attention to the edge cases.

gregfelice (approved):
Approving. Clean, well-scoped hardening. I verified the two things that mattered: the `search_path` flip is safe (all ag_catalog tables in the four pg_upgrade helpers are schema-qualified, no unqualified AGE function/operator calls, temp tables resolve via pg_temp regardless), and the install guard is reachable and effective given `schema = 'ag_catalog'` in the control file — firing only for a pre-existing, foreign-owned ag_catalog, and correctly absent from the upgrade path.

Verified the cassert PG18 gate locally (CI-mirror: REL_18_STABLE --enable-cassert -O0, -Werror clean build, EXTRA_TESTS="pgvector fuzzystrmatch pg_trgm"): **All 42 tests passed**, including the new `extension_security` test, no regression.diffs.

Non-blocking: the guard's actual RAISE-on-install path is only exercised via the detection predicate, not end-to-end — a TAP/fresh-DB test would close that loop as a follow-up.

muhammadshoaib (approved):
Looks good

jrgemignani:
> Reviewed the full diff and traced the resolution/ownership changes against the actual function bodies and the control file. This is a clean, well-scoped hardening change — I'd merge it. (Read carefully; did not rebuild locally — the 38/38 installcheck is the author's.)
> 
> ### Verified
> * **The `search_path` flip is safe.** Changing `ag_catalog, pg_catalog` -> `pg_catalog, ag_catalog` in the four pg_upgrade helpers only matters if some unqualified ag_catalog object reference relied on ag_catalog being first. I checked all four bodies: every `ag_catalog` table is already schema-qualified (`ag_catalog.ag_graph`, `ag_catalog.ag_label`, ...), there are no unqualified AGE function/operator calls, and `_graphid_mapping` is a TEMP table (resolves via the implicit `pg_temp` entry regardless of the explicit path). So the flip can only change built-in resolution, which is the intent.
> * **The install guard is reachable and effective.** With `schema = 'ag_catalog'` in `age.control` and no explicit `CREATE SCHEMA` in `age_main.sql`, PostgreSQL creates ag_catalog (owned by the installer) when it does not pre-exist, so the guard is a no-op on the common path and fires only for a _pre-existing, foreign-owned_ ag_catalog — exactly the case it is meant to catch. Placing it at the top of `age_main.sql`, before any object creation, is correct.
> * **Guard is correctly absent from the upgrade script.** It lives only in the fresh-install path; the upgrade template gets only the qualification/`search_path` changes, so an `ALTER EXTENSION ... UPDATE` on an existing (possibly foreign-owned) ag_catalog won't be broken by it. Good call.
> * **Ownership compared by OID, not role membership** — correct for the superuser-is-implicitly-a-member-of-every-role case, and the comment documents the intent.
> * Nice touch: the regress test asserts via `prosrc ~ '...'` that no unqualified `format(`/`hashtext(` calls survive.
> 
> ### Suggestions (non-blocking)
> 1. **The guard's actual firing is untested end-to-end.** The regress test exercises the _detection predicate_ (the `EXISTS`/`nspowner <>` query) in isolation, but never drives a real `CREATE EXTENSION age` against a foreign-owned ag_catalog, so the `RAISE EXCEPTION` path and its HINT are not covered. Understandable given the regress harness already has the extension installed — but a TAP test (or a scripted install into a fresh DB owned by a second role) would close the loop on the one behavior this PR actually adds.
> 2. **Minor:** the `[^.]\mformat\s*\(` regex requires a non-dot character before the keyword, so a `format(` at the very start of `prosrc` would slip through; in practice there is always leading whitespace, so this is negligible.
> 
> Solid work — the OID-based ownership comparison and the upgrade-path exclusion show good attention to the edge cases.

@gregfelice ty for the feedback! Unfortunately, with our current regression tests (and maybe we should consider changing them) that functionality (1) can't be tested due to how the regression tests work. This was tested locally to verify that it worked, though.

PR #2444, opened 2026-06-22
Fix single-node labeled pattern expressions not filtering by label (#2443)

### LINKED ISSUE #2443: Single-node labeled pattern (a:Label) does not filter by label (vertex-only EXISTS skips label quals)
### Summary

Single-node labeled pattern expressions `(a:Label)` are accepted as boolean expressions but do **not** filter by label — they evaluate as trivially true. This is a pre-existing limitation in the vertex-only pattern transform, surfaced by #2360 (which lets a bare pattern appear anywhere an expression is valid).

### Reproduction

```sql
SELECT * FROM cypher('g', $$ CREATE (:Person {name:'Alice'}), (:Animal {name:'Rex'}) $$) AS (r agtype);

-- expected: only "Alice"; actual: "Alice" AND "Rex"
SELECT * FROM cypher('g', $$ MATCH (a) WHERE (a:Person) RETURN a.name $$) AS (n agtype);

-- expected: a boolean reflecting whether a has label Person; actual: always true
SELECT * FROM cypher('g', $$ MATCH (a:Person) RETURN (a:Person) $$) AS (b agtype);
```

### Root cause

`make_path_join_quals()` in `src/backend/parser/cypher_clause.c:5454` early-returns for vertex-only patterns:

```c
/* for vertex only queries, there is no work to do */
if (list_length(entities) < 3)
{
    return NIL;
}
```

The label-filter quals (`_extract_label_id(id) = label_id`) are only emitted inside `make_join_condition_for_edge()`, which is never reached without a relationship. So a single-vertex pattern desugars to an `EXISTS` subquery with no label constraint → trivially true.

### Scope

- Relationship patterns (`(a)-[:R]->(b)`) correlate and filter correctly — the path goes through the ≥3-entity branch. This is the feature #2360 targets and it is unaffected.
- The buggy early-return is byte-identical on `master`; #2360 touches no transform code. It only makes the limitation reachable via the new `(a:Label)` expression surface.

### Suggested fix

Emit the label-filter qual for single bound vertices before the early return in `make_path_join_quals` (or hoist the label-filter logic out of `make_join_condition_for_edge` so it applies to vertex-only patterns), plus a regression test asserting `WHERE (a:Person)` filters and `RETURN (a:Label)` reflects the label.

Found during review of #2360.

gregfelice:
## Problem

A single-node labeled pattern used as a boolean expression — e.g. `WHERE (a:Person)` or `WHERE EXISTS((a:Person))` — is accepted but does **not** test the bound vertex's label. It evaluates as trivially true, so the predicate matches every row.

Repro (on `master`):

```sql
SELECT * FROM cypher('g', $$ CREATE (:Person {name:'Alice'}), (:Animal {name:'Rex'}) $$) AS (r agtype);

-- returns BOTH Alice and Rex; should return only Alice
SELECT * FROM cypher('g', $$ MATCH (a) WHERE (a:Person) RETURN a.name $$) AS (n agtype);
```

## Root cause

A single-node pattern expression desugars to an `EXISTS` sub-pattern. `make_path_join_quals()` returns early for vertex-only patterns (`list_length(entities) < 3`), emitting no quals. With no edge to carry a correlation, the sub-pattern references nothing from the enclosing query, so the planner produces an **uncorrelated one-time InitPlan** that is true whenever any vertex of that label exists — independent of the outer row.

Relationship patterns are unaffected: the edge-driven join (`start_id = a.id`) correlates them to the outer variable, which is why `WHERE (a)-[:R]->(b)` works correctly.

`EXPLAIN` before the fix — uncorrelated `InitPlan` with a `One-Time Filter`:

```
InitPlan 1
  ->  Result
One-Time Filter: ((InitPlan 1).col1)::agtype
```

After the fix — correlated `SubPlan` referencing the outer `a_1`:

```
SubPlan 1
  ->  Result
        One-Time Filter: ((_extract_label_id(a_1.id))::integer = 3)
```

## Fix

In `make_path_join_quals()`, for a vertex-only pattern whose vertex carries a non-default label **and** whose variable is declared in an ancestor parse state (a correlated reference), emit an explicit label-id filter. `make_qual()` builds a name-based id reference that resolves to the outer variable, so the filter both **correlates** the sub-pattern to that variable and **enforces** the label.

Freshly scanned, non-correlated vertices (no ancestor binding) are untouched, so `MATCH (a:Person)` and "does any X exist" `EXISTS` checks behave exactly as before.

## Testing

Added regression coverage to `pattern_expression` against a graph containing a non-`Person` vertex:

- `WHERE (a:Person)` → only the `:Person` vertices
- `WHERE NOT (a:Person)` → only the non-`Person` vertex
- `WHERE EXISTS((a:Company))` → only the `:Company` vertex

All 41 regression tests pass (`make installcheck`).

## Note

`RETURN (a:Label)` in a **projection** on an unlabeled bound variable still errors with "multiple labels for variable" — that is a separate, pre-existing guard, orthogonal to this fix, and is intentionally left unchanged.

Fixes #2443

### DIFF
--- a/regress/expected/pattern_expression.out
+++ b/regress/expected/pattern_expression.out
@@ -320,12 +320,15 @@ $$) AS (result agtype);
 --
 -- Single-node pattern on an already-bound variable: (a:Label)
 --
--- NOTE: this is an EXISTS existence check on the bound variable, NOT an
--- openCypher label predicate. A matching label is therefore always true
--- (the variable is already bound), and a *different* label is rejected by
--- AGE's pre-existing "multiple labels for variable" restriction rather than
--- evaluating to false. Both behaviours are captured here so any future change
--- to single-node-pattern semantics is caught by this test.
+-- NOTE: as of #2443 a single-node labeled pattern is a correlated label
+-- predicate -- in WHERE / EXISTS it tests whether the bound vertex actually
+-- has the label (see the WHERE (a:Person) / EXISTS((a:Company)) cases in the
+-- #2443 section below). Here the variable is already bound to the SAME label,
+-- so the predicate is trivially true (the label matches). A *different* label
+-- on an already-bound variable is still rejected by AGE's pre-existing
+-- "multiple labels for variable" restriction rather than evaluating to false;
+-- that is an orthogonal limitation, captured here so any future change to
+-- single-node-pattern semantics is caught by this test.
 SELECT * FROM cypher('pattern_expr', $$
     MATCH (a:Person)
     RETURN a.name, (a:Person)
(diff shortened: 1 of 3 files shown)


gregfelice:
Ready for review.

- Fix verified in a clean PG18 build: `WHERE (a:Person)` / `EXISTS((a:Person))` now filter correctly (correlated `SubPlan` on `_extract_label_id(a.id) = <label_id>`) instead of producing a trivially-true uncorrelated one-time `InitPlan`.
- Full regression suite green (41/41), `regression.diffs` empty, including the new `pattern_expression` cases.
- Scope kept tight: only correlated bound vertices with a non-default label are affected; plain `MATCH (a:Person)` and "does any X exist" `EXISTS` checks are unchanged.
- The pre-existing single-node-pattern NOTE in `pattern_expression.sql` was updated to reflect the corrected behavior (and the orthogonal, still-present "multiple labels" projection limitation).

Fixes #2443.

jrgemignani (approved):

PR #2445, opened 2026-06-22
ci: pin Build/Regression runner to ubuntu-24.04 and guard Bison version for GLR grammar

gregfelice:
## What

Pin the **Build / Regression** CI runner image (`installcheck.yaml`) and guard the Bison version for the Cypher GLR grammar.

- `installcheck.yaml`: `runs-on: ubuntu-latest` → `ubuntu-24.04`
- Add a "Verify Bison version" step that fails with a clear, actionable message if Bison ever drifts off 3.8.x.

## Why

`src/backend/parser/cypher_gram.y` pins exact conflict counts via `%expect 7` / `%expect-rr 3`, and Bison treats `%expect` as **exact-match** (not a ceiling) — any deviation fails the build. `ubuntu-latest` floats to new Ubuntu releases, which can ship a different Bison version that reports different conflict counts, breaking the build with a cryptic grammar-generation error unrelated to the actual change.

Freezing the runner image freezes Bison at 3.8.x, so the pinned `%expect` counts stay reproducible. The version guard turns a future drift into an explicit, self-explanatory failure ("re-run bison, update `%expect`, and bump the pinned runner together") instead of a confusing one.

Reproducibility comes from pinning the variable rather than widening the conflict-count tolerance, keeping the exact-match alarm for genuinely new grammar conflicts intact. This mirrors PostgreSQL's own `gram.y`, which keeps an exact `%expect` and updates it on intentional changes.

## Scope — intentionally only Build / Regression

Only `installcheck.yaml` builds the Cypher grammar with Bison, so it is the only workflow whose result depends on the Bison version. The four driver workflows (`jdbc-driver.yaml`, `nodejs-driver.yaml`, `go-driver.yml`, `python-driver.yaml`) run their drivers against the AGE Docker image and never invoke Bison — pinning them here would freeze their runner for no functional reason (and would collide with their own runtime pins, e.g. the Node version set in `nodejs-driver.yaml`). They are intentionally left on `ubuntu-latest`.

## Testing

CI config only; no source or test changes. The existing `Build / Regression` workflow runs unchanged on the pinned image.

### DIFF
--- a/.github/workflows/installcheck.yaml
+++ b/.github/workflows/installcheck.yaml
@@ -9,7 +9,12 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    # Pinned (not ubuntu-latest) so the Bison version stays fixed at 3.8.x.
+    # The Cypher GLR grammar pins exact conflict counts via %expect / %expect-rr
+    # in src/backend/parser/cypher_gram.y, and Bison treats %expect as exact-match:
+    # a different Bison version can report different counts and break the build.
+    # Freezing the runner image freezes Bison; bump both together, intentionally.
+    runs-on: ubuntu-24.04
 
     steps:
       - name: Get latest commit id of PostgreSQL 18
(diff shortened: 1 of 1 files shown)


jrgemignani:
@gregfelice Please see Copilot review above.

gregfelice:
Thanks @jrgemignani — addressed the Copilot note in 71fe83e0.

The version guard now parses the field with `awk 'NR==1 {print $NF}'` instead of the `grep -oE` regex, and errors explicitly when the version can't be determined (instead of falling through to the confusing `Bison  != 3.8.x` message):

```sh
ver=$(bison --version | awk 'NR==1 {print $NF}')
if [ -z "$ver" ]; then
  echo "::error::Could not determine Bison version from 'bison --version'."
  exit 1
fi
```

Verified the parse against `bison (GNU Bison) 3.8.2` → `3.8.2` and empty input → caught by the guard.

jrgemignani:
@gregfelice One last Copilot comment.

jrgemignani:
@gregfelice bump

gregfelice:
@jrgemignani On the remaining Copilot note (the other workflows still on `ubuntu-latest`): that's intentional, and I've clarified it in the PR description under a new **Scope** section.

The pin exists only to freeze **Bison** so the GLR grammar's exact `%expect`/`%expect-rr` conflict counts stay reproducible — and `installcheck.yaml` (Build / Regression) is the only workflow that builds the grammar. The four driver workflows run their drivers against the AGE Docker image and never invoke Bison, so pinning them would freeze their runner for no functional reason (and would fight their own runtime pins, e.g. the Node version in `nodejs-driver.yaml`). So this PR narrows to Build / Regression by design rather than pinning all five.

Title/description updated to say so explicitly.

jrgemignani (approved):

PR #2446, opened 2026-06-22
resolve subgraph staging sequences via regclass

jrgemignani:
The vertex/edge staging copies in create_subgraph() generated new graphids with nextval(%L), which binds the sequence as a string literal and invokes the nextval(text) overload. That re-resolves the schema-qualified sequence name on each call.

Cast the literal to regclass (nextval(%L::regclass)) so the sequence is resolved once to its OID, matching how AGE defines its label id defaults (nextval('schema.seq'::regclass)). Applied to both the vertex and edge staging queries, in sql/age_subgraph.sql and the identical body in the age--1.7.0--y.y.y.sql upgrade template so the upgrade-path catalog comparison still matches.

Behavior is unchanged; all 38 regression tests pass against PostgreSQL 18.

Addresses Copilot review feedback on apache/age#2441.

Co-authored-by: GitHub Copilot (Claude Opus 4.8) <[email protected]>

modified:   sql/age_subgraph.sql

### DIFF
--- a/age--1.7.0--y.y.y.sql
+++ b/age--1.7.0--y.y.y.sql
@@ -1013,7 +1013,7 @@ BEGIN
         EXECUTE format(
             'CREATE TEMP TABLE _ag_sg_vstage ON COMMIT DROP AS '
             'SELECT t.id AS old_id, '
-            '       ag_catalog._graphid(%s, nextval(%L)) AS new_id, '
+            '       ag_catalog._graphid(%s, nextval(%L::regclass)) AS new_id, '
             '       t.properties AS props '
             'FROM ONLY %s t '
             'WHERE EXISTS (SELECT 1 FROM _ag_sg_kept_v k WHERE k.gid = t.id)',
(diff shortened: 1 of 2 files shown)


gregfelice (approved):
LGTM — approving.

Confirmed the change is behavior-preserving: `%L` is fed `dst_seq_fqn` (the fully-qualified sequence name), so `nextval(%L::regclass)` yields `nextval('schema.seq'::regclass)`, matching how AGE defines its label-id defaults. Both the old `nextval(text)` overload and `'...'::regclass` resolve the name to an OID through the same identifier-parsing path; the cast just resolves the constant once at plan time instead of re-resolving per row via the text overload. Same semantics, less per-row work.

Also verified the identical change is applied to both `sql/age_subgraph.sql` and the `age--1.7.0--y.y.y.sql` upgrade template, so the upgrade-path catalog comparison still matches. Existing `regress/sql/subgraph.sql` exercises `create_subgraph`, and the change is behavior-preserving, so no new test is needed.

Issue #2449, opened 2026-06-29
load_labels_from_file and load_edges_from_file functions undocumented use of comma as delimiter via PostgreSQL COPY default, causes data corruption on labels and segfault on edges


AyoubKaz07:
**Describe the bug**
Both functions rely on Postgres's COPY command, `ag_load_labels.c` and `ag_load_edges.c` share an identical `create_copy_options()` that only passes two options to `BeginCopyFrom` (header and format, everything else is default from Postgres).

`load_labels_from_file` with a pipe-delimited multi-column label CSV file silently corrupts data (correctly since default is comma). The entire line is read as a single field, column names get merged with values into one token, and the result lands in properties as garbage with no error:
```pgsql
id        |                 properties
----------+---------------------------------------------
...       | {"__id__": 1, "id|islocatedin_city": "1|5"}
```

`load_edges_from_file` with a pipe-delimited file crashes the server. The 4-column edge format requires `start_id, start_vertex_type, end_id, end_vertex_type` as separate fields. A quick lookup on it shows that with no comma splitting, fields[1] through fields[3] are null, `process_edge_row` calls `trim_whitespace(fields[1])` on a bad pointer, and the server segfaults.

The documentation mentions no delimiter requirement. The example files use comma but this is never stated as a constraint. The header row is handled manually in the loop rather than by COPY itself `(HEADER false)`, so the parsing feels custom, which makes it even less obvious that COPY's delimiter default is silently in effect underneath.

**How are you accessing AGE (Command line, driver, etc.)?**
- psql

**What is the command that caused the error?**
```pgsql
SELECT load_labels_from_file('graph', 'Label', '/derived/Label.csv'); -- No error, just incorrect parsing.

SELECT load_edges_from_file('graph', 'SomeEdge', '/derived/SomeEdge.csv'); -- This segfaults
```
```
026-06-29 19:25:56.972 GMT [1] LOG:  client backend (PID 64) was terminated by signal 11: Segmentation fault
2026-06-29 19:25:56.972 GMT [1] DETAIL:  Failed process was running: SELECT load_edges_from_file('graph', 'SomeEdge', '/derived/SomeEdge.csv');
2026-06-29 19:25:56.972 GMT [1] LOG:  terminating any other active server processes
```

**Expected behavior**
Depends really, either have a delimiter option just like postgres COPY command options or clarify in documentation + correct error if comma only (instead of segfaulting).

**Environment (please complete the following information):**
- apache/age docker image.

[...]

PR #2453, opened 2026-07-02
Fix segfault and out-of-bounds reads in file loaders on malformed rows

gregfelice:
## Summary

Fixes the segfault reported in #2449 (and a related out-of-bounds read) in `load_edges_from_file()` / `load_labels_from_file()`.

Both loaders build their COPY parser with only `format=csv` + `header=false`, so COPY uses its **default comma delimiter**. A file delimited by anything else — or any malformed row — then parses with an unexpected column count, and the loaders indexed the parsed fields **without validating that count**:

- `process_edge_row()` reads the four fixed fields `fields[0..3]` unconditionally. A non-comma-delimited edge file parses as a single column, so `fields[1..3]` are out of bounds → **segfault**.
- `create_agtype_from_list()`/`_i()` pair `header[i]` with `fields[i]` for all `i < nfields`, so a row with **more** fields than the header reads `header[i]` out of bounds.

## Fix

Bounds validation that turns these into clear errors instead of a crash/UB:

- **Edge header** must have ≥ 4 columns; a smaller count almost always means the wrong delimiter, so the error carries an `errhint` pointing at the delimiter.
- **Each edge row** must have ≥ 4 columns and no more than the header's count.
- **Each label row** must have no more than the header's count.

Rows with *fewer* trailing columns than the header remain allowed (that's safe — the property loop only iterates `nfields` — and is exercised by the existing conversion tests).

Example, before vs after, on a pipe-delimited edge file:
```
-- before: server terminated by signal 11 (segfault)
-- after:
ERROR:  edge file must have at least 4 columns (start_id, start_vertex_type, end_id, end_vertex_type), but the header has 1
HINT:  load_edges_from_file expects a comma-delimited CSV; check the file's delimiter.
```

## Scope / follow-up

This closes the **segfault and out-of-bounds reads**. The *silent mis-parsing* of a non-comma file whose header and rows share the same (wrong) column count (the label-corruption half of #2449) isn't detectable here — the real fix for that is a **delimiter option** on the load functions, which I'll do as a separate follow-up. Leaving #2449 open for that.

## Tests

Adds a regression case to `age_load` using a pipe-delimited edge file (asserts the clean error). Full regression suite passes locally against PostgreSQL 18 (41/41).

### DIFF
--- a/regress/age_load/data/bad_delim_edges.csv
+++ b/regress/age_load/data/bad_delim_edges.csv
@@ -1,0 +1,2 @@
+start_id|start_vertex_type|end_id|end_vertex_type
+1|V|2|V
--- a/regress/age_load/data/edges_long_row.csv
+++ b/regress/age_load/data/edges_long_row.csv
@@ -1,0 +1,2 @@
+start_id,start_vertex_type,end_id,end_vertex_type
+1,V,2,V,extra
--- a/regress/age_load/data/edges_short_row.csv
+++ b/regress/age_load/data/edges_short_row.csv
@@ -1,0 +1,2 @@
+start_id,start_vertex_type,end_id,end_vertex_type
+1,V
--- a/regress/age_load/data/labels_long_row.csv
+++ b/regress/age_load/data/labels_long_row.csv
@@ -1,0 +1,2 @@
+id,name
+1,Alice,extra
--- a/regress/expected/age_load.out
+++ b/regress/expected/age_load.out
@@ -455,6 +455,49 @@ NOTICE:  graph "agload_conversion" has been dropped
 (1 row)
 
 --
+-- Issue 2449: mis-delimited / malformed load files must fail with a clear
+-- error instead of segfaulting or silently corrupting data. Edge files
+-- require the 4 fixed columns; a file that is not comma-delimited parses as
+-- a single column, so this must be rejected at the header.
+--
+SELECT create_graph('agload_delim');
+NOTICE:  graph "agload_delim" has been created
+ create_graph 
+--------------
+ 
+(1 row)
+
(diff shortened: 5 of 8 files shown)


gregfelice:
@jrgemignani CI is green across all five workflows (Build/Regression + all four drivers). Ready for review whenever you have a minute.

This closes the segfault + out-of-bounds reads in #2449: the loaders indexed COPY-parsed fields without validating the column count, so a non-comma-delimited edge file (parsed as one column) read `fields[1..3]` out of bounds and crashed. The fix adds bounds guards that turn it into a clear error (with a hint about the delimiter). Small, in-lane, regression test included in `age_load`.

Scope note: this closes the crash/OOB half; the silent mis-parse of a non-comma file (the label-corruption half) needs a delimiter option, which I'll do as a follow-up — so #2449 stays open for that. Small enough to consider for 1.8.0 if you want the crash fix in.

jrgemignani:
@gregfelice please see above Copilot suggestions.

gregfelice:
@jrgemignani addressed the Copilot notes in a431fc55:

- **Error wording:** both per-row messages now say "the header's %d **columns**" so the count's unit is explicit.
- **Missing regression coverage:** added cases for the per-row guards (previously only the mis-delimited-header path was asserted):
  - an edge row with fewer than 4 columns → `edge file row has 2 columns; expected at least 4 and no more than the header's 4 columns`
  - an edge row with more columns than the header → `...has 5 columns; ...header's 4 columns`
  - a label row with more columns than the header → `label file row has 3 columns, more than the header's 2 columns`

Each asserts a clean ERROR — these are exactly the out-of-bounds reads the guards now catch. Full regression suite green locally (41/41 on PG18).

jrgemignani (approved):

PR #2454, opened 2026-07-02
Add automatic header-dependency tracking to the Makefile

jrgemignani:
AGE lists OBJS explicitly and relies on PGXS, whose built-in dependency tracking only runs when the server was built with --enable-depend (often disabled). Consequently, editing a header did not recompile the .c files that include it, leaving stale .o files. This is especially dangerous for node/struct headers: a stale ag_nodes.o keeps an outdated node_size, so _readExtensibleNode under-allocates and readNode corrupts the heap ("unrecognized node type").

Emit a .d file beside each object via -MMD -MP and -include them, deriving DEPFILES from OBJS. The mechanism is self-contained (independent of --enable-depend): -MMD skips system headers and -MP tolerates deleted headers. On servers built with --enable-depend, PGXS appends its own -MF after CFLAGS (last -MF wins), so this degrades cleanly to PGXS's tracking. Add DEPFILES to EXTRA_CLEAN and *.d to .gitignore.

Co-authored-by: Copilot <copilot@github.com>

modified:   .gitignore
modified:   Makefile

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *.o
+*.d
 *.so
 build.sh
 .idea
--- a/Makefile
+++ b/Makefile
@@ -162,6 +162,10 @@ OBJS = src/backend/age.o \
        src/backend/utils/name_validation.o \
        src/backend/utils/ag_guc.o
 
+# Per-object header-dependency files (see "Automatic header-dependency
+# tracking" below the PGXS include). One .d is generated beside each .o.
+DEPFILES = $(OBJS:.o=.d)
+
 # ===== Extension SQL & data files =====
 EXTENSION = age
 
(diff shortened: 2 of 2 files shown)


gregfelice (approved):
LGTM — approving.

Verified the incremental-rebuild behavior end-to-end on a local build of this branch:
- Clean build emits one `.d` per object (53 files); `.d` target lines are correct full paths.
- No-op `make` after building recompiles nothing.
- Touching a node header (`cypher_nodes.h`) recompiles exactly its dependents — including `age.o` (confirmed by mtime; a `-MMD` capture regex initially masked it, but the object does rebuild).
- Touching a narrowly-included header (`ag_load_edges.h`) recompiles only the two loader objects — no over-rebuild.

Mechanism does what the description claims, and degrades cleanly to PGXS's `.deps` on `--enable-depend` servers. Build-tooling only; all five CI checks green.

Minor, non-blocking: dependency tracking covers `.o` (via CFLAGS) but not the `.bc` bitcode objects — the `.o` is the one that links into `age.so`, so this is fine; a `.bc` follow-up is optional.

PR #2458, opened 2026-07-03
Fix segfault on CREATE/MERGE/SET with generated or extra label columns (#2450)

### LINKED ISSUE #2450: [Segmentation Fault] when creating graph node whose vertex label table has GENERATED ALWAYS columns
**Describe the bug**
Adding a GENERATED ALWAYS AS (...) STORED column to a vertex label table (_ag_label_vertex child table) causes a segmentation fault (signal 11) during Cypher CREATE / SET operations.
The crash occurs in the vertex creation path (likely during heap_insert in the executor).

Directly via cypher() SQL function (IN psql client).

```pgsql
-- 1. Create graph and label
SELECT create_graph('age_bug');
SELECT create_vlabel('age_bug', 'Product');

-- 2. Add generated column (this triggers the bug)
ALTER TABLE age_bug."Product"
	ADD COLUMN category VARCHAR(25) GENERATED ALWAYS AS (agtype_access_operator(properties, '"category"')) STORED;
```

- [e.g. Installed PostGIS]

```pgsql
SELECT * FROM cypher('age_bug', $$
   CREATE (p:Product {category: 'disk', type: 'M1234', version: 1})
   RETURN p
$$) AS (p agtype);
```
```
ERROR: 
psql client log: 
SQL Error [08006]: An I/O error occurred while sending to the backend.

server log: 
LOG:  server process (PID XXX) was terminated by signal 11: Segmentation fault
DETAIL:  Failed process was running: [the cypher CREATE query]
LOG:  terminating any other active server processes
FATAL:  the database system is in recovery mode
FATAL:  the database system is in recovery mode
LOG:  all server processes terminated; reinitializing
LOG:  database system was interrupted; last known up at  XXXX
LOG:  database system was not properly shut down; automatic recovery in progress
LOG:  invalid record length at 5716/4C635DA8: wanted 24, got 0
LOG:  redo is not required
LOG:  checkpoint starting: end-of-recovery immediate wait
LOG:  checkpoint complete: wrote 2 buffers (0.0%); 0 WAL file(s) added, 0 removed, 0 recycled; write=0.006 s, sync=0.003 s, total=0.017 s; sync files=3, longest=0.001 s, average=0.001 s; distance=0 kB, estimate=0 kB
LOG:  database system is ready to accept connections
FATAL:  database "template0" is not currently accepting connections
LOG:  background worker "autoprewarm worker" (PID XXX) exited with exit code 1

```

*Even though MATCH query does NOT RAISE ANY error*
```pgsql
SELECT * FROM cypher('age_bug', $$
   MATCH (p:Product {category: 'disk', type: 'M1234', version: 1})
   RETURN p
$$) AS (p agtype);
```

**Expected behavior**
AGE should support standard PostgreSQL features like generated columns on label TABLES

- AGE Version: release/PG15/1.5.0
- PostgreSQL Version: PostgreSQL 15.8 (Debian 15.8-1.pgdg100+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 8.3.0-6) 8.3.0, 64-bit
- OS: Linux

**Additional context**

gregfelice:
## Problem

Fixes #2450. A `GENERATED ALWAYS ... STORED` column on a label table caused a backend segfault (signal 11) on Cypher `CREATE` / `MERGE` / `SET`:

```sql
SELECT create_graph('age_bug');
SELECT create_vlabel('age_bug', 'Product');
ALTER TABLE age_bug."Product"
  ADD COLUMN category varchar(25)
  GENERATED ALWAYS AS (agtype_access_operator(properties, '"category"')) STORED;
-- crashes the backend:
SELECT * FROM cypher('age_bug', $$ CREATE (p:Product {category:'disk'}) RETURN p $$) AS (p agtype);
```

## Root cause

AGE builds the insert/update tuple slot from the **full** label-table descriptor (`table_slot_create`), but only populates the columns it manages — `id`/`properties` for vertices, `id`/`start_id`/`end_id`/`properties` for edges. Any user-added column is left as an uninitialized slot entry. `ExecStoreVirtualTuple` then marks all attributes valid, so `heap_form_tuple()` → `heap_compute_data_size()` reads the uninitialized `tts_isnull[]`/`tts_values[]` and segfaults dereferencing garbage as a varlena.

```
#0 heap_compute_data_size
#1 heap_form_tuple
#3 ExecFetchSlotHeapTuple
#4 insert_entity_tuple_cid   cypher_utils.c
#6 create_vertex            cypher_create.c
```

The SET path crashes identically via `update_entity_tuple` (`cypher_set.c`). It affects both stored generated columns and plain `ADD COLUMN`s.

## Fix

1. **`clear_entity_slot()`** clears the slot and marks every attribute NULL before AGE fills in its own columns, so columns AGE doesn't manage default to NULL instead of stale slot memory. Used at all six slot-build sites: `create_vertex`, `create_edge`, `merge_vertex`, `merge_edge`, and the `populate_vertex_tts` / `populate_edge_tts` helpers (SET).
2. **`ExecComputeStoredGenerated()`** is called in `insert_entity_tuple_cid` (CREATE/MERGE, `CMD_INSERT`) and `update_entity_tuple` (SET, `CMD_UPDATE`) before the heap tuple is materialized when the relation has stored generated columns, so those columns are computed on insert and recomputed on update per PostgreSQL semantics.

## Testing

New `generated_columns` regression test covering stored generated columns on vertex **and** edge labels across CREATE / MERGE / SET (values verified computed and recomputed), plus a plain user column that must not crash. Passes under `pg_regress`.

Verified against the reporter's exact repro on PG18: no crash, and `category` correctly computes (`disk` on CREATE) and recomputes (`ssd` after SET).

## Note

On `SET`, AGE rebuilds the row from `id`/`properties`, so a **generated** column recomputes correctly but a **plain** (non-generated) column is reset to NULL. Preserving plain-column values across updates would be a larger change and is out of scope here; the supported pattern is a `GENERATED ... STORED` column derived from `properties`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -223,6 +223,7 @@ REGRESS = scan \
           age_reduce \
           map_projection \
           direct_field_access \
+          generated_columns \
           security \
           reserved_keyword_alias \
           agtype_jsonb_cast \
--- a/regress/expected/generated_columns.out
+++ b/regress/expected/generated_columns.out
@@ -1,0 +1,222 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+/*
+ * Regression test for issue #2450.
+ *
+ * A GENERATED ALWAYS ... STORED column (or any user-added column) on a label
+ * table adds an attribute AGE's CREATE/MERGE/SET paths do not populate. The
+ * slot's tuple descriptor is the full relation descriptor, so heap_form_tuple()
+ * used to read the uninitialized slot entry and segfault. AGE now null-inits the
+ * entity slot and computes stored generated columns before materializing the
+ * tuple. Generated columns must be computed on insert and recomputed on update;
+ * plain user columns must default to NULL rather than crash.
+ */
+LOAD 'age';
+SET search_path TO ag_catalog;
+SELECT create_graph('generated_columns');
+NOTICE:  graph "generated_columns" has been created
+ create_graph 
+--------------
+ 
+(1 row)
+
+--
+-- GENERATED ALWAYS ... STORED column on a vertex label
+--
+SELECT create_vlabel('generated_columns', 'Product');
+NOTICE:  VLabel "Product" has been created
+ create_vlabel 
+---------------
+ 
+(1 row)
+
+ALTER TABLE generated_columns."Product"
+    ADD COLUMN category varchar(25)
+    GENERATED ALWAYS AS (agtype_access_operator(properties, '"category"')) STORED;
+-- CREATE: the generated column is computed from the new properties
+SELECT * FROM cypher('generated_columns', $$
+    CREATE (p:Product {category: 'disk', type: 'M1234'}) RETURN p
+$$) AS (p agtype);
+                                                    p                                                     
+----------------------------------------------------------------------------------------------------------
+ {"id": 844424930131969, "label": "Product", "properties": {"type": "M1234", "category": "disk"}}::vertex
+(1 row)
+
+-- CREATE where the generation expression yields NULL (no such key)
+SELECT * FROM cypher('generated_columns', $$
+    CREATE (p:Product {type: 'no-cat'}) RETURN p
+$$) AS (p agtype);
+                                           p                                           
+---------------------------------------------------------------------------------------
+ {"id": 844424930131970, "label": "Product", "properties": {"type": "no-cat"}}::vertex
+(1 row)
+
+-- MERGE also computes the generated column
+SELECT * FROM cypher('generated_columns', $$
+    MERGE (p:Product {category: 'merged'}) RETURN p
+$$) AS (p agtype);
+                                             p                                             
+-------------------------------------------------------------------------------------------
+ {"id": 844424930131971, "label": "Product", "properties": {"category": "merged"}}::vertex
+(1 row)
+
+-- MERGE again on the same pattern must MATCH (not re-insert); this exercises the
+-- matched path where the slot is intentionally left uncleared (issue #2450)
+SELECT * FROM cypher('generated_columns', $$
+    MERGE (p:Product {category: 'merged'}) RETURN p
+$$) AS (p agtype);
+                                             p                                             
+-------------------------------------------------------------------------------------------
+ {"id": 844424930131971, "label": "Product", "properties": {"category": "merged"}}::vertex
+(1 row)
+
+-- SET recomputes the generated column from the updated properties
+SELECT * FROM cypher('generated_columns', $$
+    MATCH (p:Product {category: 'disk'}) SET p.category = 'ssd', p.note = 'x' RETURN p
+$$) AS (p agtype);
+                                                          p                                                           
+----------------------------------------------------------------------------------------------------------------------
+ {"id": 844424930131969, "label": "Product", "properties": {"note": "x", "type": "M1234", "category": "ssd"}}::vertex
+(1 row)
+
+-- The stored generated column always mirrors properties.category
+SELECT category, properties FROM generated_columns."Product" ORDER BY category NULLS LAST;
+ category |                    properties                     
+----------+---------------------------------------------------
+ "merged" | {"category": "merged"}
+ "ssd"    | {"note": "x", "type": "M1234", "category": "ssd"}
+          | {"type": "no-cat"}
+(3 rows)
+
+--
+-- GENERATED ALWAYS ... STORED column on an edge label
+--
+SELECT create_elabel('generated_columns', 'REL');
+NOTICE:  ELabel "REL" has been created
+ create_elabel 
+---------------
+ 
+(1 row)
+
+ALTER TABLE generated_columns."REL"
+    ADD COLUMN kind varchar(25)
+    GENERATED ALWAYS AS (agtype_access_operator(properties, '"kind"')) STORED;
+SELECT * FROM cypher('generated_columns', $$
+    CREATE (:Product {category: 'a'})-[r:REL {kind: 'link'}]->(:Product {category: 'b'})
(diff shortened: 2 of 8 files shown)


gregfelice:
@jrgemignani ready for review when you have a cycle. This fixes the #2450 segfault where a `GENERATED ALWAYS ... STORED` (or any extra) column on a label table crashed the backend on CREATE/MERGE/SET — root cause and the two-part fix (null-init the entity slot + `ExecComputeStoredGenerated`) are in the description. New `generated_columns` regression test covers vertex + edge across all three write paths, and CI is green.

One design note called out in the PR: on SET, AGE rebuilds the row from `id`/`properties`, so a *generated* column recomputes correctly but a *plain* column resets to NULL — I scoped plain-column preservation out. Happy to adjust if you'd rather handle that here.

MuhammadTahaNaveed:
@gregfelice 

**GPT 5.5 xhigh review:**

  **Findings**

  1. **[P1] Stored generated columns using `tableoid` are computed incorrectly**
     - `src/backend/executor/cypher_utils.c:354`
     - `src/backend/executor/cypher_set.c:143`

     The new `ExecComputeStoredGenerated(...)` calls compute stored generated columns, but the tuple slot does not appear to have `tts_tableOid` set first. PostgreSQL’s own executor sets
     `slot->tts_tableOid = RelationGetRelid(...)` before computing stored generated columns.

     Repro: adding this to a vertex label produces `InvalidOid` instead of the label table OID:

     ```sql
     ALTER TABLE tableoid_case."T"
       ADD COLUMN tbl oid GENERATED ALWAYS AS (tableoid) STORED;

  Then CREATE and SET through Cypher both leave tbl as - / invalid. The fix should set elemTupleSlot->tts_tableOid before ExecComputeStoredGenerated(...) in both insert/create and update/set
  paths, and add a regression case for tableoid.

  2. **[P1] PROJECT.yaml adds incorrect Apache project metadata**
      - PROJECT.yaml:16 says license: Proprietary
      - PROJECT.yaml:22 says owner: Rizlabs
      - PROJECT.yaml:25 points at a Rizlabs repo
      **This file is completely unrelated to the fix and should be removed immediately.**

  3. **[P2] Branch needs a rebase onto current master**
     The PR is based on merge base 12e2a31c, while current apache/master is d84b4578. It is missing several upstream commits, so the branch-local regression suite is stale. Please rebase so
     CI runs against the current tree.

  4. **[P3] git diff --check fails on trailing whitespace**
     The new regress/expected/generated_columns.out has multiple trailing-whitespace lines and a blank line at EOF. Please clean that up so whitespace checks pass.

MuhammadTahaNaveed:
@gregfelice 

**GPT 5.5 xhigh review**:

Found one performance issue worth commenting on:

Performance note: `SET` now calls `ExecComputeStoredGenerated(...)` for stored generated columns, but this path creates a fresh `ResultRelInfo` for each updated entity at `src/backend/executor/cypher_set.c:607`.

  `ExecComputeStoredGenerated(...)` lazily initializes `ri_GeneratedExprsU`; PostgreSQL stores those expression states in the query context. Because AGE’s `SET` path uses a new  `ResultRelInfo` per row/update item, multi-row `SET` on a label with stored generated columns will repeatedly rebuild generated-column expression state and retain that memory until query end.

  Relevant spots:

  - `src/backend/executor/cypher_set.c:607` creates a new `ResultRelInfo`
  - `src/backend/executor/cypher_set.c:143` / `:146` calls `ExecComputeStoredGenerated(...)`
  - `src/backend/executor/cypher_set.c:840` / `:841` closes indexes/relation, but the generated expression state has already been allocated in the executor query context

  This should probably cache/reuse `ResultRelInfo` per label/relation for the duration of `apply_update_list`, similar to the existing relation/index/RLS caching pattern. That would let
  `ri_GeneratedExprsU` initialize once per relation instead of once per updated row.

  **Secondary, lower-priority nit:**

  Minor: `MERGE` now calls `clear_entity_slot(...)` before checking whether the matched path will actually insert.

  - `src/backend/executor/cypher_merge.c:1202`
  - `src/backend/executor/cypher_merge.c:1529`

  On a matched `MERGE` where `should_insert == false`, the code later skips `insert_entity_tuple(...)`, so the new full-null-bitmap `memset` is avoidable work. Usually tiny for AGE’s base
  label tables, but it scales with extra user-added columns. Consider moving the slot clear closer to the insert-only path if practical.

gregfelice:
Thanks for the thorough review, @MuhammadTahaNaveed — genuinely useful. Addressed in the latest push:

**1. [P1] tableoid generated columns** — Fixed. Good catch. The slot now carries `tts_tableOid = RelationGetRelid(...)` before `ExecComputeStoredGenerated()` in both the insert (`cypher_utils.c`) and update (`cypher_set.c`) paths, matching PostgreSQL's own ExecInsert/ExecUpdate. Added a regression case with `tbl oid GENERATED ALWAYS AS (tableoid) STORED` that asserts the stored value equals the label table's real OID on both CREATE and SET.

**2. [P1] PROJECT.yaml** — Removed. That was a stray commit that rode in from a local branch base; the branch has been rebased onto current master so the PR now contains only the fix.

**3. [P3] Rebase** — Done, the branch is rebased onto current master and CI is green against the current tree.

**B. [P3] MERGE clears slot before the should_insert check** — Fixed. `clear_entity_slot()` is now gated on `should_insert` in both `merge_vertex` and `merge_edge`, so the matched path skips the null-init. Added a matched-MERGE regression case to cover that path.

**A. [perf] SET builds a ResultRelInfo per row** — Agreed this is worth doing, but I'd like to keep it out of this bugfix. The per-row `ResultRelInfo` is pre-existing structure, and the generated-column recompute is guarded by `has_generated_stored`, so only labels that actually have a stored generated column pay the cost — no regression for the common case. Caching `ResultRelInfo`/`ri_GeneratedExprsU` per label across `apply_update_list` is a cleaner standalone change (alongside the existing relation/index/RLS caching); filed as #2459 so it doesn't get lost.

**4. [P3] Trailing whitespace in the .out** — Left as-is intentionally. pg_regress does an exact byte comparison against psql's output, which emits trailing spaces on table rows; every expected file in `regress/expected/` has them (e.g. `cypher_set.out`). Stripping them would make the expected file no longer match psql's actual output and break the test, and there's no `diff --check`/whitespace gate in CI. So this one I have to decline.

Thanks again — the tableoid case in particular was a real gap.

MuhammadTahaNaveed (approved):

gregfelice:
Thanks for the careful review and the approval, @MuhammadTahaNaveed — the tableoid catch in particular was a real gap. All feedback is addressed, CI is green, and the follow-up perf item is tracked in #2459.

Since this is my PR I won't self-merge — feel free to land it whenever you have a moment, or I can leave it for another committer. Thanks again!

MuhammadTahaNaveed:
> Thanks for the careful review and the approval, @MuhammadTahaNaveed — the tableoid catch in particular was a real gap. All feedback is addressed, CI is green, and the follow-up perf item is tracked in #2459.
> 
> Since this is my PR I won't self-merge — feel free to land it whenever you have a moment, or I can leave it for another committer. Thanks again!
> 

@gregfelice Its already merged :)

PR #2468, opened 2026-07-07
support openCypher-like relationsiop-type edge alternation

eeshwarg:
Apache AGE's MATCH grammar rejects openCypher's relationship-type alternation syntax - `MATCH ()-[:A|B]->()` fails at parse time with `syntax error at or near "|"`. For example, a query like `SELECT * FROM cypher('kg_s7', $$ MATCH (p)-[:WORKS_AT|REPORTS_TO]->(c) RETURN c $$) AS (c agtype);` throws a syntax error. A workaround for this today is to run something like `SELECT * FROM cypher('kg_s7', $$ MATCH (p)-[r]->(c) WHERE type(r) IN ['WORKS_AT','REPORTS_TO'] RETURN c $$) AS (c agtype);`.

This change enables support for relationship-type alternation in accordance with the openCypher standard. Changes are made in the grammar, AST and during transformation from the AST to Postgres `Query` struct by adding the types to the `quals` list.

Variable-length relationship patterns with alternation `([:A|B*1..2])` are explicitly rejected with `ERRCODE_FEATURE_NOT_SUPPORTED`.

Testing:
- Added regress test with cases with directed/undirected edges, and variable length edges

Future work:
- openCypher also supports node-type alternation (like n:Person|Customer), which could be added to AGE in a future PR

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -199,6 +199,7 @@ REGRESS = scan \
           expr \
           cypher_create \
           cypher_match \
+          cypher_match_rel_label_alternation \
           cypher_unwind \
           cypher_set \
           cypher_remove \
--- a/regress/expected/cypher_match_rel_label_alternation.out
+++ b/regress/expected/cypher_match_rel_label_alternation.out
@@ -1,0 +1,185 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ *
+ * Tests for relationship-type alternation in MATCH patterns:
+ * `[:A|B|C]` and the equivalent `[:A|:B|:C]` openCypher form.
+ */
+LOAD 'age';
+SET search_path TO ag_catalog;
+SELECT create_graph('rel_label_alt');
+NOTICE:  graph "rel_label_alt" has been created
+ create_graph 
+--------------
+ 
+(1 row)
+
+SELECT create_elabel('rel_label_alt', 'WORKS_AT');
+NOTICE:  ELabel "WORKS_AT" has been created
+ create_elabel 
+---------------
+ 
+(1 row)
+
+SELECT create_elabel('rel_label_alt', 'REPORTS_TO');
+NOTICE:  ELabel "REPORTS_TO" has been created
+ create_elabel 
+---------------
+ 
+(1 row)
+
+SELECT create_elabel('rel_label_alt', 'OWNS');
+NOTICE:  ELabel "OWNS" has been created
+ create_elabel 
+---------------
+ 
+(1 row)
+
+SELECT * FROM cypher('rel_label_alt', $$
+    CREATE (a:Person {n:'alice'})-[:WORKS_AT]->(:Company {n:'acme'}),
+           (a)-[:REPORTS_TO]->(:Person {n:'bob'}),
+           (a)-[:OWNS]->(:Asset {n:'car'})
+$$) AS (v agtype);
+ v 
+---
+(0 rows)
+
+-- Two-label alternation. Returns two rows ordered by destination name.
+SELECT * FROM cypher('rel_label_alt', $$
+    MATCH (p)-[:WORKS_AT|REPORTS_TO]->(c) RETURN c.n ORDER BY c.n
+$$) AS (n agtype);
+   n    
+--------
+ "acme"
+ "bob"
+(2 rows)
+
+-- Three-label alternation. All three outgoing edges match.
+SELECT * FROM cypher('rel_label_alt', $$
+    MATCH (p)-[:WORKS_AT|REPORTS_TO|OWNS]->(c) RETURN c.n ORDER BY c.n
+$$) AS (n agtype);
+   n    
+--------
+ "acme"
+ "bob"
+ "car"
+(3 rows)
+
+-- Redundant-colon openCypher form [:A|:B] is equivalent to [:A|B].
+SELECT * FROM cypher('rel_label_alt', $$
+    MATCH (p)-[:WORKS_AT|:REPORTS_TO]->(c) RETURN c.n ORDER BY c.n
+$$) AS (n agtype);
+   n    
+--------
+ "acme"
+ "bob"
+(2 rows)
+
+-- Single-label edge is unchanged (regression check).
+SELECT * FROM cypher('rel_label_alt', $$
+    MATCH (p)-[:WORKS_AT]->(c) RETURN c.n
+$$) AS (n agtype);
+   n    
+--------
+ "acme"
+(1 row)
+
+-- Edge variable + multi-label, projecting type(r).
+SELECT * FROM cypher('rel_label_alt', $$
+    MATCH (p)-[r:WORKS_AT|REPORTS_TO]->(c)
+    RETURN type(r) AS t, c.n AS n ORDER BY n
+$$) AS (t agtype, n agtype);
+      t       |   n    
+--------------+--------
+ "WORKS_AT"   | "acme"
+ "REPORTS_TO" | "bob"
+(2 rows)
+
+-- Unknown labels in the alternation are silently ignored — only the
+-- valid label contributes rows.
+SELECT * FROM cypher('rel_label_alt', $$
+    MATCH (p)-[:WORKS_AT|DOES_NOT_EXIST]->(c) RETURN c.n
+$$) AS (n agtype);
+   n    
+--------
+ "acme"
+(1 row)
+
+-- All labels in the alternation are unknown — yields zero rows, not an
+-- error.
+SELECT * FROM cypher('rel_label_alt', $$
+    MATCH (p)-[:DOES_NOT_EXIST|ALSO_MISSING]->(c) RETURN c.n
+$$) AS (n agtype);
+ n 
+---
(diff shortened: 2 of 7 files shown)


gregfelice:
Hitting it tomorrow

jrgemignani:
@eeshwarg Please look at Copilot's suggestions. Additionally, how could this impact performance?

eeshwarg:
> @eeshwarg Please look at Copilot's suggestions. Additionally, how could this impact performance?

In short, this shouldn't affect performance of already supported queries, but only adds support for a new query shape. 

All the code changes are during plan generation, not execution time. Existing query shapes take an unchanged code path - the multi-label logic is gated behind list_length(labels) > 1, which is false for every single-label/unlabeled pattern, so they produce an identical Query tree and therefore an identical plan. The new path adds a single integer IN filter on a scan that already existed, with all label resolution done at plan time.

gregfelice (approved):
Reviewed and tested this end-to-end against PostgreSQL 18.4. The implementation is correct — approving with comments. Three things below, one of which answers @jrgemignani's performance question with numbers.

## What I verified

Built the branch and ran the **full 42-test regress suite: all 42 pass**, including the new `cypher_match_rel_label_alternation`. I ran unmodified `master` through the identical harness as a control to be sure nothing I saw was environmental.

**Grammar is conflict-clean.** Worth stating explicitly since the Makefile passes `-Wno-error=conflicts-sr/-rr` to bison and a new conflict would not fail the build on that flag alone. The branch measures exactly 7 shift/reduce + 3 reduce/reduce — identical to the `%expect 7` / `%expect-rr 3` declared on master for the GLR parser, and both directives are untouched. Since bison treats `%expect` as an exact count, the build would have hard-errored had this added even one conflict. So the new `rel_labels` rules introduce none.

**Paths not covered by the new tests, which I probed by hand and which all behave correctly:**
- `EXISTS((a)-[:A|B]->(b))` and a pattern expression in `RETURN` — both filter correctly (the qual does reach the pattern-expression transform).
- `OPTIONAL MATCH (a)-[:A|B]->(b)` — filters correctly.
- The VLE guard catches all four variable-length forms: `*`, `*2`, `*1..2`, `*0..`. Nothing slips through.
- Duplicate-edge qual, edge properties (`[:A|B {k:1}]`), edge variables and `type(r)` all work under alternation.
- `[:A|A]` and unknown/vertex labels inside the alternation degrade sanely.
- Unknown label ignored, all-unknown → zero rows, is the right openCypher semantics.

Also confirming the copy/read funcs are correctly left alone: `cypher_relationship` is registered with `DEFINE_NODE_METHODS`, whose `copy_ag_node`/`read_ag_node` throw, so it is never copied or read — touching only `cypher_outfuncs.c` is right, not an omission.

On Copilot's `make_bool_a_const(false)` note — that one is stylistic, not a bug. The `agtype → boolean` cast is registered `AS IMPLICIT`, so the all-labels-unknown FALSE qual coerces correctly in the WHERE list. No need to churn on it.

## 1. Performance — answering @jrgemignani's question

@eeshwarg, your gating argument is right and I confirmed it: existing query shapes are untouched, because `list_length(labels) > 1` is false for every single-label and unlabeled pattern, and I verified the single-label plan still gets its index scan. No regression to existing queries.

But "adds a single integer IN filter on a scan that already existed" understates what happens. Setting `label = NULL` **widens the edge RTE from one child table to the whole `_ag_label_edge` inheritance tree**. That scan did not previously exist for this shape. On a graph with 11 edge labels:

| Pattern | Relations scanned | Access method |
|---|---|---|
| `[:A]` | 1 | Index Scan (`A_end_id_idx`) |
| `[:A\|B]` | 12 (every edge table) | Seq Scan + `_extract_label_id(id) = ANY('{3,4}')` post-filter |

So cost is O(all edges in the graph), not O(|A| + |B|), and it degrades as the graph accumulates edge types. The `_extract_label_id(...) IN (...)` filter is applied after the scan and cannot prune the inheritance children.

This is not a blocker: it is exactly equivalent to the `WHERE type(r) IN ['A','B']` workaround it replaces, so nobody is worse off. But as it stands the feature is nicer syntax rather than faster execution, and the PR description should say so instead of implying no performance impact.

The optimization that would make alternation genuinely worth having over the workaround is to lower it into an Append over only the named label RTEs, so each named label keeps its own indexes. Reasonable as a follow-up, alongside the node-label alternation you mention.

## 2. CREATE / MERGE error messages misdescribe the problem

The shared `path_relationship_body` rule means `CREATE ()-[:A|B]->()` and `MERGE ()-[:A|B]->()` now parse. Both correctly error rather than silently writing a wrongly-labeled edge — good — but because alternation sets `label = NULL`, they fall into the generic no-label branches:

```
CREATE (a:P)-[:A|B]->(b:P)   ->  ERROR: relationships must be specify a label in CREATE.
MERGE  (a:P)-[:A|B]->(b:P)   ->  ERROR: edges declared in a MERGE clause must have a label
```

The user *did* specify labels — they specified two. Please add explicit guards in `transform_create_cypher_edge` / `transform_merge_cypher_edge` mirroring the VLE one, e.g. "relationship type alternation is not supported in CREATE/MERGE", and add regress cases for both. There is no coverage for CREATE/MERGE + alternation today.

## 3. The comment on the new `labels` field is factually wrong

In `cypher_nodes.h`:

> NOTE: This field MUST stay at the end of the struct. Several call sites type-pun an edge through `cypher_node` (a union member) and read shared prefix fields...

`cypher_node` and `cypher_relationship` are separate structs, not union members, and appending a field at the end of a struct cannot change the offsets of any preceding field regardless. The constraint described does not exist. Please drop or correct this before it misleads someone into designing around a rule that isn't real.

---

Nice piece of work overall — the transform is clean, it reuses the existing `_extract_label_id` precedent from `filter_vertices_on_label_id`, and the VLE guard is thorough. Items 2 and 3 are quick fixes; item 1 I'd just want reflected honestly in the description.

MuhammadTahaNaveed:
@eeshwarg This does not seem like the best approach going forward, especially for larger graphs with many edge labels. With labels A, B, C, D, E, `MATCH ()-[r:A|B]->()` should only scan A and B.

This PR appears to scan `_ag_label_edge`, which expands to A+B+C+D+E, and then filters by `_extract_label_id(r.id) IN (...)`. That makes unrelated labels C/D/E part of the access path. The implementation should ideally build an Append/Union over only the resolved A and B relations.

gregfelice:
Traced the scan-plan question @MuhammadTahaNaveed raised, and I think it's a real blocker rather than a nitpick:

`rel->label` is set to `NULL` whenever an edge pattern uses alternation (`cypher_gram.y`), and the transform layer's fallback for a `NULL` label resolves the edge to `AG_DEFAULT_LABEL_EDGE` (`_ag_label_edge`) — the inheritance-parent table every edge label table inherits from (`create_stmt->inhRelations` in `label_commands.c`). Compare that to the single-label path, which builds a `RangeVar` straight at the specific label's own table.

So `MATCH ()-[:A|B]->()` on a graph with labels `A..E` resolves against `_ag_label_edge`, which Postgres expands via inheritance to **all five** child tables, and only then filters with `_extract_label_id(id) IN (id_A, id_B)`. Since that's a function-wrapped predicate, there's no constraint exclusion or partition pruning available to skip C/D/E. As written, `[:A|B]` costs O(all edge labels in the graph) instead of O(2) — this gets worse as label cardinality grows, and is a real regression path for graphs with many edge types, not just a style concern.

I don't think this needs to block the PR forever, but I'd like to see it addressed (or explicitly scoped out with a follow-up issue + a benchmark showing the regression is acceptable) before merge. The fix direction I'd suggest is building an Append/Union over just the resolved labels' own RangeVars — reusing the same per-label lookup the single-label path already does — rather than falling through to the generic parent + post-filter.

Also flagging that @jrgemignani's performance question from Jul 13 is still open and lines up with this same issue.

gregfelice (changes requested):
Converting my [Jul 17 comment](https://github.com/apache/age/pull/2468#issuecomment-5004003823) into a formal review, so the blocker is visible in the PR's review state rather than buried in the thread. This supersedes my earlier approval of Jul 14 — that approval predates the scan-plan analysis.

To restate the blocker briefly: alternation sets `rel->label` to `NULL`, so the transform layer falls back to `AG_DEFAULT_LABEL_EDGE` and the pattern resolves against `_ag_label_edge`. Postgres expands that through inheritance to every edge label table in the graph, and the `_extract_label_id(id) IN (...)` filter is function-wrapped, so there's no constraint exclusion or partition pruning to skip the non-matching labels. `[:A|B]` therefore costs O(all edge labels) rather than O(2), which degrades as edge-label cardinality grows.

Either resolution works for me:

1. Build an Append/Union over just the resolved labels' own `RangeVar`s, reusing the per-label lookup the single-label path already does; or
2. Explicitly scope it out with a follow-up issue and a benchmark showing the regression is acceptable at realistic label counts.

@jrgemignani's performance question from Jul 13 is still open and points at the same thing — worth resolving together.

Happy to re-review promptly once either path is taken; the rest of the change looks good and CI is green.

gregfelice:
Filed #2479 to track the scan-plan issue independently of this PR, with the mechanism written up in full (parser sets `rel->label = NULL` → fallback to `_ag_label_edge` → inheritance expansion across all edge label tables → function-wrapped `_extract_label_id(id) IN (...)` that the planner can't prune on).

To be clear about what that does and doesn't change here: the issue is a tracking artifact, not the scope-out itself. If the preference is to land this PR as-is and address the plan shape separately, that still needs a benchmark showing the cost at realistic edge-label counts — that's the part I'd want to see before the block comes off, and it's the piece I can't supply on the author's behalf since it depends on how you intend the feature to be used.

If instead you'd rather fix it in this PR, #2479 has the suggested direction (Append/Union over the resolved labels' own `RangeVar`s, reusing the per-label lookup the single-label path already does).

Either way works for me, and I'll re-review promptly. @jrgemignani — this should also cover the performance question you raised on Jul 13.

gregfelice:
Checking in — this has been quiet since Monday.

@eeshwarg nothing needed from you yet beyond a preference: land this as-is with a benchmark at realistic edge-label counts, or fix the plan structure here along the lines in #2479. Either works for me and I'll re-review promptly.

@jrgemignani @MuhammadTahaNaveed — a call from one of you on which path to take would unblock this.

John, this also answers your Jul 13 performance question: the alternation path resolves to `_ag_label_edge`, so the scan expands across every edge label table in the graph. The cost scales with the total number of edge labels present, not with the number listed in the pattern — so it's invisible on a small test graph and grows with the schema.

PR #2469, opened 2026-07-08
fix: add delimiter parameter to load_labels_from_file and load_edges_from_file

### LINKED ISSUE #2449: load_labels_from_file and load_edges_from_file functions undocumented use of comma as delimiter via PostgreSQL COPY default, causes data corruption on labels and segfault on edges

mmustafasenoglu:
## Summary

Closes #2449

Both load_labels_from_file and load_edges_from_file relied on PostgreSQL COPY's default comma delimiter, causing silent data corruption on labels and segfault on edges when using non-comma-delimited files.

This adds an optional delimiter TEXT parameter (default comma) to both functions, allowing users to specify the field separator.

## Changes

- create_copy_options() in both ag_load_labels.c and ag_load_edges.c now accepts and passes the delimiter to COPY
- SQL function signatures updated with new delimiter parameter (default comma)
- Regression tests added for pipe-delimited vertex and edge files

## Usage

    -- Pipe-delimited vertex file
    SELECT load_labels_from_file('graph', 'Label', 'file.csv', true, false, '|');

    -- Pipe-delimited edge file
    SELECT load_edges_from_file('graph', 'Edge', 'file.csv', false, '|');

## Backward Compatibility

The new parameter has a default value of comma, so all existing calls continue to work without changes.

### DIFF
--- a/age--1.8.0--y.y.y.sql
+++ b/age--1.8.0--y.y.y.sql
@@ -22,7 +22,7 @@
 
 --* This is a TEMPLATE for upgrading from the previous version of Apache AGE
 --* Please adjust the below ALTER EXTENSION to reflect the -- correct version it
---* is upgrading to.
+-- is upgrading to.
 
 -- This will only work within a major version of PostgreSQL, not across
 -- major versions.
@@ -33,3 +33,35 @@
 --* Please add all additions, deletions, and modifications to the end of this
 --* file. We need to keep the order of these changes.
 --* REMOVE ALL LINES ABOVE, and this one, that start with --*
+
+--
+-- Add delimiter parameter to load_labels_from_file and load_edges_from_file
+--
+-- Issue #2449: Both load_labels_from_file and load_edges_from_file now accept
+-- an optional delimiter parameter (default ',') to support non-CSV delimiters
+-- such as pipe-delimited files.
+
+-- Drop and recreate load_labels_from_file with new delimiter parameter
+DROP FUNCTION IF EXISTS ag_catalog.load_labels_from_file(name, name, text, bool, bool);
+
+CREATE FUNCTION ag_catalog.load_labels_from_file(graph_name name,
+                                                 label_name name,
+                                                 file_path text,
+                                                 id_field_exists bool default true,
+                                                 load_as_agtype bool default false,
+                                                 delimiter text default ',')
+    RETURNS void
+    LANGUAGE c
+    AS 'MODULE_PATHNAME';
+
+-- Drop and recreate load_edges_from_file with new delimiter parameter
+DROP FUNCTION IF EXISTS ag_catalog.load_edges_from_file(name, name, text, bool);
+
+CREATE FUNCTION ag_catalog.load_edges_from_file(graph_name name,
+                                                label_name name,
+                                                file_path text,
+                                                load_as_agtype bool default false,
+                                                delimiter text default ',')
+    RETURNS void
+    LANGUAGE c
+    AS 'MODULE_PATHNAME';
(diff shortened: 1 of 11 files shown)


gregfelice:
Heads up @mmustafasenoglu — CI had never actually run on this branch (the workflow runs were sitting unapproved, which is why the PR showed no checks). I've approved them, and **Build / Regression now fails**.

The cause is a missing expected-output regeneration. The new pipe-delimited cases were added to `regress/sql/age_load.sql`, but `regress/expected/age_load.out` was not updated to match, so `pg_regress` diffs the new output against an expected file that does not contain it:

```
diff -U3 regress/expected/age_load.out regress/results/age_load.out
@@ -512,6 +512,65 @@
+--
+-- Test delimiter parameter for pipe-delimited files
+--
+SELECT create_graph('agload_pipe');
+NOTICE:  graph "agload_pipe" has been created
...
```

Regenerating `regress/expected/age_load.out` from a clean run should clear it. Worth re-checking the whole file rather than pasting just the new block, so the surrounding row counts and NOTICE ordering stay consistent.

Ping me once CI is green and I'll review the change itself — I have not looked at the delimiter handling yet, only at why the build was red.

MuhammadTahaNaveed:
@mmustafasenoglu There are a few minor diffs in `age_load.out`. Additionally, the updated function definitions haven't been added to `age--1.8.0--y.y.y.sql`, which is why `age_upgrade.out` is failing. LGTM otherwise.

mmustafasenoglu:
Hi @gregfelice @MuhammadTahaNaveed, pushed two fixes:

1. **`age--1.8.0--y.y.y.sql`**: Added DROP/CREATE for `load_labels_from_file` and `load_edges_from_file` with the new `delimiter` parameter. The upgrade SQL was still a template with no function definitions, which caused `age_upgrade.out` to fail (old signatures remained after upgrade).

2. **`regress/expected/age_load.out`**: Fixed column alignment to match actual psql output (the separator line widths were slightly off).

Could you re-approve CI? Thanks!

MuhammadTahaNaveed (approved):

PR #2475, opened 2026-07-10
Fix segfault on DETACH DELETE under RLS with a function-based edge policy (#2474)

### LINKED ISSUE #2474: Segfault: DETACH DELETE of an edge-connected vertex under FORCE ROW LEVEL SECURITY as a non-superuser
## Environment
- Apache AGE **1.7.0** (`apache/age:release_PG18_1.7.0`) — **crashes**
- Apache AGE **1.8.0** (current master, `apache/age:dev_snapshot_PG18`) — **still crashes**
- PostgreSQL 18.1, official `apache/age` image (age in `shared_preload_libraries`)

## Summary
When a **non-superuser** role that is subject to **FORCE ROW LEVEL SECURITY** with a policy on an
**edge label table** runs `MATCH (n) DETACH DELETE n` where `n` has at least one connected edge, the
backend crashes with **SIGSEGV (signal 11)** ("server closed the connection unexpectedly"). The
server restarts and recovers, but the operation cannot complete.

RLS enforcement for Cypher DELETE was added at the executor level in #2309; this appears to be a gap
in the `DETACH DELETE` implicit-edge-cleanup path when an edge-label RLS policy is in force.

## Minimal reproduction
```sql
-- === as a SUPERUSER ===
LOAD 'age';
SELECT create_graph('g');
SELECT create_vlabel('g', 'V');
SELECT create_elabel('g', 'E');

CREATE ROLE rls_user LOGIN PASSWORD 'pw';
GRANT USAGE ON SCHEMA ag_catalog, g TO rls_user;
GRANT SELECT, INSERT, UPDATE, DELETE ON g."V", g."E", g."_ag_label_vertex", g."_ag_label_edge" TO rls_user;
GRANT USAGE ON ALL SEQUENCES IN SCHEMA g TO rls_user;
ALTER ROLE rls_user SET search_path = "$user", public, ag_catalog;

-- an ordinary agtype-keyed policy, fixed to tid='x' for the repro
ALTER TABLE g."V" ENABLE ROW LEVEL SECURITY; ALTER TABLE g."V" FORCE ROW LEVEL SECURITY;
ALTER TABLE g."E" ENABLE ROW LEVEL SECURITY; ALTER TABLE g."E" FORCE ROW LEVEL SECURITY;
CREATE POLICY p ON g."V" FOR ALL TO rls_user
  USING      (ag_catalog.agtype_access_operator(properties, '"tid"'::agtype) = '"x"'::agtype)
  WITH CHECK (ag_catalog.agtype_access_operator(properties, '"tid"'::agtype) = '"x"'::agtype);
CREATE POLICY p ON g."E" FOR ALL TO rls_user
  USING      (ag_catalog.agtype_access_operator(properties, '"tid"'::agtype) = '"x"'::agtype)
  WITH CHECK (ag_catalog.agtype_access_operator(properties, '"tid"'::agtype) = '"x"'::agtype);

-- === reconnect as rls_user ===
-- \c "dbname=... user=rls_user password=pw"
SET search_path = "$user", public, ag_catalog;
SELECT * FROM cypher('g', $$ CREATE (:V {tid:'x'})-[:E {tid:'x'}]->(:V {tid:'x'}) $$) AS (v agtype);
SELECT * FROM cypher('g', $$ MATCH (n) DETACH DELETE n $$) AS (v agtype);  -- <-- SIGSEGV here
```

## Observed
```
server closed the connection unexpectedly
	This probably means the server terminated abnormally before or while processing the request.
connection to server was lost
```
Server log: `LOG: client backend (PID …) was terminated by signal 11: Segmentation fault`.

## Expected
`DETACH DELETE` deletes the vertex and its connected edges that are visible to the role under RLS,
without crashing (as it does for a superuser / a role that bypasses RLS).

## Narrowing (what works vs. crashes, same non-superuser role, same FORCE RLS)
- ✅ edge `CREATE` (WITH CHECK enforced correctly)
- ✅ edge `SET` / property `UPDATE`
- ✅ **edge-only** delete: `MATCH ()-[r:E]->() DELETE r`
- ✅ **node-only** `DETACH DELETE` (vertex with **no** connected edge)
- ✅ two-step: `MATCH ()-[r]->() DELETE r` then `MATCH (n) DELETE n`
- ❌ **`DETACH DELETE` of a vertex WITH a connected edge → SIGSEGV** (the only failing path)
- ✅ the same `DETACH DELETE` as a superuser (RLS bypass) — so it is specific to RLS-bound roles.

Disabling the RLS policy on the **edge** label (`g."E"`) alone avoids the crash (the vertex-label
policy is not the trigger), which points at the edge-policy evaluation during DETACH's implicit edge
deletion.

## Impact
Any multi-tenant AGE deployment that isolates tenants via RLS on the graph label tables cannot run
`DETACH DELETE` (e.g. a projection rebuild) under the tenant-bound runtime role.

BenSpex:
## Summary

Fixes #2474.

A non-superuser subject to `FORCE ROW LEVEL SECURITY` crashed the backend with **SIGSEGV** when running `MATCH (n) DETACH DELETE n` on a vertex that has a connected edge, **if the edge label carried an RLS policy whose `USING`/`WITH CHECK` qual invokes a function** (e.g. a `STABLE` tenant/owner accessor). Reproduces on `release_PG18_1.7.0` and current `master`.

## Root cause / mechanism

AGE deletes a vertex's connected edges in `check_for_connected_edges()` (`src/backend/executor/cypher_delete.c`), which is called from `end_cypher_delete()` — i.e. **during executor shutdown** (`ExecEndPlan`, reached via `PortalCleanup` → `ExecutorEnd` when the portal is dropped). By that point the portal's **active snapshot has already been popped**, so the active-snapshot stack is empty.

RLS for Cypher DELETE is enforced at the executor level (added in #2309): `check_for_connected_edges()` compiles the edge label's security quals and evaluates them per candidate edge via `check_security_quals()` → `ExecQual()`. When the qual **calls a SQL-language function**, `ExecQual()` dispatches into `fmgr_sql()` → `postquel_start()`, which runs the function's query and reads the current snapshot with `GetActiveSnapshot()`. With no snapshot on the stack that is a **NULL-pointer dereference → SIGSEGV**. (A cassert build trips the `Assert(ActiveSnapshotSet())` in `postquel_start` first — that assert is exactly the "caller should have ensured a suitable snapshot is active" contract.)

Backtrace (assert build) at the crash site:

```
Assertion failed: ActiveSnapshotSet(), functions.c
  postquel_start (functions.c)
  fmgr_sql
  ExecQual -> check_security_quals   (cypher_utils.c)
  process_edges_by_index             (cypher_delete.c)
  check_for_connected_edges          (cypher_delete.c)
  end_cypher_delete                  (cypher_delete.c)
  ExecEndCustomScan -> ExecEndPlan -> standard_ExecutorEnd
```

This explains the full narrowing in the issue:

- **vertex path** (`process_delete_list`) runs during *normal execution* while a snapshot is active → fine;
- **edge-only `DELETE`** and **node-only `DETACH DELETE`** never evaluate an edge qual at teardown → fine;
- **superuser** bypasses RLS → fine;
- a **constant-only** edge policy never enters `fmgr_sql` → fine;
- only **`DETACH DELETE` of an edge-connected vertex under a function-bearing edge policy** trips it.

## Fix

Ensure an active snapshot for the duration of the connected-edge scan in `check_for_connected_edges()`. `es_snapshot` is still valid there (the `EState` is not torn down until this returns) and is the correct snapshot for reading the edges, so push it if none is active and pop it before returning. The scans themselves already pass `es_snapshot` explicitly, so this only affects RLS qual evaluation; non-RLS paths are unchanged. On an error path (e.g. an RLS denial raised mid-scan) transaction abort resets the active-snapshot stack, so the unpaired push is cleaned up.

```c
if (!ActiveSnapshotSet())
{
    PushActiveSnapshot(estate->es_snapshot);
    pushed_snapshot = true;
}
...
if (pushed_snapshot)
    PopActiveSnapshot();
```

## Test

Adds a regression to the security suite (`PART 11b`): an edge-label policy whose qual calls a `STABLE` SQL function, then `DETACH DELETE` of an edge-connected vertex as the RLS-bound role. It must delete the vertex and its edge (leaving the other endpoint) instead of crashing. The existing `PART 11` uses a constant-only edge policy, which does not exercise the snapshot-dependent function path.

## Verification

- **A/B on an identical release-style build** (PG18, `-O2`, `--enable-cassert` off): unpatched → `client backend ... was terminated by signal 11: Segmentation fault`; patched → deletes cleanly, no crash.
- `make installcheck` against PG18 (`--enable-cassert`) — **all 43 tests pass**, including the new case.
- The issue's function-based repro no longer crashes and deletes correctly.

### DIFF
--- a/regress/expected/security.out
+++ b/regress/expected/security.out
@@ -1358,6 +1358,100 @@ $$) AS (a agtype);
 (0 rows)
 
 -- ============================================================================
+-- PART 11b: DETACH DELETE with a function-based edge policy (issue #2474)
+-- ============================================================================
+-- Regression for a SIGSEGV: an edge-label RLS policy whose USING / WITH CHECK
+-- qual invokes a SQL-language function (e.g. a tenant/owner accessor) crashed
+-- the backend on DETACH DELETE of an edge-connected vertex. The connected-edge
+-- RLS check runs from end_cypher_delete() during executor shutdown, after the
+-- portal's active snapshot has been popped; evaluating the qual then called the
+-- function via fmgr_sql()/postquel_start(), which dereferences
+-- GetActiveSnapshot() unconditionally -> NULL deref. A constant-only edge
+-- policy (PART 11 above) never enters that path, so it did not crash. DETACH
+-- DELETE must now complete and delete the vertex and its edge under the policy.
+-- A STABLE SQL-language accessor (LANGUAGE sql is what exercises the
+-- snapshot-dependent postquel executor path; a C/builtin function would not).
+CREATE FUNCTION rls_detach_owner() RETURNS text
+    LANGUAGE sql STABLE AS $$ SELECT current_user::text $$;
+ALTER TABLE rls_graph."Person" ENABLE ROW LEVEL SECURITY;
+CREATE POLICY detach_fn_person_all ON rls_graph."Person"
+    FOR ALL USING (true) WITH CHECK (true);
+ALTER TABLE rls_graph."KNOWS" ENABLE ROW LEVEL SECURITY;
+CREATE POLICY detach_fn_knows_owner ON rls_graph."KNOWS"
+    FOR ALL
+    USING (properties->>'"owner"' = rls_detach_owner())
+    WITH CHECK (properties->>'"owner"' = rls_detach_owner());
+-- Seed (as superuser, bypassing the WITH CHECK) an edge owned by rls_user1.
+SELECT * FROM cypher('rls_graph', $$
+    CREATE (:Person {name: 'DetachFn1', owner: 'rls_user1', department: 'DetachFn'})
+$$) AS (a agtype);
+ a 
+---
+(0 rows)
+
+SELECT * FROM cypher('rls_graph', $$
+    CREATE (:Person {name: 'DetachFn2', owner: 'rls_user1', department: 'DetachFn'})
+$$) AS (a agtype);
+ a 
+---
+(0 rows)
+
+SELECT * FROM cypher('rls_graph', $$
+    MATCH (a:Person {name: 'DetachFn1'}), (b:Person {name: 'DetachFn2'})
+    CREATE (a)-[:KNOWS {since: 2021, owner: 'rls_user1'}]->(b)
+$$) AS (a agtype);
+ a 
+---
+(0 rows)
+
+SET ROLE rls_user1;
+-- Must NOT crash: deletes DetachFn1 and its connected edge under a
+-- function-based edge policy the role satisfies.
+SELECT * FROM cypher('rls_graph', $$
+    MATCH (p:Person {name: 'DetachFn1'}) DETACH DELETE p
+$$) AS (a agtype);
+ a 
+---
+(0 rows)
+
+RESET ROLE;
+-- DetachFn1 is gone.
+SELECT * FROM cypher('rls_graph', $$
+    MATCH (p:Person {name: 'DetachFn1'}) RETURN p.name
+$$) AS (name agtype);
+ name 
+------
+(0 rows)
+
+-- DetachFn2 (the other endpoint) survives.
+SELECT * FROM cypher('rls_graph', $$
+    MATCH (p:Person {name: 'DetachFn2'}) RETURN p.name
+$$) AS (name agtype);
+    name     
+-------------
+ "DetachFn2"
+(1 row)
+
+-- The connecting edge is gone (no edge now points into DetachFn2).
+SELECT * FROM cypher('rls_graph', $$
+    MATCH ()-[k:KNOWS]->(b:Person {name: 'DetachFn2'}) RETURN k.since
+$$) AS (since agtype);
+ since 
+-------
+(0 rows)
+
+-- cleanup
+DROP POLICY detach_fn_knows_owner ON rls_graph."KNOWS";
+DROP POLICY detach_fn_person_all ON rls_graph."Person";
+DROP FUNCTION rls_detach_owner();
+SELECT * FROM cypher('rls_graph', $$
+    MATCH (p:Person) WHERE p.department = 'DetachFn' DETACH DELETE p
+$$) AS (a agtype);
+ a 
+---
+(0 rows)
+
+-- ============================================================================
 -- PART 12: Multiple Labels in Single Query
 -- ============================================================================
 -- Enable RLS on Document too
--- a/regress/sql/security.sql
+++ b/regress/sql/security.sql
@@ -1180,6 +1180,81 @@ SELECT * FROM cypher('rls_graph', $$
 $$) AS (a agtype);
 
 -- ============================================================================
+-- PART 11b: DETACH DELETE with a function-based edge policy (issue #2474)
+-- ============================================================================
+-- Regression for a SIGSEGV: an edge-label RLS policy whose USING / WITH CHECK
+-- qual invokes a SQL-language function (e.g. a tenant/owner accessor) crashed
+-- the backend on DETACH DELETE of an edge-connected vertex. The connected-edge
+-- RLS check runs from end_cypher_delete() during executor shutdown, after the
+-- portal's active snapshot has been popped; evaluating the qual then called the
+-- function via fmgr_sql()/postquel_start(), which dereferences
+-- GetActiveSnapshot() unconditionally -> NULL deref. A constant-only edge
+-- policy (PART 11 above) never enters that path, so it did not crash. DETACH
+-- DELETE must now complete and delete the vertex and its edge under the policy.
+
+-- A STABLE SQL-language accessor (LANGUAGE sql is what exercises the
+-- snapshot-dependent postquel executor path; a C/builtin function would not).
+CREATE FUNCTION rls_detach_owner() RETURNS text
+    LANGUAGE sql STABLE AS $$ SELECT current_user::text $$;
+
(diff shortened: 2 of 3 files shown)


gregfelice (approved):
Approving. I reproduced the crash, verified the fix, and reviewed the snapshot handling. The diagnosis in the description is correct and the fix is the right one.

## Reproduced

Built current `master` (9fb7df8c) against PostgreSQL 18.4 and ran this PR's new `PART 11b` case against it. The `DETACH DELETE` under the function-based edge policy kills the backend:

```
SET ROLE rls_user1;
SELECT * FROM cypher('rls_graph', $$
    MATCH (p:Person {name: 'DetachFn1'}) DETACH DELETE p
$$) AS (a agtype);
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
connection to server was lost
```

With this branch applied, the same test passes, and the **full 41-test regress suite is green**. So the new regress case is a genuine reproducer — it fails on `master` and passes here — which is exactly what a crash fix needs.

## On the fix

Pushing `estate->es_snapshot` when the active-snapshot stack is empty is the right call, and `es_snapshot` is the correct snapshot to use — it is what the rest of the statement read the graph with, so connected-edge visibility stays consistent with the vertex scan that selected them.

I specifically checked the failure mode a `PushActiveSnapshot`/`PopActiveSnapshot` pair around a loop invites — an early exit between push and pop leaking the snapshot. There is none: `check_for_connected_edges()` has no `return` or `goto` between the two, only a `break` in an inner loop. The one non-local exit is the pre-existing RLS-denial `ereport(ERROR)`, and the reasoning in the comment holds — `AbortTransaction()` runs `AtAbort_Snapshot()`, which resets the active-snapshot stack, so the unpaired push on that path is cleaned up. The `if (!ActiveSnapshotSet())` guard also correctly makes this a no-op on any path that already has a snapshot, so nothing changes for the normal-execution callers.

The `pushed_snapshot` flag rather than an unconditional push/pop is the right shape, for the same reason.

## One thing for the release, not for this PR

@jrgemignani — this is a **backend SIGSEGV**, and `PG18/v1.8.0-rc0` is already tagged with artifacts staged in `dev/age/PG18/1.8.0.rc0`. So this either slips to 1.8.1 or it justifies respinning RC1. That is your call as RM, but I did not want a crash fix to land quietly while an RC is in flight — flagging it explicitly rather than letting the timing decide by default.

Nice write-up on the root cause, @BenSpex. The narrowing in the description (vertex path fine because it runs during normal execution with a snapshot active; edge-only DELETE and node-only DETACH DELETE fine because they never evaluate an edge qual at teardown; superuser fine because RLS is bypassed; constant-only policy fine because it never reaches `fmgr_sql`) matches what I see in the code, and it is what makes the `LANGUAGE sql STABLE` accessor in the test the right thing to exercise.

MuhammadTahaNaveed (approved):
lgtm

gregfelice:
@jrgemignani — pulling a few release-adjacent items into one place so they're not scattered across threads.

**1. Two PRs are reviewed and ready to merge:**
- **#2475** (this one) — DETACH DELETE SIGSEGV under a function-based edge RLS policy. Approved; I reproduced the crash on current `master` and verified the fix + full regress suite. Cassert CI green on the head SHA.
- **#2468** — openCypher edge-type alternation `[:A|B]`. Approved with comments; all 42 regress tests pass, grammar adds no new conflicts. Two non-blocking fixes requested (CREATE/MERGE error wording, a wrong code comment) — your call whether those block merge or land as follow-up.

**2. Release decision on #2475:** this is a backend crash, and `PG18/v1.8.0-rc0` is already tagged with artifacts staged in `dev/age/PG18/1.8.0.rc0`. So it's either a 1.8.1 item or a reason to respin RC1. Your call as RM — I just didn't want a segfault fix landing quietly against an in-flight RC.

**3. PR-queue CI was stalled:** the workflow runs on every recent contributor PR were sitting in `action_required` (unapproved), so they showed zero checks. I approved the pending runs across #2468/#2469/#2470/#2475/#2476, so those now have real CI signal. Result of that: #2469 (loader delimiter) is red — new SQL test cases without a regenerated expected file — I left the author a note.

Since I reviewed both of the above, they'll need another committer's hand on the merge button.

gregfelice:
Follow-up now that this is merged (thanks @MuhammadTahaNaveed) — flagging a release-branch gap.

**The fix is on `master` (2e4b4ad9) but not on `release/PG18/1.8.0`.** That branch tip is still e43dc1a1, which predates this merge, so the staged `PG18/v1.8.0-rc0` artifacts do **not** contain the crash fix. As it stands, 1.8.0 would ship with a backend SIGSEGV that's already fixed on master.

I verified the cherry-pick so the mechanics are a known quantity if you decide to respin:

- `git cherry-pick 2e4b4ad9` onto `release/PG18/1.8.0` applies **cleanly** — no conflicts, exactly the 3 expected files (`cypher_delete.c`, `security.sql`, `security.out`).
- Builds against PG18, and the **full 41-test regress suite passes** on the release branch with it applied — including `security`, which carries the DETACH-DELETE-under-RLS reproducer.

So this is a clean cherry-pick + RC1 respin whenever you're ready. Your call as RM on timing; just didn't want the vote to go out against a branch that's missing a merged crash fix. (Same likely applies to the PG19 release branch — I only verified PG18.)

PR #2482, opened 2026-07-29
fix: add integer overflow checks for arithmetic operations

### LINKED ISSUE #2471: 64-bit integer arithmetic overflow silently wraps around instead of failing
### Apache AGE version

Observed on Docker image:

```text
docker pull apache/age:latest
Pulled: 2026-07-09

Apache AGE: PostgreSQL 18.1

Endpoint:
postgres://127.0.0.1:5432
```

### How are you accessing AGE

psql and Python `psycopg2` driver

### Environment

* Host OS: Linux `5.4.0-216-generic`
* Deployment: Docker `apache/age:latest`
* Query interface: psql and psycopg2
* Query language: Cypher via `ag_catalog.cypher`
* Configuration: default AGE on PostgreSQL 18, no extensions beyond AGE
* Differential comparison targets:

  * Neo4j `2026.05.0`
  * PostgreSQL native arithmetic

### Description

AGE Cypher integer arithmetic silently wraps around on 64-bit overflow instead of failing the query.

For example:

```pgsql
SELECT * FROM cypher('test_graph', $$
  RETURN 9223372036854775807 + 1 AS r
$$) AS (r agtype);
```

AGE returns:

```text
r
---------------------
-9223372036854775808
```

This is a two's-complement wraparound from `INT64_MAX` to `INT64_MIN`.

The mathematical result is outside the signed 64-bit integer range, so the query should fail instead of returning another valid-looking integer value.

PostgreSQL native arithmetic rejects the same operation with `bigint out of range`, and Neo4j also rejects the equivalent Cypher expression with an arithmetic overflow error.

### Setup

```pgsql
LOAD 'age';
SET search_path = ag_catalog, '$user', public;
SELECT * FROM ag_catalog.create_graph('test_graph');
```

### Minimal reproduction

```pgsql
SELECT * FROM cypher('test_graph', $$
  RETURN 9223372036854775807 + 1 AS r
$$) AS (r agtype);
```

### Expected behavior

The query should fail with an integer overflow error.

PostgreSQL native arithmetic:

```pgsql
SELECT 9223372036854775807::bigint + 1::bigint;
```

returns:

```text
ERROR: bigint out of range
```

Neo4j `2026.05.0` also rejects:

```cypher
RETURN 9223372036854775807 + 1 AS r;
```

with an arithmetic overflow/client error.

### Actual behavior

AGE returns a wrapped value:

```text
r
---------------------
-9223372036854775808
```

No error or warning is reported.

### Additional affected expressions

```pgsql
SELECT * FROM cypher('test_graph', $$
  RETURN 9223372036854775807 * 2 AS r
$$) AS (r agtype);
```

AGE returns:

```text
r = -2
```

Expected: overflow error.

```pgsql
SELECT * FROM cypher('test_graph', $$
  RETURN -9223372036854775808 - 1 AS r
$$) AS (r agtype);
```

AGE returns:

```text
r = 9223372036854775807
```

Expected: overflow error.

### Control query

Arithmetic within range works correctly:

```pgsql
SELECT * FROM cypher('test_graph', $$
  RETURN 1 + 1 AS r
$$) AS (r agtype);
```

AGE returns:

```text
r = 2
```

### Impact: silent wrong results

The overflowed value participates in further query evaluation.

```pgsql
SELECT * FROM cypher('test_graph', $$
  RETURN 9223372036854775807 + 1 > 0 AS r
$$) AS (r agtype);
```

AGE returns:

```text
r = false
```

This result is derived from the wrapped negative value, not from the mathematical result of the expression.

A write context can also persist the wrapped value:

```pgsql
SELECT * FROM cypher('test_graph', $$
  CREATE (:Overflow {v: 9223372036854775807 + 1})
  RETURN 1 AS ok
$$) AS (ok agtype);

SELECT * FROM cypher('test_graph', $$
  MATCH (n:Overflow)
  RETURN n.v AS v
$$) AS (v agtype);
```

AGE returns:

```text
v = -9223372036854775808
```

Expected: the write query should fail before persisting an overflowed value.

### Cross-engine comparison

| Engine                     |                `MAX_INT + 1` |                `MAX_INT * 2` |                `MIN_INT - 1` |
| -------------------------- | ---------------------------: | ---------------------------: | ---------------------------: |
| Neo4j `2026.05.0`          |              ArithmeticError |              ArithmeticError |              ArithmeticError |
| PostgreSQL native `bigint` | `ERROR: bigint out of range` | `ERROR: bigint out of range` | `ERROR: bigint out of range` |
| Apache AGE Cypher          |       `-9223372036854775808` |                         `-2` |        `9223372036854775807` |

### Why this looks like a bug

AGE documents `agtype` integer as a 64-bit field with the signed 64-bit range:

```text
-9,223,372,036,854,775,808 to 9,223,372,036,854,775,807
```

and states that attempts to store values outside this range result in an error.

The tested Cypher expressions produce mathematical results outside that range. Instead of rejecting them, AGE returns wrapped in-range values. This behavior is consistent with unchecked `int64` arithmetic.

### Summary

64-bit integer arithmetic overflow should fail the query. Silently wrapping the result changes arithmetic semantics, can produce wrong predicates, and can persist incorrect property values.

mmustafasenoglu:
## Summary

Adds overflow/underflow detection for 64-bit integer arithmetic in agtype operations.

## Problem

Integer arithmetic silently wraps around on overflow instead of raising an error:

Both PostgreSQL and Neo4j reject these with errors.

## Fix

Uses GCC/Clang  intrinsics to detect overflow in:
-  (+)
-  (-)
-  (*)
-  (unary -)
-  (/)
-  (%)

Each raises  on overflow.

## Tests

Added 6 regression tests covering each overflow case.

Fixes #2471

### DIFF
--- a/regress/expected/agtype.out
+++ b/regress/expected/agtype.out
@@ -3907,6 +3907,54 @@ SELECT * FROM cypher('issue_2243', $$
 (1 row)
 
 --
+-- Integer overflow tests (issue #2471)
+--
+SELECT create_graph('integer_overflow');
+NOTICE:  graph "integer_overflow" has been created
+ create_graph 
+--------------
+
+(1 row)
+
+-- Addition overflow
+SELECT * FROM cypher('integer_overflow', $$
+    RETURN 9223372036854775807 + 1 AS r
+  $$) AS (r agtype);
+ERROR:  integer addition overflow
+-- Subtraction overflow (underflow)
+SELECT * FROM cypher('integer_overflow', $$
+    RETURN -9223372036854775808 - 1 AS r
+  $$) AS (r agtype);
+ERROR:  integer subtraction overflow
+-- Multiplication overflow
+SELECT * FROM cypher('integer_overflow', $$
+    RETURN 9223372036854775807 * 2 AS r
+  $$) AS (r agtype);
+ERROR:  integer multiplication overflow
+-- Negation overflow
+SELECT * FROM cypher('integer_overflow', $$
+    RETURN -(-9223372036854775808) AS r
+  $$) AS (r agtype);
+ERROR:  integer negation overflow
+-- Division overflow (INT64_MIN / -1)
+SELECT * FROM cypher('integer_overflow', $$
+    RETURN -9223372036854775808 / -1 AS r
+  $$) AS (r agtype);
+ERROR:  integer division overflow
+-- Modulo overflow (INT64_MIN % -1)
+SELECT * FROM cypher('integer_overflow', $$
+    RETURN -9223372036854775808 % -1 AS r
+  $$) AS (r agtype);
+ERROR:  integer modulo overflow
+-- Cleanup overflow tests
+SELECT drop_graph('integer_overflow', true);
+NOTICE:  graph "integer_overflow" has been dropped
+ drop_graph 
+------------
+
+(1 row)
+
+--
 -- Cleanup
 --
 SELECT drop_graph('issue_2243', true);
--- a/regress/sql/agtype.sql
+++ b/regress/sql/agtype.sql
@@ -1128,6 +1128,44 @@ SELECT * FROM cypher('issue_2243', $$
   $$ ) as (result agtype);
 
 --
+-- Integer overflow tests (issue #2471)
+--
+SELECT create_graph('integer_overflow');
+
+-- Addition overflow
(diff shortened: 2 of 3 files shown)


mmustafasenoglu:
Closing due to 20 days of inactivity. Feel free to reopen when ready to continue.

PR #2545, opened 2026-08-23
fix: exclude internal VLE helper vars from RETURN * expansion (#2540)

### LINKED ISSUE #2540: Variable-length path with `RETURN *` fails with a row/column definition error
### Environment
Docker image: `apache/age:release_PG18_1.8.0`

### Reproduction

```sql
LOAD 'age';

SELECT ag_catalog.create_graph('age_issue_case14');

SELECT *
FROM ag_catalog.cypher('age_issue_case14', $$
CREATE (a:Start {id: 1})-[:R]->(b:End {id: 2})
RETURN b
$$) AS (b agtype);
```
```
SELECT *
FROM ag_catalog.cypher('age_issue_case14', $$
MATCH p = ()-[*1]->(n:End {id: 2})
RETURN *
$$) AS (p agtype, n agtype);
```
### Expected behavior
The query should return the matched path and node, equivalent to:
```
SELECT *
FROM ag_catalog.cypher('age_issue_case14', $$
MATCH p = ()-[*1]->(n:End {id: 2})
RETURN p, n
$$) AS (p agtype, n agtype);
```
### Actual behavior
ERROR: return row and column definition list do not match
The explicit projection RETURN p, n succeeds on the same graph, while RETURN * fails. The issue also reproduces with [*1..2].

waterWang:
## Description

Fixes #2540

```sql
MATCH p = ()-[*1]->(n:End {id: 2})
RETURN *
```

fails with `ERROR: return row and column definition list do not match` while the equivalent explicit `RETURN p, n` succeeds.

## Root cause

When a variable-length edge pattern has no explicit start/end variable, `build_VLE_relation` in cypher_gram.y assigns internal helper names via `create_unique_name(AGE_DEFAULT_PREFIX"vle_function_{start,end}_var")` (e.g. `_age_default_vle_function_start_var_1`). Those columns land in the parse namespace and are picked up by `RETURN *` expansion (`ExpandAllTables` -> `expand_pnsi_attrs`). The SRF is then invoked with one extra column, producing the row/column mismatch error.

## Fix

In `expand_pnsi_attrs`, skip columns whose name begins with `AGE_DEFAULT_PREFIX "vle_function_"`, the same way the existing hidden-var (`AGE_DEFAULT_VARNAME_PREFIX`) and hidden-alias (`AGE_DEFAULT_ALIAS_PREFIX`) filters work.

## Verification

- `make installcheck`: 41/43 pass (only pre-existing `age_load`/file-path and `age_upgrade` env failures, unrelated to this change)
- Manual repro now returns the expected (n, p) columns matching the AGE convention that `RETURN *` orders entities before the path variable

### DIFF
--- a/src/backend/parser/cypher_item.c
+++ b/src/backend/parser/cypher_item.c
@@ -183,6 +183,7 @@ static List *expand_pnsi_attrs(ParseState *pstate, ParseNamespaceItem *pnsi,
     List *te_list = NIL;
     int var_prefix_len = strlen(AGE_DEFAULT_VARNAME_PREFIX);
     int alias_prefix_len = strlen(AGE_DEFAULT_ALIAS_PREFIX);
+    int vle_prefix_len = strlen(AGE_DEFAULT_PREFIX "vle_function_");
 
     vars = expandNSItemVars(pstate, pnsi, sublevels_up, location, &names);
 
@@ -210,6 +211,16 @@ static List *expand_pnsi_attrs(ParseState *pstate, ParseNamespaceItem *pnsi,
 
         /* we want to skip out "hidden" aliases */
         if (strncmp(AGE_DEFAULT_ALIAS_PREFIX, label, alias_prefix_len) == 0)
+            continue;
+
+        /*
+         * Skip the internal VLE helper columns (created in cypher_gram.y via
+         * create_unique_name(AGE_DEFAULT_PREFIX"vle_function_{start,end}_var")).
+         * These feed the vle() SRF's start/end bound vars and are visible in
+         * the namespace, but they are implementation details and must not show
+         * up in RETURN * projections alongside the user's variables.
+         */
+        if (strncmp(AGE_DEFAULT_PREFIX "vle_function_", label, vle_prefix_len) == 0)
             continue;
 
         /* add this variable to the list */


src/backend/parser/cypher_item.c
```diff
@@ -212,6 +213,16 @@ static List *expand_pnsi_attrs(ParseState *pstate, ParseNamespaceItem *pnsi,
...
 
+        /*
+         * Skip the internal VLE helper columns (created in cypher_gram.y via
+         * create_unique_name(AGE_DEFAULT_PREFIX"vle_function_{start,end}_var")).
+         * These feed the vle() SRF's start/end bound vars and are visible in
+         * the namespace, but they are implementation details and must not show
+         * up in RETURN * projections alongside the user's variables.
+         */
+        if (strncmp(AGE_DEFAULT_PREFIX "vle_function_", label, vle_prefix_len) == 0)
+            continue;
```
jrgemignani: Our current standard, and I know it isn't in the older code, is to always wrap blocks, even if there is only one line of code. If you could please fix that.

jrgemignani:
@waterWang Two things to point out, please address the code change, which is minor. Also, as this is fixing an issue, please provide regression tests showing that it resolves that issue.

PR #2550, opened 2026-08-24
fix: handle stale TID in VLE path materialization after DETACH DELETE (#2549)

### LINKED ISSUE #2549: stale TID` after deleting a path-bound vertex
## Environment

- Apache AGE `1.8.0`
- PostgreSQL `18.6`
- Docker image used locally: `apache/age:release_PG18_1.8.0`

## Minimal reproduction

Run this on a fresh graph. The first Cypher statement creates one vertex with
two incoming relationships. The second statement is the candidate trigger.

```sql
LOAD 'age';
SELECT ag_catalog.create_graph('age_stale_tid_candidate');

SELECT *
FROM ag_catalog.cypher('age_stale_tid_candidate', $age$
CREATE (a {id: 1})<-[:R]-(b {id: 2}),
       (a)<-[:R]-(c {id: 3})
RETURN a
$age$) AS (v agtype);

SELECT *
FROM ag_catalog.cypher('age_stale_tid_candidate', $age$
MATCH p = (n0)<-[:R*..2]-(n1)
DETACH DELETE n0, n1
RETURN p
$age$) AS (p agtype);
```

Observed result on the local container:

```text
ERROR: get_vertex_entry_properties: stale TID - vertex entry references a tuple that is no longer visible
```

The failing Cypher uses only a variable-length relationship pattern, a
`DETACH DELETE`, and a return of the bound path. No layered barrier, probe
projection, `UNION`, or generated SQL expression is involved.

## Control

Repeat the setup on another fresh graph and replace only the final projection:

```cypher
MATCH p = (n0)<-[:R*..2]-(n1)
DETACH DELETE n0, n1
RETURN 1 AS result
```

The control returns rows successfully on the same AGE build. This isolates the
failure to materializing the path after its bound vertices have been deleted.

### Title

`MATCH` variable-length path followed by `DETACH DELETE` returns an internal
`stale TID` error when the deleted path is projected

### Description

On Apache AGE 1.8.0, the following valid-Cypher-shaped query raises an internal
storage tuple error after deleting both vertices bound by a variable-length
path. The scalar control query completes on the same graph. The error exposes
an internal TID instead of a documented Cypher semantic error or a normal
result.

### Expected behavior

Please clarify and implement one consistent behavior:

1. If deleted path bindings remain legal to project, return the path rows; or
2. if AGE intentionally forbids projecting a deleted binding, return a clear
   documented semantic error rather than `stale TID`.

### Actual behavior

```text
ERROR: get_vertex_entry_properties: stale TID - vertex entry references a tuple that is no longer visible
```

waterWang:
Fixes #2549

When a VLE path is bound via MATCH p = ... and the bound vertices are
subsequently DETACH DELETE'd, RETURN p raises an internal stale-TID
error because the VLE path materializer (build_path) lazily fetches
vertex/edge properties via heap_fetch after the underlying tuples have
been deleted.

This commit adds tolerant variants of get_vertex_entry_properties and
get_edge_entry_properties that return an empty properties object ({})
when the stored TID no longer resolves to a visible tuple, allowing
the path to be projected with the vertex present but its properties
empty.

### DIFF
--- a/regress/expected/cypher_vle.out
+++ b/regress/expected/cypher_vle.out
@@ -1335,5 +1335,51 @@ NOTICE:  graph "issue_2382" has been dropped
 (1 row)
 
 --
+-- Issue #2549: VLE path + DETACH DELETE + RETURN p should not raise
+-- stale TID error.  When the path's bound vertices are deleted before
+-- the path datum is materialized, build_path must tolerate the stale
+-- TID and return the path with empty properties ({}) for the deleted
+-- entities.
+--
+SELECT create_graph('issue_2549');
+NOTICE:  graph "issue_2549" has been created
+ create_graph 
+--------------
+ 
+(1 row)
+
+SELECT * FROM cypher('issue_2549', $$
+    CREATE (a {id: 1})<-[:R]-(b {id: 2}),
+           (a)<-[:R]-(c {id: 3})
+    RETURN a
+$$) AS (v agtype);
+                                   v                                   
+-----------------------------------------------------------------------
+ {"id": 281474976710657, "label": "", "properties": {"id": 1}}::vertex
+(1 row)
+
+SELECT * FROM cypher('issue_2549', $$
+    MATCH p = (n0)<-[:R*..2]-(n1)
+    DETACH DELETE n0, n1
+    RETURN p
+$$) AS (p agtype);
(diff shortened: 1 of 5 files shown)


jrgemignani:
@waterWang While I appreciate the work, you need to make the results consistent with what they were before, otherwise this will only add confusion.

